PR #3, opened 2024-12-20
Fix unreadable code in logs

floppyhammer:
Before

image

After

image


widgetii:
Thank you for your help, we really appreciate it!

PR #17, opened 2025-07-27
更新 EepromManager.cpp

KennyPlus:
Change index check offset

### DIFF
--- a/src/EepromManager.cpp
+++ b/src/EepromManager.cpp
@@ -495,7 +495,7 @@ void EepromManager::Hal_EfuseReadEFuse8812A(uint16_t _offset,
                 eFuse_Addr);
 }
 
-#define EEPROM_TX_PWR_INX_8812 0x20
+#define EEPROM_TX_PWR_INX_8812 0x22
 
 bool EepromManager::IsEfuseTxPowerInfoValid(uint8_t *efuseEepromData) {
   uint16_t tx_index_offset = EEPROM_TX_PWR_INX_8812;


josephnef:
@vertexodessa What was the purpose of changing EEPROM_TX_PWR_INX_8812 index from 0x10 to 0x20 in (PR#16)[https://github.com/OpenIPC/devourer/pull/16]?

The same question about it for @KennyPlus but this time changing it to 0x22?

KennyPlus:
for ignore the 2G detection of the driver initialize, which causes NIC driver loading crashes.

josephnef:
We cannot change it in every single PR back and forth. Let’s figure out how to change it from a constant to a configurable setting instead.

KennyPlus:
> We cannot change it in every single PR back and forth. Let’s figure out how to change it from a constant to a configurable setting instead.

This offset is constant and it's a fixed address on HW.

vertexodessa:
hi @josephnef sorry about the confusion and late reply.

The point of the fix for Bonnet is to switch from checking presence of power calibration values in 2.4GHz range, to 5.8GHz range, since 2.4GHz range is not calibrated in Bonnet (values are all 0xFF).

While my fix also works, fix from @KennyPlus is more correct because it's using correct offset for 5.8GHz range calibration values
please merge this PR

vertexodessa:
it doesn't need to be configurable, the correct value is 0x22 as suggested by Kenny

josephnef:
Now it’s clear, thanks a lot!

PR #18, opened 2025-09-03
Fix macOS build

floppyhammer:


josephnef:
Thank you for your contribution. Could you also add MacOS to CI to ensure that the very next commit doesn’t break MacOS support?

floppyhammer:
@josephnef It's done.

PR #21, opened 2026-05-18
Add RTL8811AU support via RF_TYPE_ID autodetect

josephnef:
## Summary

- **Root cause:** `EepromManager::read_chip_version_8812a` hardcoded `RFType = RF_TYPE_2T2R`, so the existing 1T1R code paths (`PHY_BB8812_Config_1T`, the `numTotalRfPath` loops in `PHY_RF6052_Config_8812` / TX power, per-BW `L1pkVal` branches, 2.4 GHz eLNA threshold) were unreachable on RTL8811AU dongles even though the driver could otherwise drive the chip.
- **Fix:** derive `RFType` from `REG_SYS_CFG` bit 27 (`RF_TYPE_ID`) the same way Realtek's upstream rtl8812au driver does — set → 1T1R (RTL8811AU), clear → 2T2R (RTL8812AU). Same firmware, same power sequence, same EFUSE layout; only the chain count and a handful of BB constants differ, all of which are already conditionalised.
- **Manual override preserved:** `registry_priv::special_rf_path == 1` still forces 1T1R after autodetect, in case a board's EFUSE/strap is mis-burnt.
- **Demo PID widening:** `demo/main.cpp` now tries `0x8812 → 0x0811 → 0xa811 → 0xb811` before giving up. `0xc811` is intentionally excluded — it is shared with the unrelated **RTL8811CU** chipset, which uses a different driver family entirely and is not supported by this code.

## Test plan

Need hardware verification before merge:

- [ ] **8812AU regression:** plug a known RTL8812AU dongle, run `WiFiDriverDemo`, confirm log shows `RF_Type is 2 TotalTxPath is 2` and beacon RX still works on ch36/20 MHz.
- [ ] **8811AU positive:** plug a known RTL8811AU dongle (verified via `lsusb` — VID `0bda`, PID one of `0811`/`a811`/`b811`/`8812`, **not** `c811`), run `WiFiDriverDemo`, confirm log shows `RF_Type is 0 TotalTxPath is 1` and beacons are received.
- [ ] **TX smoke test:** `WiFiDriverTxDemo` on the 8811AU and confirm the hardcoded beacon is seen by a separate sniffer.
- [ ] **macOS / Android builds:** verify the demo's `kRealtekProductIds` table still links under `_MSC_VER`, `__ANDROID__`, `__APPLE__` paths (compile-tested on macOS; the table is `static constexpr`, no platform-specific behaviour).

## Open questions for the reporter (#20)

[...]

Issue #20, opened 2026-05-18
Suppost for rtl8811au chipset


BadPotato1007:
Hey! 
So I was wondering if this project will be supporting rtl8811au based chips aswell? The only difference is that the 8812 supports two mimo streams, and the 8811 nly supports 1, but otherwise both cards use the same drivers and are able perform all the same functions, ofc, given that the speed is a little bit slower

josephnef:
Hey @BadPotato1007 — yes, your read is correct: 8811AU is the 1×1 cut of the same Jaguar silicon as 8812AU, and the 1T1R code paths were already present in the driver but unreachable because chip detection hardcoded `RFType = 2T2R`.

Opened #21 which derives `RFType` from `REG_SYS_CFG` bit 27 (`RF_TYPE_ID`) the way Realtek's upstream rtl8812au driver does, and also widens the demo's USB PID list to include the common 8811AU SKUs.

Would you mind testing on your hardware? A few things would help me close this out:

1. `lsusb` output for your dongle — we need to confirm it's an 8811**AU** (PIDs `0bda:0811`, `a811`, `b811`, sometimes `8812`) and not an 8811**CU** (`0bda:c811` — different silicon, won't work with this driver no matter what we do; sellers often mislabel them).
2. Run `WiFiDriverDemo` after building #21 and paste the log line that starts `RF_Type is …` — should read `RF_Type is 0 TotalTxPath is 1` on an 8811AU.
3. Any 5 GHz association / scan oddities? A few upstream 8811AU issues trace back to RFE-Type / antenna-byte mis-reads from EFUSE; if you hit that, the log will tell us.

Thanks for the report.

BadPotato1007:
Thanks for your reply! I shall definitely take some time out for this in a
day or two! I was also wondering about a couple of rtl8821au cards that I
have, it seems that the only difference between them is the fact that the
21 chip supports bluetooth aswell? Possible to add that?

josephnef:
Sure, take your time with the 8811AU test.

[...]

PR #22, opened 2026-05-19
rtl8821au support

RomanLut:
Attempt to add rtl8811au support.

Still no luck, waiting for hardware.

The missing parts were taken from https://github.com/svpcom/rtl8812au/tree/v5.2.20-rssi-fix-but-sometimes-crash

### DIFF
--- a/hal/basic_types.h
+++ b/hal/basic_types.h
@@ -24,22 +24,12 @@
 #define le16_to_cpu(x) ((__u16)(x))
 
 
-#ifndef LE_BITS_TO_4BYTE
 #define LE_BITS_TO_4BYTE(__pStart, __BitOffset, __BitLen)                      \
   ((LE_P4BYTE_TO_HOST_4BYTE(__pStart) >> (__BitOffset)) &                      \
    BIT_LEN_MASK_32(__BitLen))
-#endif
-
-#ifndef LE_P4BYTE_TO_HOST_4BYTE
 #define LE_P4BYTE_TO_HOST_4BYTE(__pStart) (le32_to_cpu(*((u32 *)(__pStart))))
-#endif
-
-#ifndef BIT_LEN_MASK_32
 #define BIT_LEN_MASK_32(__BitLen) ((u32)(0xFFFFFFFF >> (32 - (__BitLen))))
-#endif
 
-#ifndef ReadLE2Byte
 #define ReadLE2Byte(_ptr) le16_to_cpu(*((u16 *)(_ptr)))
-#endif
 
 #endif /* BASIC_TYPES_H */
--- a/hal/basic_types.h
+++ b/hal/basic_types.h
@@ -24,12 +24,22 @@
 #define le16_to_cpu(x) ((__u16)(x))
 
 
+#ifndef LE_BITS_TO_4BYTE
 #define LE_BITS_TO_4BYTE(__pStart, __BitOffset, __BitLen)                      \
   ((LE_P4BYTE_TO_HOST_4BYTE(__pStart) >> (__BitOffset)) &                      \
    BIT_LEN_MASK_32(__BitLen))
-#define LE_P4BYTE_TO_HOST_4BYTE(__pStart) (le32_to_cpu(*((u32 *)(__pStart))))
-#define BIT_LEN_MASK_32(__BitLen) ((u32)(0xFFFFFFFF >> (32 - (__BitLen))))
+#endif
 
+#ifndef LE_P4BYTE_TO_HOST_4BYTE
+#define LE_P4BYTE_TO_HOST_4BYTE(__pStart) (le32_to_cpu(*((u32 *)(__pStart))))
+#endif
+
+#ifndef BIT_LEN_MASK_32
+#define BIT_LEN_MASK_32(__BitLen) ((u32)(0xFFFFFFFF >> (32 - (__BitLen))))
+#endif
+
+#ifndef ReadLE2Byte
 #define ReadLE2Byte(_ptr) le16_to_cpu(*((u16 *)(_ptr)))
+#endif
 
 #endif /* BASIC_TYPES_H */
--- a/src/EepromManager.cpp
+++ b/src/EepromManager.cpp
@@ -80,48 +80,119 @@ void EepromManager::LateInitFor8814A() {
                 LNAType_2G, LNAType_5G);
 }
 
+/* RTL8821AU is Jaguar-family silicon (CHIP_8821 = 7 in Realtek's HalVerDef),
+ * 1T1R AC + BT combo. It responds to REG_SYS_CFG with an 8812-shaped value
+ * (low byte 0x35), so plain SYS_CFG-bit inspection misidentifies it as 8812.
+ * Distinguish by VID:PID at probe time. Genuine 8821AU SKUs we know about:
+ *  - TP-Link Archer T2U Plus / T4U / etc:  2357:0120 / 011E / 0122
+ *  - Realtek reference PIDs:               0bda:0820 / 0821 / 0823 / 8822
+ *  - OEM rebadges (Senao, ASUS, Edimax, D-Link, Hawking, …):
+ *    0411:0242 / 029B   (Senao)
+ *    0846:9052          (ASUS)
+ *    0e66:0023          (Hawking)
+ *    2001:3314 / 3318   (D-Link)
+ *    20f4:804b          (TRENDnet)
+ *    7392:a811/a812/a813/b611 (Edimax)
+ *    3823:6249          (HP/MaxNet)
+ *    2019:ab32          (Planex)
+ *    04bb:0953          (I-O Data)
+ *    056e:4007 / 400e / 400f (ELECOM)
+ *
+ * NOT included on purpose:
+ *  - 0bda:0811 / a811 / b811  — those are 8811AU (1T1R cut of 8812 silicon);
+ *    they ride on CHIP_8812 + RFType=1T1R, NOT through the 8821 HAL.
+ *  - 0bda:c811  — RTL8811CU, unrelated silicon, not supported here. */
+static bool is_rtl8821a_pid(uint16_t vid, uint16_t pid) {
+  const uint32_t key = (static_cast<uint32_t>(vid) << 16) | pid;
+  switch (key) {
+  case 0x0BDA0820:
+  case 0x0BDA0821:
+  case 0x0BDA0823:
+  case 0x0BDA8822:
+  case 0x0411'0242:
+  case 0x0411'029B:
+  case 0x04BB'0953:
+  case 0x056E'4007:
+  case 0x056E'400E:
+  case 0x056E'400F:
+  case 0x0846'9052:
+  case 0x0E66'0023:
+  case 0x2001'3314:
+  case 0x2001'3318:
+  case 0x2019'AB32:
+  case 0x20F4'804B:
+  case 0x2357'011E:
+  case 0x2357'0120:
+  case 0x2357'0122:
+  case 0x3823'6249:
+  case 0x7392'A811:
+  case 0x7392'A812:
+  case 0x7392'A813:
+  case 0x7392'B611:
+    return true;
+  default:
+    return false;
+  }
+}
+
 void EepromManager::read_chip_version_8812a(RtlUsbAdapter device) {
   uint32_t value32 = device.rtw_read32(REG_SYS_CFG);
   _logger->info("read_chip_version_8812a SYS_CFG(0x{:X})=0x{:08X}", REG_SYS_CFG,
                 value32);
 
-  const bool is_8814a = device.idProduct() == 0x8813;
-  const bool is_8821 = !is_8814a && device.IsRtl8821A();
+  const uint16_t pid = device.idProduct();
+  const uint16_t vid = device.idVendor();
+  const bool is_8814a = (pid == 0x8813);
+  const bool is_8821 = !is_8814a && is_rtl8821a_pid(vid, pid);
+
+  HAL_IC_TYPE_E ic_type;
+  if (is_8814a) {
+    ic_type = CHIP_8814A;
+  } else if (is_8821) {
+    ic_type = CHIP_8821;
+  } else {
+    ic_type = CHIP_8812;
+  }
+
+  /* RFType:
+   *   8814AU      → 4T4R RF (baseband caps at 3 spatial streams)
+   *   8821AU      → always 1T1R (Jaguar-family AC+BT combo, single chain)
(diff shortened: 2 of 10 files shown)


josephnef:
Thanks for porting this — adding the 8821a HAL tree is exactly the missing piece for actual RTL8821AU dongles. Two coordination notes plus one logistics ask.

**Scope / dispatch.** The PID allow-list in `RtlUsbAdapter::InitDvObj` currently lumps real RTL8811AU PIDs (`0bda:0811`, `0bda:a811`) into `RtlChipType::RTL8821`. Those are Jaguar silicon (CHIP_8812 with `RFType=1T1R`), not Jaguar2 — sending them through the 8821 HAL with the 8821 firmware blob shouldn't work. Suggest narrowing the table to genuinely-8821 SKUs (e.g. `2357:0120`, `0bda:0820/0821/0823`, the 8821-combo OEM IDs) and letting the existing 8812 path handle 8811AU. Even better long-term would be a SYS_CFG-based family check (IC-ID bits) so new SKU revs don't need a code patch — that's what upstream rtl8812au does.

**Conflict with #21.** I have a small PR open (#21) that touches the same lines in `EepromManager::read_chip_version_8812a` — it derives `RFType` from `REG_SYS_CFG` bit 27 (`RF_TYPE_ID`) so the already-present 1T1R BB/RF paths reach actual 8811AU dongles. Whichever lands first will force the other to rebase. Happy to coordinate — I think the two compose cleanly once #22's dispatch is narrowed:
- #22's family detect chooses CHIP_8821 vs CHIP_8812 HAL tree
- inside CHIP_8812, #21's SYS_CFG read picks 1T1R vs 2T2R

**Hardware / logistics.** The reporter on #20 (BadPotato1007) has a TP-Link Archer T2U Plus (`2357:0120`) — that's already in your allow-list — and is up for testing #22, which would close the loop. Also: we'd like to keep one in the OpenIPC lab for regressions. Where did you source / where are you planning to source your test dongle from? Happy to order from the same place so we're testing against an identical SKU rev.

RomanLut:
>Where did you source / where are you planning to source your test dongle from? 

https://www.amazon.de/-/en/dp/B07P681N66?ref=ppx_yo2ov_dt_b_fed_asin_title&th=1
I hope correct chipset arrive. You may wait a few days untill I check.

josephnef:
Putting this on hold — RTL8811AU support landed via #23 (squash-merged f05b743), so master now carries the 8811 code path. We'd like to keep this PR around as a second-source data point but are postponing review/merge until we have RTL8811AU hardware in hand to verify against before integrating any deltas. Thanks for the contribution.

josephnef:
**Hardware test (2026-05-23).** Finally got a TP-Link Archer T2U Plus (`2357:0120`, RTL8821AU) in-lab on a Mac. Ran current master (`5de6d62`) against it to characterise the failure mode before re-attempting this PR.

RX and TX die at the same stage:

```
<devourer>read_chip_version_8812a SYS_CFG(0xF0)=0x04412535
<devourer>Chip Version Info: CHIP_8812_Normal_Chip_TSMC_D_CUT_2T2R_RomVer(0)
... (EFUSE autoload OK, sensible defaults) ...
<devourer>power-on :REG_SYS_CLKR 0x09=0x30. REG_CR 0x100=0xEA
<devourer>MAC has not been powered on yet
<devourer>InitPowerOn: run power on flow fail
```

Master applies `Rtl8812_NIC_ENABLE_FLOW` to 8821AU silicon and `HalPwrSeqCmdParsing` returns false — chip stays in CARDEMU, never transitions to ACT. This is exactly what your `hal/Hal8821APwrSeq.{c,h}` is for. USB open / claim / reset / EFUSE autoload all work; the gap is purely in the HAL.

RomanLut:
Renamed topic to to rtl8821au support because actual chip in TPLink Archer T2U Plus is rtl8821au despite the markings.

This PR is verified in hx-esp32cam-fpv Android GS app https://github.com/RomanLut/hx-esp32-cam-fpv/actions/runs/26363043264 and hx-esp32cam-fpv Oculus Quest GS https://github.com/RomanLut/hx-esp32-cam-fpv/actions/runs/26363043273 development builds with rtl8812au and rtl8821au.

What is working:
- High rate RX, low rate TX
- hotplug
- dual adapter mode

Not checked: 
- rtl8811au, rtl8814au

josephnef:
**Lab validation against master (post-#38).** Ran this branch (`292d28a`) through `tests/regress.py --full-matrix` on a 3-adapter rig (`0bda:8812 + 0bda:8813 + 2357:0120`) with kernel cells in the pinned-kernel VM. Channel 100, 15 s/cell, 2026-05-24.

### Kernel-only (rig sanity)

| TX \ RX | RTL8814AU | RTL8812AU | RTL8821AU |
|---|---|---|---|
| RTL8814AU | — | 148 ✓ | 413 ✓ |
| RTL8812AU | 346 ✓ | — | 425 ✓ |
| RTL8821AU | 376 ✓ | 164 ✓ | — |

6/6 pass — rig is sound.

### devourer-TX → kernel-RX

| TX \ RX | RTL8814AU | RTL8812AU | RTL8821AU |
|---|---|---|---|
| RTL8814AU | — | 0 ✗ (4500 fail) | 0 ✗ (4610 fail) |
| RTL8812AU | 6043 ✓ | — | 6851 ✓ |
| RTL8821AU | 6412 ✓ | 0 ✗ (7000 fail) | — |

- 8812 TX works (≈6000 hits vs master's 4114/4693 — within noise/rig variance)
- 8821 TX → 8814 RX works (6412 vs master 4341)
- 8821 TX → 8812 RX: 0, same as master
- 8814 TX: 0 in every cell, same as master

### kernel-TX → devourer-RX (the row that gates 8821)

| TX \ RX | RTL8814AU | RTL8812AU | RTL8821AU |
|---|---|---|---|
| RTL8814AU | — | 100 ✓ | 400 ✓ |
| RTL8812AU | 0 ✗ | — | **0 ✗** |
| RTL8821AU | 0 ✗ | 100 ✓ | — |

- 8812 RX works (100 hits)
- 8821 RX from 8814 TX works (400 hits — slightly up from master's 200)
- **8812 TX → 8821 RX still 0** — master's RX asymmetry is NOT fixed on this branch
- 8814 RX still 0 (known TODO)

### devourer ↔ devourer

All 0 — every cell hits at least one broken side.

---

**Read.** Counts on the cells that already pass on master are comparable here (within the variance you'd expect across two RF-environment-different runs), but no new cell flips from fail to pass. Specifically, the asymmetric `8812-TX → 8821-RX` cell that master's PR #37 punts on (\"partial — does not fix RX asymmetry\") is still 0 on this branch. Net: I can't see what improves vs master.

The other thing to flag is that this branch is now significantly behind master — PRs #25/#27/#29/#30/#37/#38 landed since you opened it, and the diff would revert a lot of that 8814 and 8821 trace-derived parity work.

Two questions, since your hx-esp32cam-fpv CI builds say it's working end-to-end:
1. Is there a specific TX rate / bandwidth / channel-width / radiotap configuration where your branch beats master that this 15 s/cell sweep wouldn't catch? Happy to add a targeted cell if so.
2. Have you tried master (post-#38) against the same Android / Quest GS build? If 8821 works there too, that's worth confirming before we choose between the two trajectories.

Happy to re-run with whatever test parameters you'd like. Logs at `/tmp/devourer-regress-last/` if you want raw inject/RX traces.

RomanLut:
rtl8821au works with the master branch, but after unplugging and plugging the adapter back in, it stops receiving packets until the application is restarted. The same issue happens with PixelPilot compiled against master, which suggests that the master branch does not perform proper init/deinit handling.

With this PR, hotplug works reliably. For rtl8812au, hotplug already worked before and still works correctly.

BTW, rtl8821au supports LDPC TX, but not LDPC RX. PixelPilot can not receive video stream from Eachine Sphere Link if LDPC is enabled. You may also want to add LDPC and STBC settings to the tests.

josephnef:
Thanks Roman — that's a fair pushback, and the hotplug case is a genuine gap in my matrix (each cell does one cold `CreateRtlDevice`, so I would not have caught a "second-Init-after-replug fails" bug). Independent reproduction in PixelPilot strengthens it.

Going to repro on master with the T2U Plus (cold init → confirmed RX → sysfs-unbind/rebind → second init → check RX), then run the same sequence on this branch. If I can confirm master regresses and this PR fixes it, the next question is what the minimal delta is — happy to credit you on a smaller landed patch if we can isolate it, or rebase yours.

Captured the 8821AU LDPC-TX-yes / LDPC-RX-no datum separately — useful for OpenIPC link planning and a TODO for me to extend `tests/regress.py` to iterate LDPC / STBC.

Will report back with the hotplug test.

josephnef:
Update from the hotplug investigation. Three forms of "hotplug" tested against the T2U Plus (`2357:0120`) on channel 6:

1. **Subprocess restart + sysfs unbind/rebind**: cold (1800 hits) → unbind/rebind → warm (1700 hits). Master works.
2. **Back-to-back subprocess runs, no unbind**: 1100 / 1400 / 1000 across 3 runs. Master works.
3. **5-cycle unbind/rebind stress**: 1700, 1400, 1700, 1700, 800 — wobbly counts (RF noise on ch6) but no failure mode.

Same three tests on this PR pass with comparable counts. So under any form of hotplug where the libusb context is torn down + re-initialised between sessions, master is fine.

That narrows your bug to the in-process case: app keeps libusb_init alive, destroys the old `RtlJaguarDevice`, opens a fresh handle for the replugged adapter, calls `CreateRtlDevice` again. That second Init produces a chip that doesn't RX. WiFiDriverDemo doesn't do this — each run is a full libusb_init/exit cycle.

Looking at your diff for things that could affect the in-process re-init path, the standout is `RtlUsbAdapter.cpp` ctor — you set `rxagg_usb_size=0x7, timeout=0x1a` (USB3) / `0x5, 0x20` (USB2), with the comment "Smaller thresholds can leave RTL8821AU monitor traffic stuck in the RXDMA path". Master uses `0x3, 0x01` / `0x1, 0x01`. Also `infinite_read` now returns early on rc<0 (master rate-limits + sleeps but stays in the bulk loop, which would hold a stale handle across an in-process replug).

Two questions:
1. Can you check if applying just the `RtlUsbAdapter.cpp` changes on top of current master (without the HAL init-flow changes) fixes hotplug for you? If yes, that's a tiny patch we can land with credit to you, no rebasing your whole branch needed.
2. If the HAL init-flow changes are also load-bearing, can you point at the specific one? `_InitQueueReservedPage_8821AUsb`, the LLT page boundary, the `REG_SYS_CFG+3 BIT0 → 0x7C |= BIT6` write, etc.

Saved this investigation in our internal notes; happy to test whichever isolation you can describe.

josephnef:
Follow-up. Built a custom in-process hotplug test (compiles directly against `libWiFiDriver.a`, single `libusb_init`, two `CreateRtlDevice` sessions across a sysfs unbind/rebind on the T2U Plus) to exercise exactly the in-process replug scenario your app does. Master `7239a68` vs PR #22 `292d28a`, channel 6:

| | iter1 (cold) | iter2 (post-replug) |
|---|---|---|
| master, clean teardown of h1 before h2 | 2538 | 2050 |
| master, leak h1 + RX thread across replug | 2902 | 3617 |
| PR #22, clean teardown | 2126 | 5665 |

Both branches survive in-process hotplug in our lab. PR #22 produces higher iter2 counts on average (which matches the rxagg threshold change — bigger DMA bursts), but master is well above zero RX in every case.

So I can't reproduce the "stops receiving packets until app restart" failure mode you describe — at least not with sysfs unbind/rebind + Linux libusb. Two scenarios we genuinely can't test from here that I suspect are at play:

1. **Android USB Host API hotplug semantics** — on Android, the disconnect/reconnect path goes through the Java `UsbDevice` permission flow + `libusb_wrap_sys_device` rather than `libusb_open_device_with_vid_pid`. That's a different libusb construction path; if devourer happens to be missing some init step on the `wrap_sys_device` codepath that the traditional open codepath provides for free, that'd produce exactly the symptom you describe and we'd never see it from a Linux test.

2. **High-rate RX under video bitrates** — our 2.4 GHz beacon traffic doesn't stress the RXDMA the way a 5 GHz video link at 50+ Mbps does; the rxagg threshold mismatch may only surface under that load.

Two asks (sorry for the back-and-forth, this is a slippery bug):
- Could you confirm whether your PixelPilot test goes through `libusb_wrap_sys_device` (i.e. you pass an Android-derived fd)? If so, that's almost certainly where the gap is, and the fix is probably much smaller than the wholesale supersession.
- Either way, if you can isolate which subset of your branch's changes you actually need (rxagg only? infinite_read only? HAL init flow?), we can land a targeted patch with attribution and you don't have to maintain a long-running fork.

If neither isolation is straightforward, an alternative path is: you keep this PR open as your stable branch for downstream users, and we explicitly document master as not-recommended-for-8821AU-hotplug until we figure it out. Happy to take the L on the comparison if your branch is what your users need today.

RomanLut:
1) yes, this test path goes through libusb_wrap_sys_device() using an fd obtained from Android’s UsbDeviceConnection.getFileDescriptor().

2) Here are the changes. I cannot guarantee that all of them are related to the hotplug issue, but all of them were derived from https://github.com/svpcom/rtl8812au/tree/v5.2.20-rssi-fix-but-sometimes-crash, so they are probably all important?

josephnef:
Update — your `caa46d92 "minimal fix"` isolation made this much easier to land. I cherry-picked the HalModule-only delta onto current master as #42 (now merged as `01cad75`). Linux matrix validation: no regressions, 8821 cells uniformly improved (271 → 430, 200 → 400, 260 → 365 hits; 8821-TX→8814-RX: 4341 → 5865).

You're credited via `Co-Authored-By:` on the squash commit.

**Could you build PixelPilot / hx-esp32cam-fpv against current master (post-`01cad75`) and confirm the hotplug failure mode goes away on Android?** That's the validation the Linux lab can't do. If confirmed, we can close this PR; if not, we'll keep investigating (e.g. revisit your `RtlUsbAdapter.cpp` rxagg / `infinite_read` changes that you reverted).

Thanks for doing the isolation — without your minimal-fix commit this would have stayed a 580-line wholesale port indefinitely.

josephnef:
Closing — superseded by #42 (merged as `01cad75`), which @RomanLut [just confirmed](https://github.com/OpenIPC/devourer/pull/42#issuecomment-4536700... ) fixes Android replug on both rtl8812au and rtl8821au end-to-end in PixelPilot + hx-esp32cam-fpv GS.

Thanks @RomanLut — your `caa46d92` minimal-fix isolation is what made the merge possible. You're credited via `Co-Authored-By:` on `01cad75`. The original svpcom-derived rxagg + `infinite_read` deltas (reverted in your minimal-fix) didn't turn out to be needed for the hotplug case; we can revisit those separately if any other behaviour calls for them.

PR #28, opened 2026-05-22
RTL8814AU: match TX descriptor to kernel-driver field-for-field

josephnef:
## Summary

Usbmon capture of the working morrownr/aircrack-ng kernel driver injecting a probe-request on the 8814 (channel 6, monitor mode) was diffed against devourer's TX descriptor, both 140-byte single-fragment frames. The first 32-byte descriptor block differed in seven fields. Update devourer's `send_packet` so the descriptor is byte-identical to the kernel-driver's.

| Field | Was | Now | Rationale |
|---|---|---|---|
| `MACID` | 0 | 1 | broadcast/default CAM |
| `RATE_ID` (non-VHT) | 7 | 8 | rate-table index |
| `GID` | 0 | 63 (`0x3F`) | no-group default |
| `SW_DEFINE` | 0 | 1 | `DriverFixedRate` flag |
| `RETRY_LIMIT_ENABLE` | 0 | 1 | mgmt-frame default |
| `DATA_RETRY_LIMIT` | 0 | 12 | per upstream `rtl8814au_xmit.c:267` |
| `SPE_RPT` | 1 | 0 | kernel does not set |
| `DISABLE_FB` | 1 | 0 | kernel does not set |

## Background

The previous values came from earlier WIP comments speculating about chip behavior. Live usbmon diff against the working kernel driver shows the actual field set the chip expects.

## Verification

- Devourer's first TX bulk-OUT now reads:
  `64002885 01120800 0000003f 00010000 00003200 00000000 01000000 76a90000` — byte-identical to kernel-driver's TX descriptor.
- With chip primed by the kernel driver (`insmod 8814au.ko`, monitor, ch6, `rmmod`) and devourer started with `DEVOURER_SKIP_INIT=1`: **781 of 781 bulk-OUT URBs complete with status=0**. With the previous descriptor, URBs also completed status=0 but on-air TX never happened (the chip silently dropped frames whose descriptor field values it didn't accept).
- 8814 RX still works end-to-end (10+ packets in demo window).

## What this does NOT fix

End-to-end 8814 TX without `DEVOURER_SKIP_INIT`. Devourer's HAL init still leaves the chip in a state where EP `0x02` times out URBs (a separate gate, tracked in task #21). With this PR, when the chip-init gate is fixed, the descriptor will already be correct.

## Test plan

- [x] Build green on macOS + Arch Linux 6.18
- [x] RX regression: 10+ packets on `0bda:8813` ch6
- [x] TX descriptor byte-equality with kernel-driver verified via usbmon
- [x] `DEVOURER_SKIP_INIT=1` path: 781/781 bulk-OUT URB completions status=0

### DIFF
--- a/src/RtlJaguarDevice.cpp
+++ b/src/RtlJaguarDevice.cpp
@@ -137,12 +137,8 @@ bool RtlJaguarDevice::send_packet(const uint8_t *packet, size_t length) {
 
   SET_TX_DESC_DATA_BW_8812(usb_frame, BWSettingOfDesc);
 
-  /* Upstream rtl8814a_xmit.c sets ONLY LAST_SEG=1 (not FIRST_SEG) for a
-   * single-fragment frame, and MACID = bmc_camid which is 0 for the
-   * broadcast/default CAM entry. Our previous values (FIRST_SEG=1, MACID=1)
-   * caused the chip to silently reject every bulk-OUT — verified by
-   * usbmon trace comparing the OOT-driver's working TX descriptor to
-   * ours at byte offsets 3 (flags) and 4 (MACID). */
+  /* Single-fragment frame: LAST_SEG=1 (no FIRST_SEG); OWN=1 so chip
+   * processes the descriptor. */
   SET_TX_DESC_LAST_SEG_8812(usb_frame, 1);
   SET_TX_DESC_OWN_8812(usb_frame, 1);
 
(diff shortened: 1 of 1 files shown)


josephnef:
Closing per user request — value comes only when TX works end-to-end. Will bundle the descriptor change into the eventual PR that fixes the chip-init gate as well.

PR #30, opened 2026-05-23
RTL8821AU: partial bring-up — chip init OK, RX silent — WIP

josephnef:
## Status: WIP — chip init succeeds, RX data flow still silent

Brings up the RTL8821AU end-to-end except for RX bulk-IN data flow. Posting so @RomanLut and others can continue the investigation from a clean checkpoint that builds against current master instead of resurrecting #22 (which deletes the 8814AU work and has a parallel dispatch enum that doesn't compose with the `ICType`-based convention #23-#29 established).

## What works on T2U Plus 2357:0120 (verified on macOS)

- Chip detection: `CHIP_8821_Normal_Chip_TSMC_D_CUT_1T1R`
- 8821 power-on flow (`rtl8821A_card_enable_flow`)
- Firmware download — 8821 blob, signature `0x2101`, FW ready in ~30ms
- MAC/BB/AGC/RadioA register tables via existing `PhyTableLoader` (same phydm conditional encoding as 8814AU)
- RFE pinmux (`phy_SetRFEReg8821`), band switch 2.4G/5G, channel + TX power table on ch6/36/100
- USB endpoint discovery: bulk IN 0x84, OUTs 0x05/0x06/0x08/0x09 (vs 8812/8814's 0x81 and 0x02-0x05)
- `libusb_clear_halt` on IN + `REG_USB_HRPWM=0x84` LPS wake

## What's NOT working

- **RX bulk-IN reads succeed at the USB layer but the chip never pushes data.** 0 RX packets across 15s on ch100 even with the host Mac actively associated to a busy 5GHz AP. The chip-internal RX-DMA → bulk-IN-EP binding isn't engaging despite all known init steps.
- TX path is wired (correct OUT EP, no `LIBUSB_ERROR_NOT_FOUND`) but unvalidated end-to-end on 8821AU — no peer sniffer in this session.

## Regression matrix (Linux trainer-arch, master vs `feat/rtl8821au-support`)

| Adapter | Test | Result | vs master |
|---|---|---|---|
| 8812AU (`0bda:8812`) | RX | 41 pkts / 15s, 0 errors, `CHIP_8812` detected | ✓ no regression |
| 8812AU | TX | 15 prints all `rc=1`, 0 failures (2 runs) | ✓ no regression |
| 8814AU (`0bda:8813`) | RX | 0 pkts | ✓ matches master baseline (pre-existing) |
| 8814AU | TX | `rc=1`, ~270-320 async failures (timing-variant) | ✓ matches master baseline |

The new `CHIP_8821` dispatch correctly routes 8812 → 8812 path and 8814 → 8814 path. No misrouting.

## Suggested next steps for whoever picks this up

1. usbmon trace of `aircrack-ng/rtl8812au`'s RX bring-up against an 8821AU on Linux; diff post-fwdl register writes vs ours (REG_TRXDMA_CTRL, REG_USB_AGG_TH/TO, REG_RXDMA_AGG_PG_TH, REG_USB_SPECIAL_OPTION).
2. Compare register state post-init (kernel-driver readback vs our post-init pyusb dump). Same technique that unblocked the 8814AU TX work.
3. Re-read @RomanLut's #22 `HalModule.cpp` for any 8821-specific init steps I didn't carry over when rewiring through `ICType`.

## What this preserves

- 8812AU support — untouched (default dispatch branch)
- 8814AU support — untouched (#23-#29 work preserved)
- Same `HAL_IC_TYPE_E` / `ICType` dispatch pattern; no parallel enum

## Attribution

8821a HAL data (`Hal8821APwrSeq`, `Hal8821PhyReg`, `hal8821a_fw`) ported verbatim from @RomanLut's #22 (svpcom/rtl8812au v5.2.20). Wiring re-done to follow master's existing convention.

Refs #20 (BadPotato1007's underlying request) and #22 (RomanLut's original port).


josephnef:
**Update: the "RX silent" conclusion in this PR appears to be wrong — RX path actually works under at least some conditions.**

The full-matrix regression rig (#34) was just run against a 3-adapter setup (0bda:8812 + 0bda:8813 + 2357:0120) with `aircrack-ng/88XXau` as the kernel-side reference. Two cells directly contradict the assumptions this PR was merged under:

- **devourer-RX 8821 caught 200 of 281 frames** sent by kernel-TX 8814 (~71% delivery rate). The PR concluded RX was silent based on a single test session against an actively-associated 5GHz AP; the regression rig instead injects a known-SA beacon from a peer adapter and confirmed devourer's 8821 RX path moves frames.

- **devourer-TX 8821 hit 4341 frames** to kernel-RX 8814 (~96% delivery). The PR documented TX as "wired but unvalidated" because no peer sniffer was available that session; that's also now confirmed working.

So both halves of the 8821 path that this PR labelled as gaps are actually functional. Two follow-up questions worth pinning down:

1. **Why was 8821 RX silent in the original PR session?** Possible candidates: the chip was in a degraded post-fwdl state from earlier test runs, ambient 5GHz channel was too quiet, or the SA-match filter logic missed something. Rerunning the original test conditions with the regression rig would isolate this.

2. **The 8821→8812 direction at the wire** showed 0 hits in mode 2 of the full matrix (devourer-TX 8821 → kernel-RX 8812), even though devourer-TX 8821 → kernel-RX 8814 worked fine in the same session. That's a different signal — possibly chip-state degradation by that cell, possibly 8812 kernel-RX in VM was stale. Worth a focused single-pair retest.

Tracking the deeper investigation in #35.

cc @RomanLut — this updates the situation since your #22 port. The dispatch is in master, and the RX path that we initially thought was broken seems to actually work — the chip just needs the right conditions / state.

PR #100, opened 2026-06-23
StreamTxDemo: fix mingw/Windows-GCC build and binary stdin

GingerFluffyCat:
## Problem

Building `StreamTxDemo` (the stream-link TX demo) with **mingw-w64 (GCC on Windows)** is broken in two independent ways. Both are mingw-only and do not affect the MSVC/vcpkg build.

### 1. Binary stdin not set on mingw → first frame truncated

`txdemo/stream_tx_demo/main.cpp` gates both the Windows includes (`io.h`/`fcntl.h`) and the `_setmode(_fileno(stdin), _O_BINARY)` call on `#if defined(_MSC_VER)`. mingw/GCC defines `_WIN32` but **not** `_MSC_VER`, so it falls into the POSIX `#else` branch and leaves stdin in **text mode**. A `0x1A` (or CRLF) byte in the binary `<u32_le len><PSDU>` stream then triggers:

```
stream_tx_demo: short read on stdin (76/269); record truncated mid-PSDU
```

and the demo exits before transmitting a single frame.

### 2. CMake aborts under mingw (no vcpkg)

`CMakeLists.txt` unconditionally sets `CMAKE_TOOLCHAIN_FILE` to `"$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"` on `WIN32`. With `VCPKG_ROOT` unset (the normal mingw / pkg-config case) this expands to the bogus `/scripts/buildsystems/vcpkg.cmake` and configuration fails with *"Could not find toolchain file"*.

## Fix

- Add a `__MINGW32__ || __MINGW64__` include branch (io.h + fcntl.h + unistd.h + libusb) and gate `_setmode` on `_WIN32` — identical behavior on MSVC, now also correct on mingw.
- Only set the vcpkg toolchain file when `VCPKG_ROOT` is non-empty.

## Verification

mingw-w64, libusb 1.0.28 via pkg-config, real RTL8812AU. After the fix `StreamTxDemo.exe` reads the full binary stream and radiates every frame:

```
<stream-tx>TX #1 ok=1 psdu=269 total=306
...
<stream-tx>done; sent 512 PSDUs
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,8 @@
 cmake_minimum_required(VERSION 3.15)
 
-# Set the toolchain file for vcpkg on Windows.
-if(WIN32)
+# Set the toolchain file for vcpkg on Windows (only if VCPKG_ROOT is set;
+# on mingw we get libusb from pkg-config directly).
+if(WIN32 AND NOT "$ENV{VCPKG_ROOT}" STREQUAL "")
     set(CMAKE_TOOLCHAIN_FILE "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake")
 endif()
 
--- a/txdemo/stream_duplex_demo/main.cpp
+++ b/txdemo/stream_duplex_demo/main.cpp
@@ -41,6 +41,12 @@
   #include <windows.h>
   typedef int pid_t;
   #define sleep(seconds) Sleep((seconds)*1000)
+#elif defined(__MINGW32__) || defined(__MINGW64__)
+  // mingw builds: POSIX libusb/unistd PLUS io.h/fcntl.h for binary stdin.
+  #include <io.h>
+  #include <fcntl.h>
+  #include <unistd.h>
+  #include <libusb-1.0/libusb.h>
 #elif defined(__ANDROID__)
   #include <libusb.h>
   #include <unistd.h>
@@ -233,7 +239,9 @@ int main(int argc, char **argv) {
     }
   }
 
-#if defined(_MSC_VER)
+#if defined(_WIN32)
+  // mingw/GCC defines _WIN32 but not _MSC_VER, yet still has _setmode — make
+  // stdin binary so a 0x1A/CRLF doesn't corrupt the length-prefixed PSDU stream.
   _setmode(_fileno(stdin), _O_BINARY);
 #endif
 
--- a/txdemo/stream_tx_demo/main.cpp
+++ b/txdemo/stream_tx_demo/main.cpp
@@ -46,6 +46,12 @@
   #include <process.h>
   typedef int pid_t;
   #define sleep(seconds) Sleep((seconds)*1000)
+#elif defined(__MINGW32__) || defined(__MINGW64__)
+  // mingw builds: POSIX libusb/unistd PLUS io.h/fcntl.h for binary stdin.
+  #include <io.h>
+  #include <fcntl.h>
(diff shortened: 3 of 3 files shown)


josephnef (commented):
Reviewed the diff against the source and CI. Both fixes are correct and minimal:

- **`_setmode` regate `_MSC_VER` → `_WIN32` + the `__MINGW32__/__MINGW64__` include branch** — correct. mingw needs `io.h`+`fcntl.h` for `_setmode`/`_O_BINARY` and `unistd.h` for `sleep()` (used at L214); `pid_t` and `<windows.h>` aren't referenced in this file, so their absence from the new branch is harmless. MSVC is unchanged — it defines both `_MSC_VER` and `_WIN32`, so it still takes the first include branch and still calls `_setmode`.
- **CMake vcpkg guard** — correct, and correctly placed *before* `project()` (required for `CMAKE_TOOLCHAIN_FILE` to take effect). CI's Windows job always sets `VCPKG_ROOT`, so there's no CI behavior change; the only effect is the unset-`VCPKG_ROOT` (mingw) case, which is the intent.

**One gap before merge:** `txdemo/stream_duplex_demo/main.cpp` has the byte-for-byte identical pattern — include block gated on `_MSC_VER` (L38) and `_setmode` gated on `_MSC_VER` (L202–204) — and it reads the same `<u32_le len><PSDU>` binary stream via `read_exact`. So it hits the exact first-frame text-mode truncation this PR fixes for `StreamTxDemo`. Worse: this PR's CMake change is what makes the duplex demo buildable under mingw in the first place, so as written this PR effectively ships a buildable-but-broken `StreamDuplexDemo` (mingw isn't in CI, so nothing catches it).

Please mirror the same `_WIN32` gate + mingw include branch into `stream_duplex_demo/main.cpp` so the two stdin-driven stream demos stay in lockstep.

For completeness: `precoder_demo/main.cpp` and `txdemo/main.cpp` also gate includes on `_MSC_VER`, but neither reads length-prefixed stdin nor calls `_setmode`, so they're unaffected — no change needed there.

GingerFluffyCat:
Good catch — fixed in 2880c0d.

`stream_duplex_demo/main.cpp` had the identical `_MSC_VER`-gated include block + `_setmode`, and since this PR's CMake guard is exactly what makes the duplex demo buildable under mingw, you're right that as-was it would ship buildable-but-broken. Mirrored the same change: `__MINGW32__/__MINGW64__` include branch + `_setmode` gated on `_WIN32`.

Verified on mingw-w64 (libusb 1.0.28) with the real RTL8812AU — `StreamDuplexDemo` now reads the full binary stream end to end:
```
<stream-duplex>tx #1 ok=1 psdu=269
...
<stream-duplex>tx #500 ok=1 psdu=269
<stream-duplex>tx EOF after 512 PSDUs
```
(pre-fix it died at `short read on stdin (76/269)` on frame 1).

Agreed on `precoder_demo/main.cpp` and `txdemo/main.cpp` — they gate includes on `_MSC_VER` too but neither reads length-prefixed stdin nor calls `_setmode`, so they're unaffected; left them as-is. The two stdin-driven stream demos are now in lockstep.

josephnef (approved):
Approving — the duplex-demo gap is resolved in 2880c0d (exact mirror of the StreamTxDemo fix; verified on hardware), and all 5 required CI checks are green. The two stdin-driven stream demos are now in lockstep.

PR #125, opened 2026-07-01
Split Jaguar1 into src/jaguar1/ + compile-time per-chip selection

josephnef:
## What

Two-part refactor of the source tree:

1. **Split `src/jaguar1/`** (commit `93b9756`) — moves the Jaguar1 driver files into `src/jaguar1/`, mirroring the existing `src/jaguar3/` layout, and decouples the two generations. Neither `#include`s the other; shared symbols (`MGN_RATE`, `Packet`/RX types, the TX-descriptor bit macros) moved to neutral `src/RateDefinitions.h`, `src/RxPacket.h`, `src/TxDescBits.h`. `RtlUsbAdapter` is now generation-agnostic (RX parse relocated into `RtlJaguarDevice`).

2. **Compile-time per-chip selection** (commit `562084b`) — four CMake options, all **default ON** so a plain `cmake -S . -B build` is unchanged:

   | option | chips |
   |---|---|
   | `DEVOURER_JAGUAR1` | 8812AU / 8811AU / 8821AU |
   | `DEVOURER_8814` | 8814AU (requires `DEVOURER_JAGUAR1`) |
   | `DEVOURER_JAGUAR3_8822C` | 8812CU / 8822CU |
   | `DEVOURER_JAGUAR3_8822E` | 8812EU / 8822EU |

   Invalid combos (none selected; 8814 without jaguar1) fail configure with `FATAL_ERROR`. Each group exports a PUBLIC `DEVOURER_HAVE_*` define; every site that references a dropped group's symbols is guarded with `#if defined(DEVOURER_HAVE_*)`. The factory returns `nullptr` (and logs) when a plugged chip's support isn't built; demos null-check and exit cleanly.

## Size win

8812AU-only `WiFiDriverDemo` drops from **2.58 MB → 1.03 MB** (−60%); 8822e-only is 0.84 MB.

## CI

New `build-configs` matrix (jaguar1-only, jaguar1+8814, 8822c-only, 8822e-only, jaguar3-only) + `reject-bad-configs` (the two FATAL cases), so a missing guard — which only fails to link when a chip's sources are dropped — is caught. Default matrix still covers all-chips-ON.

## Verification

- All configs: build + ctest green (fresh configure each)
- Cross-generation include hygiene: `grep jaguar3 src/jaguar1` / `grep jaguar1 src/jaguar3` both empty
- Hardware regression benchmark across the bench DUTs (results in a comment below)


josephnef:
## Hardware validation (bench, all-on build)

Re-ran the bench across every plugged DUT — both generations — to confirm the refactor is behavior-neutral. The all-on binary compiles the same code as before (every `DEVOURER_HAVE_*` defined), so this is a regression check on factory dispatch + init + RX/TX per chip.

**TX** (`tests/bench_init.py`, exec → first bulk-OUT) — all pass:

| Chip | Generation | TX init |
|---|---|---|
| RTL8812AU | Jaguar1 | ✅ |
| RTL8814AU | Jaguar1 (4T4R) | ✅ |
| RTL8821AU | Jaguar1 | ✅ |
| RTL8812CU (c812) | Jaguar3 rtl8822c | ✅ |
| RTL8812EU (a81a) | Jaguar3 rtl8822e | ✅ |

**RX** (controlled beacon inject → receive, `tests/refactor_rx_check.sh`) — all pass. Every DUT's factory dispatch selects the right HAL, init completes (monitor RX config + IQK), and the read+parse path delivers frames:

| Chip | RX result |
|---|---|
| RTL8812AU | ✅ ch6 (ambient) |
| RTL8821AU | ✅ ch6 (ambient) |
| RTL8812EU (a81a) | ✅ ch6 (58 canonical-beacon hits) |
| RTL8814AU | ✅ ch6 (43 hits) |
| RTL8812CU (c812) | ✅ ch36 (68 hits) — no-frame on ch6, a 2.4 GHz band quirk; rtl8822c's validated band is 5 GHz, and its RX path is shared with a81a which passes on ch6 |

No regression: factory dispatch, bring-up, TX, and RX all work on every chip across both generations.

josephnef:
## Correction + rtl8822c A/B (master vs branch)

My earlier comment leaned on beacon injection and hand-waved the rtl8822c ch6 RX miss. Redone properly with `tests/bench_init.py` (ambient RX, no injection), A/B against **master** on the identical chip/channel to isolate whether the refactor caused it:

| build | RTL8812CU (c812) RX ch6 | c812 RX ch36 (5 GHz) |
|---|---|---|
| master (pre-refactor, f542b06) | **0/2** | — |
| this branch | **1/2** | **400 frames captured** |

master receives nothing on ch6 either (0/2); the branch is equal-or-better (1/2), and pulls 400 frames on ch36. **rtl8822c did not degrade** — its weak 2.4 GHz/ch6 reception is pre-existing (marginal ambient RX on its non-primary band), not introduced by the split. The all-on binary compiles byte-identical code to master, so this is expected.

TX init (bench_init, exec → first bulk-OUT) still passes on all five DUTs: 8812AU, 8814AU, 8821AU, rtl8822c, rtl8822e.

Issue #133, opened 2026-07-01
Spatial × FEC — spatial diversity keeps corruption localized (sim + measure)


josephnef:
Part of the spatial-diversity roadmap #127. Depends on #B, #C.

## Goal
Establish and measure the strongest cross-layer link in the roadmap: **spatial
diversity keeps corruption localized**, which is the precondition the fused-FEC
sub-block-salvage layer needs to work.

## The argument
The fused-FEC stack (`docs/fused-fec.md`) only gains when corruption is localized
within a frame. Deep fades push corruption into the *frame-wide* regime where
sub-block salvage collapses (measured survivor fraction ~0.36; "neither inter-frame
scheme recovers"). Spatial diversity is precisely the tool that kills deep fades —
so it does not merely lower the corrupt-frame rate, it moves corruption *out of the
frame-wide regime and into the localized regime salvage exploits*. On RX, combining
before the soft demapper also yields cleaner soft information → better localization.

## Scope
- Extend the fused-FEC simulation to model per-chain fading + combining and show
  the shift in the survivor-fraction distribution (frame-wide → localized).
- Measure on-air: sub-block survival and delivered-block gain with vs without
  spatial diversity, at matched corrupt-frame rate.
- Quantify how diversity + soft-decode compose (both localize; are they additive?).

josephnef:
Sim done → PR #147. Diversity cuts SNR variance (deep fades are variance), moving frames out of the frame-wide-loss bin into the SBI-salvageable one — improving SBI's precondition. Modeled at matched mean SNR (isolating the variance effect): static/correlated barely helps (lost 0.31 vs single 0.33), mobile/decorrelated collapses it (0.12) and lifts salvageable-of-corrupt 0.59→0.86. Corruption localization is a mobility effect.

Issue #132, opened 2026-07-01
Close STBC on-air — resolve the 2-channel-alignment root cause


josephnef:
Part of the spatial-diversity roadmap #127. Depends on #A, #D.

## Goal
Close the **on-air STBC** path so devourer gains a diversity-order-2 transmit
knob (one stream, full spatial diversity, same MCS). STBC currently lives in the
radiotap path but does not produce correct output on air — the open root cause is
a 2-channel alignment problem.

## Scope
- Resolve the 2-channel-alignment root cause behind the current on-air STBC failure.
- Validate diversity order 2 on-air with an SDR: fade-margin / outage improvement
  vs single-chain and vs CSD (#D) at matched throughput.
- On the 8814, explore extending beyond the 2x1 Alamouti case where the extra
  chains allow it.

## Relation to #D
CSD and STBC are complementary transmit-diversity knobs; this issue adds the
higher-diversity-order option once the CSD knob (#D) is in place.

josephnef:
## Reframe (informed by #130 mobility + #128 audit)

Two different "STBC on-air" problems have been conflated here; the mobility result
makes the tractable one worth doing and clarifies the blocked one.

**(a) SDR-TX STBC DSP validation — the blocked, elusive one.** Transmitting an
Alamouti waveform from a 2-antenna B210 and having the Realtek chip decode it. The
DSP is bit-validated in a fork-RX self-loop, but over-air the chip reads *nothing*
for a real 2-antenna STBC frame while a same-signal 2× coherent frame decodes —
an unexplained 2-channel/chip-interaction problem (diagnosis stopped at
diminishing returns). This is **not** on the spatial-diversity critical path.

**(b) The chip's OWN STBC TX — tractable, and the one that matters here.** #128
found devourer already sets the STBC descriptor field from radiotap
(`DEVOURER_TX_RATE=MCS1/STBC`), and the chip encodes Alamouti across its two TX
antennas internally. This **sidesteps the SDR 2-channel-alignment blocker
entirely** — the chip aligns its own antennas — so it's directly measurable.

**Why it's now worth measuring.** #130 confirmed the RX-side story: spatial
diversity's payoff is under motion (it eliminates the deep-fade dropouts a fixed
antenna hits). STBC is the *transmit-side* dual — a moving drone (the TX) using
STBC across two antennas should present the ground receiver a link with fewer deep
fades, exactly as RX-MRC did for a moving receiver.

[...]

Issue #131, opened 2026-07-01
ANTSEL per-frame antenna-selection diversity probe (rescoped from CSD)


josephnef:
Part of the spatial-diversity roadmap #127. Depends on #A.

## Goal
Promote **cyclic shift diversity** from a validated primitive to a first-class
*robustness knob* on the TX (drone) side. CSD sprays the single video stream
across 2–4 chains with cyclic shifts — flat-fade transmit diversity at no rate
cost and no extra USB load. It is the lowest-risk entry point to TX spatial
diversity (STBC on-air is still unresolved; CSD is chip-validated).

## Scope
- Expose CSD as a selectable per-frame mode (like rate/BW/STBC in the radiotap path)
  so it can be driven per-packet and per-layer.
- Measure the fade-margin improvement on-air (SDR ground truth) vs single-chain TX
  across 2 and (on 8814) 3–4 chains.
- Characterize the energy cost (extra chains' baseline draw) as input to #H.

## Reality check
Diversity gain bounded by antenna decorrelation (#B).

josephnef:
## Reframe from the #128 HAL audit

The audit (#128) found that **CSD is not exposed in devourer's Realtek chip HAL**. The "CSD chip-validated" result is from the SDR waveform builder (`gr-ieee802-11`/`sdr2wifi`, transmitting from a B210), where cyclic shifts are applied in host DSP before the IFFT. The Realtek chip applies cyclic shifts **internally in the PHY** when transmitting a multi-stream frame; there is no register knob for host-driven, single-stream CSD.

So this issue as written ("CSD as a validated diversity knob") has **no chip path today**. Options:
1. **Fold into #132** — on the chip, STBC *is* the transmit-diversity mode; treat STBC as the single Tx-diversity knob and close this issue.
2. **Rescope to antenna-selection** — the `ANTSEL_A..D` TX-descriptor fields exist but are never written; probe whether per-frame antenna selection gives a cheap diversity mode the descriptor already supports.
3. **Keep CSD as an SDR-only technique** — relevant to the SDR-TX path, out of scope for the chip driver.

Needs a decision before proceeding. Leaning (1)+(2): STBC is the chip's real Tx-diversity route, with an `ANTSEL` probe as a low-cost side experiment.

josephnef:
## Decision (from #128 audit)

[...]

Issue #130, opened 2026-07-01
RX-side in-chip MRC on an 8814 ground station


josephnef:
Part of the spatial-diversity roadmap #127. Depends on #A (per-chain capability).

## Goal
Exploit a 4T4R chip (8814) as a **ground receiver** doing in-chip
maximal-ratio combining across its RF chains — the single biggest USB-free win in
the roadmap. One 4-chain adapter partially replaces wfb-ng's multi-adapter
diversity array, entirely in-chip, adding zero USB load (still one stream out).

## Scope
- Confirm the chip performs MRC across active RX chains and quantify the SNR gain
  vs a single chain, at matched operating point.
- Compare against the wfb-ng multi-adapter diversity baseline (several separate
  dongles, combined in software) for delivered-block rate at equal antenna count.
- Feed the cleaner combined signal into the fused-FEC RX path and measure the
  effect on corrupt-frame rate and sub-block survival (links to #F).

## Reality check
Gains are bounded by antenna decorrelation — qualify the layout with #B first.

josephnef:
## Hardware validation on two RTL8814AU dongles

Ran the harness on both plugged 8814s (CF-938AC, 2 external dipoles; CF-960AC,
4 on-PCB internal antennas) — which enumerate with an **identical
VID:PID:serial**, so this also validated the new `DEVOURER_USB_BUS`/`PORT`
topology selector. One controlled beacon TX (8812AU, ch6), each 8814 as RX,
~7.5k canonical-SA frames each.

**What worked end-to-end:** the selector targeted the two identical devices by
bus/port (distinct per-chain fingerprints confirm different hardware was
opened), the opt-in 4-path emission delivered all four chains, and the analyser
ran clean.

**Two findings that hardened the tool:**

1. **A static bench beacon can't measure decorrelation.** Per-chain std came out
   at ~0.15–1.8 units — the near-field beacon is a static channel with almost no
   fading, so envelope correlation just tracks quantisation noise. The harness
   now reports **INCONCLUSIVE** below `MIN_FADING_STD` instead of a bogus
   GOOD/MODERATE. A valid decorrelation number needs a *fading* stimulus:
   motion, varied multipath, or a mobile link. This is the key methodology note
   for the on-air validation step.
2. **8814 path-C RSSI rails at a constant 82.00** on *both* devices (std ~0.02) —
   a per-path reading (`gain_trsw_cd`) that carries no signal here; now flagged
[...]

PR #140, opened 2026-07-02
Add antenna-decorrelation measurement harness (spatial-diversity #129)

josephnef:
## Summary

Foundation tooling for the spatial-diversity roadmap (#127) — the effort to
exploit the RF chains a single-stream video link leaves idle (1 on a 2T2R part,
up to 3 on a 4T4R 8814) as **diversity/robustness**, the third axis alongside
frequency hopping (`docs/frequency-hopping.md`) and the fused FEC
(`docs/fused-fec.md`).

Closes the measurement half of #129. Every downstream diversity claim (RX MRC
#130, `ANTSEL` probe #131, STBC #132, spatial×FEC #133) rests on the chains'
antennas being **decorrelated** — on a small airframe they may not be. This adds
the harness that *measures* the decorrelation a given antenna layout actually
delivers, so gains are reported as measured rather than theoretical.

## Changes

- **`demo/main.cpp`** — opt-in `DEVOURER_RX_ALLPATHS=1` emits a
  `<devourer-rxpath>` line per canonical-SA frame carrying all four RX chains
  (A,B,C,D) of `rssi`/`snr`/`evm`. The library already populates paths C/D on the
  8814AU; the demo previously printed only A,B. On a **distinct tag** so the
  canonical `<devourer-stream>`/`<devourer-body>` two-path format (and its regex
  consumers) is untouched. Paths C/D read 0 on 2T2R parts.
- **`tests/antenna_decorrelation.py`** — parses the per-chain metrics and reports
  per-chain level stats, the pairwise **envelope correlation matrix** (Pearson on
  linear amplitude — the diversity-relevant correlation, not on dB), and a
  **combining analysis**: selection- and maximal-ratio-combining gain over the
  best single chain at the 10 % and 1 % outage points vs the theoretical
  10·log₁₀(N) array gain. Verdict is keyed off the worst pairwise |ρ| (ρ<0.7 =
  "diversity still useful"). A hardware-independent `--self-test` validates the
  estimator and combining math against synthesised correlated Rayleigh fading.
- **`tests/run_antenna_decorrelation.sh`** — build + capture + analyse runner
  (exact-comm cleanup trap, uv venv, sudo for the USB claim); optionally starts
  the beacon TX (`TX_PID`).
- **`CLAUDE.md`** — document `DEVOURER_RX_ALLPATHS`.

## Why measure against the canonical beacon

The library only surfaces per-chain metrics for the canonical SA
`57:42:75:05:d6:00`, so the harness measures a steady, known source (the standard
two-adapter bench setup) — the right controlled condition for a correlation
estimate, rather than noisy ambient traffic.

## Testing

- `uv run python tests/antenna_decorrelation.py --self-test` — green. Recovers a
  known envelope correlation to ±0.05 (synthesising with field-corr = √ρ_env,
  since ρ_env = |ρ_field|² for Rayleigh), confirms independent 4-chain MRC mean
  gain ≈ 10·log₁₀(4), and that correlated chains yield less low-outage gain than
  independent ones.
- `WiFiDriverDemo` builds with the new emission.
- **Not yet** run on hardware — the on-air 8814 capture across antenna
  spacings/polarisations is the field-validation step, tracked in #129/#130.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -137,6 +137,12 @@ Both `WiFiDriverDemo` and `WiFiDriverTxDemo` honour:
   `0x8813` for RTL8814AU); without it, the demo iterates every Realtek PID.
 - `DEVOURER_VID=0xNNNN` — override VID (default `0x0bda`); needed for
   OEM-rebadged dongles like the TP-Link Archer T2U Plus (`2357:0120`).
+- `DEVOURER_USB_BUS=N` (+ optional `DEVOURER_USB_PORT=a.b.c`) — select a device
+  by USB topology instead of first-match VID:PID. Needed when two adapters share
+  one VID:PID **and serial** and can't otherwise be told apart (two RTL8814AU
+  dongles, e.g. CF-938AC vs CF-960AC). `DEVOURER_USB_PORT` is the dotted libusb
+  port path (sysfs `devpath` / `lsusb -t`). Unset = the normal VID:PID open loop.
+  Used by `tests/compare_8814_decorrelation.sh`.
 - `DEVOURER_CHANNEL=N` — override monitor channel.
 - `DEVOURER_SKIP_RESET=1` — skip `libusb_reset_device` before claim; useful
   when picking up a chip whose firmware is already running.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -150,6 +150,13 @@ Both `WiFiDriverDemo` and `WiFiDriverTxDemo` honour:
   output. This is the entry point for the fused-FEC sub-block-salvage layer
   (see `docs/fused-fec.md`); opt-in, since a body with a corrupt tail is the
   worst-case input for an IP-stack consumer that didn't ask for it.
+- `DEVOURER_RX_ALLPATHS=1` — emit a `<devourer-rxpath>` line per canonical-SA
+  frame carrying all four RX chains (A,B,C,D) of per-stream `rssi`/`snr`/`evm`,
+  where the canonical `<devourer-stream>`/`<devourer-body>` lines surface only
+  A,B. Paths C/D are non-zero only on the 8814AU (4T4R); 0 on 2T2R parts.
+  Opt-in and on a distinct tag so the two-path format its regex consumers key on
+  is untouched. Consumed by `tests/antenna_decorrelation.py` to measure
+  inter-chain envelope correlation and realised diversity gain.
 - `DEVOURER_USB_DEBUG=1` — raise libusb log level from the default WARNING to
   DEBUG (produces ~7 MB per 15 s — has filled `/tmp` mid-capture and adds
   0.5-0.8 s to init even with stderr discarded). `DEVOURER_USB_QUIET` is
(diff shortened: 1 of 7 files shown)


josephnef:
## Hardware validation on two RTL8814AU dongles

Ran the harness on both plugged 8814s (CF-938AC, 2 external dipoles; CF-960AC,
4 on-PCB internal antennas) — which enumerate with an **identical
VID:PID:serial**, so this also validated the new `DEVOURER_USB_BUS`/`PORT`
topology selector. One controlled beacon TX (8812AU, ch6), each 8814 as RX,
~7.5k canonical-SA frames each.

**What worked end-to-end:** the selector targeted the two identical devices by
bus/port (distinct per-chain fingerprints confirm different hardware was
opened), the opt-in 4-path emission delivered all four chains, and the analyser
ran clean.

**Two findings that hardened the tool:**

1. **A static bench beacon can't measure decorrelation.** Per-chain std came out
   at ~0.15–1.8 units — the near-field beacon is a static channel with almost no
   fading, so envelope correlation just tracks quantisation noise. The harness
   now reports **INCONCLUSIVE** below `MIN_FADING_STD` instead of a bogus
   GOOD/MODERATE. A valid decorrelation number needs a *fading* stimulus:
   motion, varied multipath, or a mobile link. This is the key methodology note
   for the on-air validation step.
2. **8814 path-C RSSI rails at a constant 82.00** on *both* devices (std ~0.02) —
   a per-path reading (`gain_trsw_cd`) that carries no signal here; now flagged
   as railed. Relevant to #130 (MRC wants valid per-path level/SNR).

**Per-chain level fingerprint still distinguishes the adapters** (means, static
channel): CF-960AC A/B/D ≈ 74–78; CF-938AC shows a wider spread (B≈60, D≈68),
consistent with only 2 driven external antennas vs 4 internal.

Next: repeat with a fading stimulus (walk the beacon / rotate the adapter /
outdoor multipath) to get a real correlation figure per antenna layout — that's
the field step feeding #130.

Pushed as `dad0469` on this branch.

josephnef:
## First valid decorrelation measurement: two 8814s compared (rotation stimulus)

Since a static bench beacon can't excite fading, the RX adapter was **rotated /
tilted continuously** through the capture window to sweep each antenna's pattern
against the fixed beacon TX (8812AU, ch6). This measures **pattern/orientation
decorrelation** under strong LOS — a proxy for (not identical to) multipath
envelope correlation, but the *relative* comparison between two adapters on the
same bench is robust.

Both dongles are the **same RTL8814AU silicon**, differing only in antenna
front-end:

| Adapter | Antennas | frames | per-chain std | worst pairwise \|ρ\| | MRC gain @1% outage | Verdict |
|---|---|---|---|---|---|---|
| **CF-960AC** (round AP) | 4 on-PCB internal | 14348 | ~4.0–4.7 | **0.19** (all pairs <0.2) | ~13.9 dB | GOOD |
| **CF-938AC** (stick) | 2 external dipoles | 6546 | ~3.2–4.2 | **0.64** (A↔C coupled) | ~11.1 dB | MODERATE |

**Interpretation.** The CF-938AC drives 4 RF chains from only **2 physical
antennas**, so chains A and C are strongly coupled (ρ≈0.64) — effectively ~2
independent diversity branches out of 4. The CF-960AC's genuine 4-antenna array
decorrelates well on every pair (<0.2) — closer to 4 usable branches. So for the
RX-MRC goal (this issue), **the 4-internal-antenna adapter is materially better
despite identical silicon**; the 2-external-antenna stick gives roughly half the
effective diversity order.

**Incidental finding.** The 8814 path-C RSSI that railed at a constant 82.00
under the static beacon varies normally once the adapter moves — it was
AGC-pinned in strong static LOS, not dead hardware.

**Caveat / next.** Rotation-on-a-bench is angular/pattern decorrelation under
strong LOS. The gold-standard figure is a genuine NLOS multipath capture — the
plan is to relocate the beacon TX to another room/balcony (a separate host
running the beacon) and re-measure. The relative ordering (960AC > 938AC) is not
expected to change.

Tooling: `tests/compare_8814_decorrelation.sh` (auto-detects every 8814 by
bus/port, one beacon TX) + the `ROTATE=1` mode of
`tests/run_antenna_decorrelation.sh`. Runner reliability fixes pushed (33f54a3).

PR #251, opened 2026-07-11
docs: multi-ap-cellular — coordinated cells, make-before-break handover, roaming robot UEs (+ animation)

josephnef:
Conceptual map of what the time-distribution primitives (#242–#250) build at facility scale — three applications, LTE analogies explicit, every per-link claim linked to its bench numbers.

## `docs/multi-ap-cellular.md`

1. **ICIC over the existing Ethernet backhaul.** The classic Wi-Fi obstacle — cells that can't agree on *when* — is gone: the shared clock makes a slot boundary a real network-wide instant with µs-grade guards. Coordination rides the same wire as PTP; centralized and peer (X2-style) scheduler shapes; orthogonality in time, frequency (~ms retune), or a reuse pattern — a policy, not a driver feature.

2. **Make-before-break handover.** The ~100 ms roam decomposed and dissolved: no time re-acquisition (the network clock is valid across cells), no discovery scan (passive neighbor beacons + ms retune in idle slots), **network-side decision** (every AP hardware-timestamps and RSSI-tags every uplink it hears — LTE measurement reports without client cooperation), handover = a schedule-ownership update between two slots. Honest residue named: security-context motion + the ms retune, both inside a slot guard.

3. **Robots as roaming UEs.** Camp / scheduled DL+UL slots / per-AP TA loop tracking motion / handover as above — ending on the point that matters for robotics: **determinism becomes a schedule property, not a contention statistic.**

A boundary paragraph keeps the doc honest as it ages: per-link primitives are bench-proven with numbers in the linked docs; the multi-cell layer is ordinary distributed software on the wired side — nothing blocked on the radio.

## The animation

`tools/warehouse_cellular_gif.py` → `docs/img/warehouse_cellular.gif` (house monitor style): a floor plan with two overlapping cells and a driving robot, stacked over the network-wide slot ruler. Three acts: uncoordinated collisions in the overlap → the scheduler splitting the grid orthogonally over the backhaul → the robot crossing cells with RSSI crossover, `>> HO: robot → B` on the bus, and its DL/UL slots recoloring between one slot and the next — against the ghost bar *"ordinary Wi-Fi roam: ~100 ms dead"* vs **"handover gap here: 0 slots"**.

Cross-linked from `time-distribution.md` and the rf-primer's where-to-next. Doc + tools only; `ctest` 17/17.


josephnef:
Review feedback addressed: (1) floor geometry fixed — the coverage circles no longer cross the box's top edge and the dead canvas under the ruler is gone (920×640); (2) handover semantics corrected — no raw-RSSI decisions: the doc now spells out the LTE/5G measurement discipline (filtered estimates, hysteresis + time-to-trigger, the A3 event) with the fast-fading ping-pong rationale, and the animation's bars/captions are relabelled as the scheduler's *filtered* link-quality estimate crossing with margin+TTT.

josephnef:
Added the missing idea (review feedback): **soft handover** — through the crossing window the robot holds slots in *both* cells and duplicates its uplink across them with a ~ms retune between slots (both APs' copies race over the backhaul, dedup by sequence). Doc gets a dedicated paragraph (CDMA soft handoff / 5G PDCP-duplication analogy, and why standard Wi-Fi can't express it while public cellular gates it behind heavyweight dual-connectivity); the animation's act 3 is now SOFT HANDOVER: dual links, "dup"-tagged slot pairs in both cells' territory, then collapse onto B.

PR #289, opened 2026-07-13
jaguar3(8822e): port kernel eFEM GPIO pinmux — RX path B was antenna-less under the static DPDT route

gilankpam:
## Summary

The 8822E front-end fix (#268 / b5a6df7) routes the DPDT antenna transfer switch with a static two-bit write: `REG_LED_CFG 0x4c[24]` set, `[22]` clear. That made MCS4+ TX work — but it parks the switch in a TX-favoring position that **disconnects RX path B's antenna on every 8812EU**: chain-B pwdb pins at the noise floor (~10 raw / −99 dBm) no matter which antenna is fitted.

Nobody saw it because every RX validation to date — including #268's full-duplex proof and the 0x41e8 desense recheck — counts **total frames**, and chain A alone carries the stream. The loss is real, though: a dead chain B forfeits MRC diversity (3–5 dB, more in NLOS multipath).

This PR replaces the static write with a port of the kernel's actual mechanism: `_efem_pinmux_config` (rtl8822e_halinit.c, rfe 21–24) → halmac `pinmux_set_func_8822e` for `RFE_CTRL_3/5/7/8/9/11`, transcribed entry-for-entry from `halmac_gpio_8822e.c`. GPIO13 becomes `WL_DPDT_SEL` driven by the RFE engine's RFE_CTRL_9, so the transfer switch follows TX/RX **in hardware**: PA on TX, both LNAs on RX — kernel parity.

## Evidence

Two-adapter bench (BL-M8812EU2 ↔ 8812EU, ch149, 20 MHz, MCS5 at 6 Mbps offered unless noted, txagc 30 near-field, per-chain rssi/snr from the phystatus via a per-chain-reporting RX tool). Same card, back-to-back bring-ups, only the DPDT route changing:

| DPDT route | RX chain A | RX chain B | TX @ MCS7 (10M offered) |
|---|---|---|---|
| static `0x4c[24]`+`[22]`-clear (current) | −68 dBm / 27 dB | **−99 dBm / dead** | 9.9M goodput, clean |
| `0x4c[24]` only (halmac `WL_DPDT_SEL` bit alone) | −68 / 27 | −77 / 19 | **airs nothing** |
| eFEM pinmux (this PR) | −70 / 26 | **−77 / 19** | **9.7M goodput, clean** |

- The bit-level bisect that found it: with the full merge in place, skipping only the DPDT block brings chain B back; skipping only the 1SS TX mapping does not. Within the block, keeping `[22]` untouched revives chain B (and breaks TX — hence neither bit combination is sufficient; the pin needs its function claim).
- Read-back before the write shows `0x4c[22]` SET from reset (`0x0062e282`), i.e. GPIO13 parked as BT_WAKE — which is also why `[24]`-only TX airs nothing: no one has claimed the pin for the RFE engine. The missing piece is the pin-function claim (`0x42[1]` on GPIO13 + `0x43[3]/[4]` fixups), which only the full pinmux walk performs.
- Validated with both link ends running this port **simultaneously**: TX MCS7 9.7M / MCS5 6.0M at 0.0% loss / MCS0 clean, while RX reports both chains live (−70/−77 dBm, snr 26/19) — the combination that was previously impossible.
- 0x41e8 history (quirks doc updated): the register is exonerated by direct A/B; the "RX desense" it was blamed for was this antenna disconnection all along, and both prior rechecks were blind to it because they counted total frames.

## Implementation notes

- `RtlJaguar3Device::efem_pinmux_8822e()`: per-pin priority-list walk (halmac semantics — disable every claim above the target: `field := ~val & msk`; enable the target: `field := val & msk`; stop) + the WL/BT ownership fixups, as data tables. Kernel call order preserved (RFE_CTRL_3, 5, 7, 8, 9, 11).
- Runs post-coex in `InitWrite` — coex `GPIO_MUXCFG` writes would mask it; same ordering lesson as the existing RFE pad block. The `PAD_CTRL1[29:28]` post-coex re-assert is kept.
- The old routes remain selectable for board A/B: `DEVOURER_DPDT_MODE=legacy|bit24|skip` (default `efem`).
- `docs/8822e-quirks.md` updated: the pin-mux section now describes the hardware-controlled route and its A/B matrix, and adds the honest moral — per-chain RSSI is the only trustworthy RX-health metric on a 2T2R part.

## Suggested follow-up (not in this PR)

`GetActiveRxPaths`/per-chain health could gate CI-style validation so a single-chain regression can't hide behind total-frame counts again.

### DIFF
--- a/src/jaguar3/RtlJaguar3Device.cpp
+++ b/src/jaguar3/RtlJaguar3Device.cpp
@@ -434,6 +434,142 @@ void RtlJaguar3Device::coex_runtime_loop() {
   }
 }
 
+namespace {
+
+/* --- 8822E eFEM GPIO pin-function tables --------------------------------
+ * Transcribed from halmac_gpio_8822e.c PINMUX_LIST_GPIO{2,3,4,13,14}_8822E /
+ * PINMUX_LIST_EECS_8822E, truncated at the target function. halmac's
+ * pinmux_switch_8822e walks a pin's priority list top-down: every claim
+ * ABOVE the target is disabled (field := ~val & msk), the target itself is
+ * enabled (field := val & msk), and the walk stops — entries below the
+ * target are never touched. Byte-register RMW throughout. */
+struct EfemEnt {
+  uint8_t off; /* MAC byte register */
+  uint8_t msk;
+  uint8_t val; /* the claim's ENABLE value; disable writes ~val & msk */
+};
+struct EfemFix { uint8_t off; uint8_t bit; bool set; };
+struct EfemFunc {
+  const char *name;
+  const EfemEnt *ents;
+  size_t n_ents;        /* last entry is the target */
+  const EfemFix *fixes; /* post-switch WL/BT ownership fixups */
+  size_t n_fixes;
+};
+
+constexpr uint8_t B0 = 0x01, B1 = 0x02, B2 = 0x04, B3 = 0x08, B4 = 0x10,
+                  B5 = 0x20, B6 = 0x40, B7 = 0x80;
+
+/* RFE_CTRL_3 = GPIO3 -> WL_LNAON_SEL */
+constexpr EfemEnt kEntsRfe3[] = {
+    {0x66, B2, B2},      /* BT_GPIO3 */
+    {0x4F, B6, B6},      /* BT_ANT_SW_3 */
+    {0x41, B1, 0},       /* WL_PRI (BT_PTA) */
+    {0x41, B2, B2},      /* BT_PRI (WL_PTA) */
+    {0x40, B1 | B0, B0}, /* WLMAC_DBG */
+    {0x40, B1 | B0, B1}, /* WLPHY_DBG */
+    {0x40, B1 | B0, B1 | B0}, /* BT_DBG */
+    {0x4F, B2, B2},      /* SW LNAON_SEL */
+    {0x42, B0, B0},      /* BT_RFE_CTRL_5 (BT_LNAON_SEL) */
+    {0x42, B0, B0},      /* RFE_CTRL_3 (WL_LNAON_SEL) — target */
+};
+constexpr EfemFix kFixRfe3[] = {{0x43, B7, false}, {0x67, B4, true}};
+
+/* RFE_CTRL_5 = GPIO14 -> WLPHY_RFE_CTRL2GPIO */
+constexpr EfemEnt kEntsRfe5[] = {
+    {0x4E, B6, B6}, /* UART_WAKE (GPIO13_14_WL_CTRL_EN) */
+    {0x4F, B6, B6}, /* BT_ANT_SW_1 */
+    {0x40, B2, B2}, /* RFE_CTRL_5_4_5 — target */
+};
+constexpr EfemFix kFixRfe5[] = {{0x3F, B2, true}};
+
+/* RFE_CTRL_7 = EECS pad -> WLPHY_RFE_CTRL2GPIO_2 */
+constexpr EfemEnt kEntsRfe7[] = {
+    {0x67, B1, B1}, /* BT_GPIO17 */
+    {0x40, B3, B3}, /* RFE_CTRL_7 — target */
+};
+
+/* RFE_CTRL_8 = GPIO4 -> WL_DPDT_SEL */
+constexpr EfemEnt kEntsRfe8[] = {
+    {0x66, B4, B4},      /* BT_SPI_D0 */
+    {0x42, B3, B3},      /* WL_SPI_D0 */
+    {0x67, B0, B0},      /* BT_JTAG_TRST */
+    {0x65, B7, B7},      /* WL_JTAG_TRST */
+    {0x73, B3, B3},      /* DBG_GNT_WL */
+    {0x40, B1 | B0, B0}, /* WLMAC_DBG */
+    {0x40, B1 | B0, B1}, /* WLPHY_DBG */
+    {0x40, B1 | B0, B1 | B0}, /* BT_DBG */
+    {0x4E, B7, B7},      /* ANT_SWB (SW_DPDT_SEL) */
+    {0x42, B1, B1},      /* BT_RFE_CTRL_0 (BT_DPDT_SEL) */
+    {0x42, B1, B1},      /* WL_RFE_CTRL_8 (WL_DPDT_SEL) — target */
+};
+constexpr EfemFix kFixRfe8[] = {{0x43, B2, true}, {0x4F, B0, true}};
+
+/* RFE_CTRL_9 = GPIO13 -> WL_DPDT_SEL — the DPDT transfer switch itself. */
+constexpr EfemEnt kEntsRfe9[] = {
+    {0x4E, B6, B6}, /* BT_WAKE (GPIO13_14_WL_CTRL_EN — 0x4c[22]) */
+    {0x4F, B6, B6}, /* BT_ANT_SW_0 */
+    {0x4E, B7, B7}, /* ANT_SW_GPIO13 (SW_DPDT_SEL) */
+    {0x42, B1, B1}, /* BT_RFE_CTRL_1 (BT_DPDT_SEL) */
+    {0x42, B1, B1}, /* WL_RFE_CTRL_9 (WL_DPDT_SEL) — target */
+};
+constexpr EfemFix kFixRfe9[] = {
+    {0x43, B3, false}, {0x43, B4, true}, {0x4F, B0, true}};
+
+/* RFE_CTRL_11 = GPIO2 -> WLPHY_RFE_CTRL2GPIO */
+constexpr EfemEnt kEntsRfe11[] = {
+    {0x66, B2, B2},      /* BT_GPIO2 */
+    {0x4F, B6, B6},      /* BT_ANT_SW_2 */
+    {0x41, B1, 0},       /* WL_STATE (BT_PTA) */
+    {0x41, B2, B2},      /* BT_STATE (WL_PTA) */
+    {0x40, B1 | B0, B0}, /* WLMAC_DBG */
+    {0x40, B1 | B0, B1}, /* WLPHY_DBG */
+    {0x40, B1 | B0, B1 | B0}, /* BT_DBG */
+    {0x40, B2, B2},      /* RFE_CTRL_11_4_5 — target */
+};
+constexpr EfemFix kFixRfe11[] = {{0x3F, B0, false}};
+
+/* Kernel order: rtw_halmac_rfe_ctrl_cfg(28..33) = RFE_CTRL_3,5,7,8,9,11. */
+constexpr EfemFunc kEfemFuncs[] = {
+    {"RFE_CTRL_3/GPIO3", kEntsRfe3, std::size(kEntsRfe3), kFixRfe3,
+     std::size(kFixRfe3)},
+    {"RFE_CTRL_5/GPIO14", kEntsRfe5, std::size(kEntsRfe5), kFixRfe5,
+     std::size(kFixRfe5)},
+    {"RFE_CTRL_7/EECS", kEntsRfe7, std::size(kEntsRfe7), nullptr, 0},
+    {"RFE_CTRL_8/GPIO4", kEntsRfe8, std::size(kEntsRfe8), kFixRfe8,
+     std::size(kFixRfe8)},
+    {"RFE_CTRL_9/GPIO13", kEntsRfe9, std::size(kEntsRfe9), kFixRfe9,
+     std::size(kFixRfe9)},
+    {"RFE_CTRL_11/GPIO2", kEntsRfe11, std::size(kEntsRfe11), kFixRfe11,
+     std::size(kFixRfe11)},
+};
+
+} /* namespace */
+
+void RtlJaguar3Device::efem_pinmux_8822e() {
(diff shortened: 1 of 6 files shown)


josephnef (commented):
Reviewed with the vendor tree open and the local bench running — details per finding inline.

**Verified:**
- **Vendor-source parity is exact.** Diffed every table against `reference/rtl88x2eu/.../halmac_gpio_8822e.c`: all 6 entry lists match entry-for-entry, correctly truncated at the first `switch_func` match per `pinmux_switch_8822e` (GPIO2/14 stop at the `_4_5` variant, GPIO4/13 at `WL_RFE_CTRL_8/9`); walk semantics (byte RMW, `~val & msk` disable, enable target, stop) and all post-switch fixups match, address arithmetic checked (`0x43` = `REG_GPIO_MUXCFG+3`, `0x4F` = `REG_LED_CFG+3` — bit0 ≙ `0x4c[24]`, so the eFEM path still sets the old TX bit via the proper claim; `0x67` = `REG_PAD_CTRL1+3`, `0x3F` = `REG_GPIO_MUXCFG_2`). Call order 28..33 → RFE_CTRL_3,5,7,8,9,11 matches `_efem_pinmux_config` / `_gpio_to_func_for_rfe_ctrl`.
- **Build clean, ctest 20/20.**
- **On-air TX A/B on this rig** (`regress.py --modes devourer:devourer`, 8812EU TX → 8812AU RX, ch36, 10 s cells, two RX cells per mode): legacy **270/200** hits, efem **283/200**, bit24 **90/10** — efem at TX parity with legacy, bit24 collapsed. Independently confirms the TX side of your matrix. The RX chain-B half isn't locally verifiable here (single 8812EU on the rig), but the mechanism is vendor-exact and the bench evidence is coherent.

**Asks before merge:** the env-read convention violation (inline) and preferably the one-line `rfe_type` 21–24 gate (inline). The `Init`-vs-`InitWrite` asymmetry can be a follow-up bench check alongside your suggested `GetActiveRxPaths` CI gating.

Good work — the bisect narrative is convincing and the legacy A/B escape hatch is the right safety net.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -656,26 +792,46 @@ void RtlJaguar3Device::InitWrite(SelectedChannel channel) {
...
+     * RFE_CTRL_9 (WL_DPDT_SEL) so the switch follows TX/RX like the
+     * kernel: PA on TX, BOTH LNAs on RX. Runs post-coex (coex GPIO_MUXCFG
+     * writes would mask it — same ordering lesson as the pre-coex RFE
+     * block above).
+     * DEVOURER_DPDT_MODE knob for A/B: efem (default) | legacy (b5a6df7
+     * write) | bit24 ([24] only) | skip. */
+    const char *dpdt_env = std::getenv("DEVOURER_DPDT_MODE");
+    std::string dpdt_mode = dpdt_env ? dpdt_env : "efem";
+    if (std::getenv("DEVOURER_SKIP_DPDT") != nullptr)
+      dpdt_mode = "skip";
```
josephnef: **Convention: the library reads no environment** (CLAUDE.md **Configuration**). Every existing knob — including `apply_replay_wseq` just below, which reads `_cfg.debug.replay_wseq` — routes through `DeviceConfig` with the env mapping in `examples/common/env_config.cpp`. Please move this to a `debug` (or `tuning`) field with the usual `env:` doc tag and map `DEVOURER_DPDT_MODE` there.

Also `DEVOURER_SKIP_DPDT` is an undocumented second spelling of `mode=skip` (it appears nowhere in the docs or `DeviceConfig`) — drop it or fold it into the documented knob.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -656,26 +792,46 @@ void RtlJaguar3Device::InitWrite(SelectedChannel channel) {
...
+    if (dpdt_mode == "efem") {
+      efem_pinmux_8822e();
```
josephnef: The kernel gates `_efem_pinmux_config` on **rfe_type 21–24** (`rtl8822e_halinit.c`); this applies unconditionally on every 8822E. The gate is cheap kernel parity: `_phy_ctx.rfe_type` is already read from efuse, and the existing RFE pad block (`HalJaguar3.cpp`, `config_rfe`) already gates on 21..24. All benched boards are rfe 21 (and devourer defaults efuse 0/0xff → 21), so low practical risk today — but a non-eFEM 8822E board would get GPIOs claimed for an RFE engine its front end doesn't wire.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -434,6 +434,142 @@ void RtlJaguar3Device::coex_runtime_loop() {
...
+    for (size_t i = 0; i < f.n_fixes; ++i) {
+      const EfemFix &x = f.fixes[i];
+      uint8_t v = _device.rtw_read8(x.off);
+      v = x.set ? static_cast<uint8_t>(v | x.bit)
+                : static_cast<uint8_t>(v & ~x.bit);
+      _device.rtw_write8(x.off, v);
+    }
+  }
+  _logger->warn("Jaguar3(8822e): eFEM pin-mux applied (RFE_CTRL 3/5/7/8/9/11 "
+                "-> RFE engine; DPDT under hardware TX/RX control)");
```
josephnef: `warn` for the default/successful path — `info` is the house style for applied-config lines (the `warn`s on the legacy/skip branches are right, those *should* be loud).

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -656,26 +792,46 @@ void RtlJaguar3Device::InitWrite(SelectedChannel channel) {
...
-      _device.rtw_write<uint32_t>(0x0064, v64 | 0x30000000u);
-    _logger->info("Jaguar3(8822e): DPDT/pad pin-mux applied (0x4c[24], "
-                  "0x64[29:28])");
+    /* eFEM GPIO pin-function routing (rfe 21-24) — port of the kernel's
+     * _efem_pinmux_config (rtl8822e_halinit.c) → halmac
+     * pinmux_set_func_8822e for RFE_CTRL_3/5/7/8/9/11. Replaces the
+     * improvised b5a6df7 DPDT write, which took two 0x4c bits in isolation
+     * (set [24], clear [22]) and thereby parked the DPDT antenna transfer
+     * switch in a static TX-favoring position: TX MCS4+ worked, but RX
+     * path B lost its antenna on every 8812EU (bench 2026-07-14: chain-B
```
josephnef: Nit: the hardcoded bench date will rot; the fact stands without it (git is the changelog).

src/jaguar3/RtlJaguar3Device.h
```diff
@@ -281,6 +281,12 @@ class RtlJaguar3Device : public IRtlDevice {
    * captured kernel register-write stream verbatim at the end of InitWrite —
    * the hardware-diff lever (same as Jaguar2's; found the 8822B RF18 bug). */
   void apply_replay_wseq();
+  /* 8822E eFEM (rfe 21-24) GPIO pin-function routing — kernel-parity port of
+   * _efem_pinmux_config/pinmux_set_func_8822e for RFE_CTRL_3/5/7/8/9/11.
+   * Routes the DPDT antenna transfer switch to the RFE engine (hardware
+   * TX/RX switching) instead of the b5a6df7 static write that deafened RX
+   * path B. Called from InitWrite post-coex; see the call site. */
+  void efem_pinmux_8822e();
```
josephnef, line 289: The kernel runs this in `rtl8822e_init_misc` during **general** HAL init (both directions); this port only touches `InitWrite`, so RX-only `Init` sessions keep the reset-default pins (GPIO13 parked as BT_WAKE per your own read-back). Worth one bench check: is chain B alive in an RX-only session? If not, the pinmux belongs in the shared bring-up; if it is, a one-line comment on the asymmetry would prevent the next reader from "fixing" it. Fine as a follow-up alongside your suggested per-chain CI gating.

josephnef (changes requested):
Marking changes requested per the inline review (https://github.com/OpenIPC/devourer/pull/289#pullrequestreview-4692613962): (1) move `DEVOURER_DPDT_MODE` off `std::getenv` into `DeviceConfig` + `env_config.cpp` (the library reads no environment) and drop/fold the undocumented `DEVOURER_SKIP_DPDT`; (2) gate the eFEM pinmux on `rfe_type` 21–24 like the kernel. The `Init`-vs-`InitWrite` asymmetry and the nits can be follow-ups.

src/jaguar3/RtlJaguar3Device.cpp
gilankpam: Fixed in 5d901cb — the applied/default path logs `info` now; the legacy/bit24/skip branches stay `warn`.

src/jaguar3/RtlJaguar3Device.cpp
gilankpam: Removed in 5d901cb — git carries the date.

src/jaguar3/RtlJaguar3Device.cpp
gilankpam: Fixed in 5d901cb. `DEVOURER_DPDT_MODE` now routes through `DeviceConfig` (`tuning.dpdt_8822e`, new `Dpdt8822eMode` enum) with the env mapping in `examples/common/env_config.cpp` — no `getenv` in the library, same shape as `apply_replay_wseq`/`debug.replay_wseq`. The undocumented `DEVOURER_SKIP_DPDT` second spelling is dropped (folded into `mode=skip`).

src/jaguar3/RtlJaguar3Device.cpp
gilankpam: Fixed in 5d901cb. Gated on `rfe_type` 21..24 via a new `HalJaguar3::rfe_type()` (the resolved efuse value, i.e. the same set `config_rfe` keys on); non-eFEM boards keep the reset-default pins and log a skip warning. Verified on hardware — the gate passes on an rfe-21 8812EU (no gate-fail warning fired), so `efem_pinmux_8822e()` executed on the real chip.

src/jaguar3/RtlJaguar3Device.h
gilankpam: Addressed in 5d901cb, including the bench check you suggested.

Extracted the DPDT routing into `apply_dpdt_route_8822e()` and call it from **both** `Init` (RX) and `InitWrite` (TX), each right after `coex_wlan_only_init` — matching the vendor, which runs `_efem_pinmux_config` from `rtl8822e_init_misc` in both directions. `InitWrite` is byte-identical (same code, same call site), so the TX path is unchanged.

Ran the RX-only check on a second 8812EU (aarch64 GS, ch149, `rxdemo` pure-`Init` path, `DEVOURER_RX_ALLPATHS`, **every trial VBUS cold-cycled** so no retained pin state):

```
DPDT mode   chainA_med   chainB_med    n
legacy      41           255 (DEAD)    15894
skip        42           46            14354
efem        46           47            15078
```

So chain B *is* dead under the legacy static route (the 255 is the pwdb floor sentinel = antenna-less) — this independently reproduces the PR-author's finding on a board that could be verified. The eFEM route keeps both chains live. One honest nuance: on *this* board the reset-default (`skip`) already connects both chains, so a pre-change RX-only session was not broken here by default — but aligning `Init` to `efem` guarantees it uses the same good route as the fixed `InitWrite` regardless of board wiring, which is the point of the parity.

josephnef (approved):
Approving — both blocking items from the review are fixed in 5d901cb, and I re-validated on hardware at the new head (205dac0).

**Code:** `DEVOURER_DPDT_MODE` now routes through `DeviceConfig` (`tuning.dpdt_8822e`) with the env mapping in `env_config.cpp` and the undocumented `DEVOURER_SKIP_DPDT` folded into `mode=skip`; the eFEM pin-mux is gated on `rfe_type` 21–24 via the new `HalJaguar3::rfe_type()` (ordering safe — the efuse value resolves before either call site). The deferred `Init`/`InitWrite` asymmetry got a proper fix too: `apply_dpdt_route_8822e()` runs in both bring-ups post-`coex_wlan_only_init`, matching `rtl8822e_init_misc`. CI green including the `8822e-only` config.

**Bench (second independent board — slot BL-M8812EU2, rfe 21, ch36, 8812AU peer):**

- TX regression, `InitWrite` path (`tests/regress.py --tx-pid a81a --rx-pid 8812`, devourer:devourer): 1100/8000 hits / 15 s — TX alive, no regression.
- The new RX-only `Init` path, per-chain RSSI (`DEVOURER_RX_ALLPATHS=1`), VBUS cold-cycle per trial, ~5.5k `rx.path` frames/mode:

  | mode | chain A med | chain B med |
  |---|---|---|
  | efem | 63 | 67 |
  | legacy | 9 | 22 |
  | efem (repeat) | 64 | 68 |

  Independently reproduces the fix on the `Init` path. Interesting datapoint: on this board the legacy route degrades **both** chains, not just B (your board showed A alive / B dead) — same direction, board-dependent damage. The efem repeat matches the first run exactly, the gate passed (`[I] eFEM pin-mux applied`), and the legacy branch warns with the `0x4c` pre-value as designed.

Thanks for the thorough response — the extraction into `apply_dpdt_route_8822e()` plus your own cold-cycled A/B table made this easy to verify.

PR #290, opened 2026-07-14
Discover composite Wi-Fi interfaces and standardize RX FCS

RomanLut:
RTL8822BU support and standardize RX FCS

- discover the Wi-Fi bulk interface/endpoints on composite BT/Wi-Fi adapters
- retain the trailing FCS in Packet::Data across Realtek RX paths
- enable APPFCS for Jaguar2 monitor reception

image

Tested in PixelPilot and hx-esp32-cam android gs.

### DIFF
--- a/src/RxPacket.h
+++ b/src/RxPacket.h
@@ -79,6 +79,8 @@ struct rx_pkt_attrib
 struct Packet
 {
     rx_pkt_attrib RxAtrib;
+    /* Full 802.11 frame including the trailing FCS. Every Realtek RX parser
+     * follows this contract */
     std::span<uint8_t> Data;
 
     /* The transmitter's hardware TX-egress TSF, when the frame carries one.
--- a/src/RxPacket.h
+++ b/src/RxPacket.h
@@ -80,7 +80,11 @@ struct Packet
 {
     rx_pkt_attrib RxAtrib;
     /* Full 802.11 frame including the trailing FCS. Every Realtek RX parser
-     * follows this contract */
+     * follows this contract; consumers remove the FCS at their protocol
+     * boundary rather than making the frame length chip-specific. Retaining it
+     * also lets DEVOURER_RX_KEEP_CORRUPTED and fused-FEC salvage inspect a
+     * failed frame, while tools/bf_report_decode.py trims the trailing four
+     * bytes when decoding beamforming reports. */
     std::span<uint8_t> Data;
 
     /* The transmitter's hardware TX-egress TSF, when the frame carries one.
--- a/src/UsbTransport.cpp
+++ b/src/UsbTransport.cpp
@@ -46,6 +46,11 @@ UsbTransport::UsbTransport(libusb_device_handle *dev_handle, Logger_t logger,
     _info.pid = desc.idProduct;
     _logger->info("USB device {:04x}:{:04x}", _info.vid, _info.pid);
   }
+  // Endpoint addresses are not stable across Realtek USB variants. In
+  // particular, RTL8822BU/"RTL8812BU" is composite: interfaces 0 and 1 are
+  // Bluetooth, while Wi-Fi uses interface 2 with bulk IN endpoint 0x84 rather
+  // than the 0x81 used by RTL8812AU. Discovering the active bulk interface is
+  // therefore required before firmware download, RX, or TX can work.
   discover_endpoints();
   _info.valid = true;
 }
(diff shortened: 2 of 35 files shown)


src/UsbTransport.cpp
```diff
@@ -189,16 +194,42 @@ void UsbTransport::discover_endpoints() {
...
-    const libusb_interface *interface = &config->interface[0];
 
-    if (!interface->altsetting) {
+    if (interface_desc == nullptr) {
+      libusb_free_config_descriptor(config);
       continue;
     }
-    const libusb_interface_descriptor *interface_desc =
-        &interface->altsetting[0];
+    _logger->info("selected USB interface {}", interface_desc->bInterfaceNumber);
```
josephnef, line 217: **Must fix:** the discovered interface number is only logged here — nothing ever *claims* it. Every in-repo consumer hardcodes `claim_interface_then_reset(handle, 0, ...)` (all demos), and per the factory contract that claim happens **before** `UsbTransport` exists. On a composite 8822BU (BT on interfaces 0/1, Wi-Fi on 2) that means: kernel-driver detach hits btusb on interface 0 while rtw88 stays bound to the Wi-Fi interface, and Linux usbfs rejects bulk URBs on an unclaimed interface (EACCES) — so fw download / RX / TX fail in the demos and `tests/regress.py`, exactly on the adapters this PR targets. It works in PixelPilot because the app's own USB stack claims the right interface.

Suggested shape: hoist this walk into a standalone pre-open helper in `UsbOpen.h` (e.g. `find_wifi_interface(libusb_device_handle*)` — same vendor-spec + bulk-IN/OUT scan), and have the demos pass its result to `claim_interface_then_reset` instead of the literal `0`. `discover_endpoints` can then reuse the helper.

src/UsbTransport.cpp
```diff
@@ -189,16 +194,42 @@ void UsbTransport::discover_endpoints() {
...
-    if (!config->bNumInterfaces) {
-      continue;
+    // Do not assume config->interface[0]. Composite adapters expose Bluetooth
+    // first; select the vendor Wi-Fi interface by its bulk IN and OUT pipes.
+    const libusb_interface_descriptor *interface_desc = nullptr;
+    for (uint8_t interface_index = 0;
+         interface_index < config->bNumInterfaces && interface_desc == nullptr;
+         interface_index++) {
+      const libusb_interface *interface = &config->interface[interface_index];
+      for (int alt_index = 0; alt_index < interface->num_altsetting;
```
josephnef: Nit: this inner loop keeps overwriting `interface_desc`, so the **last** matching altsetting of the interface wins. devourer never calls `libusb_set_interface_alt_setting`, so altsetting 0 is what's actually active on the wire — a later altsetting with a different endpoint layout would make us capture endpoints that aren't live. Break on first match (or only consider `altsetting[0]`).

src/jaguar2/HalJaguar2.cpp
```diff
@@ -2298,7 +2298,7 @@ void HalJaguar2::enable_rx() {
    * (rtw88's own default RCR never sets them). Injected canonical-SA beacons
    * passed even WITH the bits set, which is how regress.py stayed green while
    * every real AP's beacons were silently dropped. */
-  uint32_t rcr = 0x7000002Fu;
+  uint32_t rcr = 0xF000002Fu;
```
josephnef, line 2304: The comment block above (line 2288) still derives the old value — `RCR_AAP|APM|AM|AB|APWRMGT|APP_PHYST_RXFF|APP_MIC|APP_ICV = 0x7000002F` — which no longer matches the constant. In this codebase the register comments are the source of truth, so please add `APP_FCS` (BIT31) to the list and note the rationale: parity with Jaguar1 (`RCR_APPFCS`) and Jaguar3 (bit31 in `0xF410400F`) so `Packet::Data` carries the trailing FCS on every generation.

Verified on-air on an RTL8812BU: ambient beacons grow exactly +4 B vs master (282/352 → 286/356).

src/RxPacket.h
```diff
@@ -79,6 +79,8 @@ struct rx_pkt_attrib
 struct Packet
 {
     rx_pkt_attrib RxAtrib;
+    /* Full 802.11 frame including the trailing FCS. Every Realtek RX parser
```
josephnef, line 82: Contract confirmed across generations once this PR lands: Jaguar1 sets `RCR_APPFCS` (`HalModule.cpp`, `RadioManagementModule.cpp:139`), Jaguar3's monitor RCR `0xF410400F` has bit31, and this PR brings Jaguar2 in line. Worth extending the comment with *why* callers care: the FCS is what `DEVOURER_RX_KEEP_CORRUPTED` / the fused-FEC salvage layer and `tools/bf_report_decode.py` (which trims the trailing 4 B) rely on.

josephnef (changes requested):
Thanks — the direction is right on both halves, and the core FCS change **verifies on-air**: A/B on an Archer T3U (RTL8812BU), same ambient beacon populations decode at 282/352 B on master vs 286/356 B on this branch — exactly the +4 appended FCS. The parity claim also checks out at the register level (Jaguar1 sets `RCR_APPFCS` in `HalModule.cpp` / `RadioManagementModule.cpp`; Jaguar3's RCR `0xF410400F` already has bit31 — Jaguar2 was the outlier). Downstream is mechanically clean: `frame_len` passes through untouched, C2H reports carry no FCS, the aggregation walk uses `next_offset`. All 20 ctest selftests pass, and the rewrite incidentally fixes a config-descriptor leak on the old `bNumInterfaces==0` path. Nice.

Requesting changes for one functional gap: the discovered Wi-Fi interface number is only logged, never claimed — on a true composite adapter the in-repo demos claim interface 0 (Bluetooth) and bulk I/O on the Wi-Fi interface fails. Details inline. Two smaller items also inline (stale RCR bit-list comment, last-match altsetting selection).

One more thing to note in the PR description: `enable_rx()` is shared with the RTL8821CE vfio-pci transport, so PCIe RX frames also grow by 4 — `tests/pcie_rx_smoke.py` should get a re-run before merge. And consider retitling (the 8822BU is already supported; the real content is composite-interface discovery + FCS standardization, which is a deliberate +4 B breaking change for existing Jaguar2 consumers).

josephnef (approved):
Force-push `5a3c753d` addresses all three findings — thanks for the quick turnaround.

**Main fix (interface claim):** the new `find_wifi_interface(handle)` helper in `UsbOpen` is exactly the right shape — it runs the vendor-spec + bulk-IN/OUT scan on the raw handle *before* the claim, returns `bInterfaceNumber` (correct for both `libusb_claim_interface` and the `discover_endpoints` match), and all 11 demos + the tests now pass its result instead of the literal `0`. Falls back to `0` for single-interface parts (8812AU/8814AU), so no regression there. `discover_endpoints` reuses the helper and now matches altsetting 0 only — which also resolves the last-match altsetting nit.

**Comments:** both the RCR bit-list (now `... = 0x7000002F` + `APP_FCS (BIT31) = 0xF000002F` with the J1/J3 parity rationale) and the `Packet::Data` contract are updated and accurate.

**Re-verified on hardware** (RTL8812BU / Archer T3U): builds clean, all 20 ctest selftests pass, and rxdemo brings up green through the new path — `selected USB interface 0` / `bulk IN endpoint: 0x84`, RX loop entered, ambient beacon decoded at len=285 (body + 4-byte FCS). The RCR constant is byte-identical to the first commit, so the earlier A/B (+4 B, 282/352 → 286/356) still stands.

Two non-blocking follow-ups for whenever convenient: (1) the two commits share the title `added rtl8822bu support` — a squash-merge with a title reflecting the real content (composite-interface discovery + Jaguar2 RX-FCS standardization, a deliberate +4 B change for existing consumers) would read better in history; (2) `enable_rx()` is shared with the 8821CE vfio-pci transport, so a `tests/pcie_rx_smoke.py` pass on that rig before/after merge is still worth doing. LGTM.

josephnef:
Closed out my open follow-up — **ran `tests/pcie_rx_smoke.py` on the RTL8821CE (PCIe)**. Since `enable_rx()` is shared between the USB and vfio-pci transports and the 8821C has finicky phy-status/drvinfo alignment, I did an A/B of just the RCR line on the vfio-bound 8821CE (radxa-x4, `0000:01:00.0`):

| RCR | ch6 | ch36 |
|---|---|---|
| baseline `0x7000002F` | frames=1087 beacons=845 crc_fail=0 **PASS** | frames=6460 beacons=1160 crc_fail=0 **PASS** |
| PR `0xF000002F` | frames=1051 beacons=792 crc_fail=0 **PASS** | frames=9898 beacons=1276 crc_fail=0 **PASS** |

`beacon_crc_fail=0` on both bands in both configs → the `APP_FCS` (BIT31) bit doesn't disturb 8821C PCIe RX. No regression. (The smoke test keys on CRC cleanliness rather than frame length, but the +4 B FCS itself is already confirmed on USB Jaguar2 and the RCR write is transport-agnostic.) NIC restored to `rtw_8821ce` afterward. That was the last item on my list — no action needed from you.

PR #291, opened 2026-07-15
radiotap TX: honour HT LDPC/STBC known bits on Jaguar2/3

gilankpam:
## Summary

`build_tx_block` on Jaguar2/3 read only bw/sgi/mcs from the HT MCS radiotap field and dropped the FEC/STBC known bits, so an HT frame requesting **LDPC** or **STBC** aired as plain BCC / no-STBC — only VHT frames were honoured. This adds a shared decode helper and wires it into both the Jaguar2 and Jaguar3 HT paths so HT injects apply LDPC/STBC too.

The descriptor fill and STBC-cap guard were already correct: `DATA_LDPC` dword5[7], `DATA_STBC` dword5[9:8] & 0x3 — byte-identical to the vendor driver (`rtl88x2eu-20230815` `hal/rtl8822e/usb/rtl8822eu_xmit.c`, `SET_TX_DESC_DATA_STBC_8822E(.., pattrib->stbc & 3)`).

## Changes

- **`src/RadiotapTxFlags.h`** — `decode_radiotap_mcs_field()`: one shared reading of the HT MCS known/flag/index bytes (bw/sgi/mcs + LDPC/STBC).
- **Jaguar2 + Jaguar3** `build_tx_block` `IEEE80211_RADIOTAP_MCS` cases now use it; the STBC-cap guard below still clamps to what the chip can do.

## Testing

- New host selftest `tests/radiotap_txflags_selftest.cpp` covers the decode. `ctest` 21/21 green.
- Bench-verified on 8822E hardware via `rxdemo`: HT MCS0 injects decoded `stbc=0 ldpc=0` before this change and `stbc=1 ldpc=1` after; live video stayed clean at MCS5 with STBC+LDPC active.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -459,6 +459,11 @@ target_link_libraries(ToneMaskSelftest PRIVATE devourer)
 
 add_test(NAME tone_mask_math COMMAND ToneMaskSelftest)
 
+add_executable(RadiotapTxFlagsSelftest tests/radiotap_txflags_selftest.cpp)
+target_compile_features(RadiotapTxFlagsSelftest PRIVATE cxx_std_20)
+target_link_libraries(RadiotapTxFlagsSelftest PRIVATE devourer)
+add_test(NAME radiotap_txflags COMMAND RadiotapTxFlagsSelftest)
+
 # Headless guard for the burst-TDMA example's pure logic (examples/tdma/tdma.h):
 # the TsfClock host<->TSF least-squares fit, the frame-tag round-trip, and the
 # schedule phase math. No hardware — the tsf_probe / tsf_tdoa_probe /
--- a/src/RadiotapTxFlags.h
+++ b/src/RadiotapTxFlags.h
@@ -1,0 +1,45 @@
+#pragma once
+
+/* Decoded view of the 3-byte radiotap MCS field (known, flags, index) on the
+ * TX/inject path — one shared reading for the per-generation send_packet
+ * parsers, so the HT LDPC/STBC known/flag bits can't silently diverge again
+ * (J2/J3 historically ignored them; only the VHT case honoured them; J1
+ * carried its own correct copy). Bit positions per the radiotap MCS spec and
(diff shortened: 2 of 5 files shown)


josephnef (approved):
Approving — the fix is correct and I verified it end-to-end.

**Correctness:** the shared decode is bit-exact against the radiotap spec and behaviorally identical to Jaguar1's proven copy; `ldpc`/`stbc` init to 0, the `SetTxMode` default still applies only to rate-less frames after the walk, both STBC-cap guards sit downstream, and MCS+VHT dual-field frames stay correct (VHT iterates after MCS and only ever sets). The old J2/J3 behavior of not gating `sgi`/`bw` on their known bits is preserved, so no semantic regressions.

**Impact check:** the encoding matrix missed this because `txdemo`'s rate-less beacon rides the `SetTxMode` default path (already correct). The real victims are `streamtx`/`svctx`/`tdma`/`timesync`/`duplex`, which embed the mode into per-frame radiotap via `build_stream_radiotap` — every HT `/LDPC` stream link on Jaguar2/3 silently aired BCC.

**On-air A/B** (streamtx `MCS1/LDPC/STBC` → 8812AU `rx.txhit` decode, master vs PR build):

| arm | frames | decoded |
|---|---|---|
| 8812EU (J3), master | 12 | `ldpc=0 stbc=0` all |
| 8812EU (J3), PR | 10 | `ldpc=1 stbc=1` all |
| 8812BU (J2), master | 99 | `ldpc=0 stbc=0` all |
| 8812BU (J2), PR | 82 | `ldpc=1 stbc=1` all |

Unanimous truth table on both generations; rate stays MCS1 throughout. `ctest` 21/21.

Nits (will handle as follow-ups on master, no action needed here): `RadiotapTxFlags.h` is included twice in `RtlJaguar3Device.cpp`; the decoder could use the named `IEEE80211_RADIOTAP_MCS_*` constants; and Jaguar1 can adopt the shared decoder so the copies truly can't diverge.

PR #323, opened 2026-07-20
Jaguar3 8822E: caller-supplied per-rate TXAGC diffs (SetTxPowerRateDiffs)

gilankpam:
Adds a third Runtime-TX-power knob on the 8822E: `SetTxPowerRateDiffs`, a
caller-supplied per-rate TXAGC diff table that replaces the chip's default
`phy_reg_pg` by-rate walk. Motivating consumer is a wall-equalized rate
ladder — each rate parked a uniform margin below its measured PA-compression
wall.

## API (`src/IRtlDevice.h`)
```cpp
virtual bool SetTxPowerRateDiffs(
    const std::optional<devourer::TxRateDiffsQdb>& diffs);
```
- `TxRateDiffsQdb` = signed qdB diffs for `cck`, `legacy`, `mcs0..7`
  (`src/TxPower.h`). On the 8822E one qdB == one TXAGC index step.
- `std::nullopt` restores the default by-rate walk.
- Follows the Runtime-TX-power family contract: applies live, sticks across
  `SetMonitorChannel` / `FastRetune` and an override set/clear round-trip,
  folded on top of whichever reference (efuse table or flat override) is
  active.
- Default base returns `false` — 8822E-only in v1; every other generation
  is unaffected.
- `GetTxPowerState` now reports post-diff per-rate indices and a
  `rate_diffs_custom` flag so a caller can read back chip truth.

## Reference consumer & validation
- `examples/txpower/`: `--rate-diffs cck,legacy,m0..m7|clear` and a
  `--flat-pulse N` helper (override + dump + clear in one process).
- `tests/rate_diffs_selftest.cpp` (ctest `rate_diffs`) — diff-word packing.
- `tests/txpwr_rate_diffs_regcheck.sh` — register-level 0x3a00 validation.

## Scope
Jaguar3 8822E only; no behavioral change to any other chip/generation
(the default virtual returns false). 12 files, +513/-16.

## Testing
- `cmake --build` clean; `ctest` 37/37 pass (incl. new `rate_diffs`).
- Rebased cleanly onto current `master` (only doc/`TxPower.h` context
  overlaps with the recently-merged per-packet-TXpower work, resolved to
  keep both features).


src/TxPower.h
```diff
@@ -97,6 +108,22 @@ inline int txpkt_pwr_db_for_step(uint8_t step) {
   return step < 6 ? db[step] : 0;
 }
 
+/* Caller-supplied per-rate TXAGC diffs (signed qdB vs the reference anchor).
+ * v1 consumer: mabur's wall-equalized ladder — each rate parked a uniform
+ * margin below its measured PA-compression wall. Programmed by
+ * IRtlDevice::SetTxPowerRateDiffs; 8822E-only in v1. On the 8822E one qdB
+ * equals one TXAGC index step (step_qdb = 1), so values are used verbatim
+ * as index diffs. */
+struct TxRateDiffsQdb {
```
josephnef, line 118: The hardware field is 7-bit two's complement, so the usable range is **[−64, 63]** — but these are `int8_t` (±127) and `pack_rate_diff_word` masks with `& 0x7f`, which silently **aliases the sign**: a caller asking for −100 gets **+28**. For the adaptive-controller callers this knob targets, that's a power *boost* where a big cut was requested.

The codebase already guards the analogous ref wrap ("an unclamped over-range ref would wrap to near-zero TX silently", `RadioManagementJaguar3.h`). Same treatment here please: clamp to [−64, 63] in `SetTxPowerRateDiffs` (or reject), and document the range on these fields.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -1359,6 +1364,23 @@ void RtlJaguar3Device::SetTxPowerIndexOverride(int idx) {
...
+  if (_variant != jaguar3::ChipVariant::C8822E) {
+    _logger->warn("SetTxPowerRateDiffs: 8822E-only in v1");
+    return false;
+  }
+  if (_cw_active) {
+    _logger->warn("SetTxPowerRateDiffs refused: CW tone active");
+    return false;
+  }
+  std::lock_guard<std::mutex> lk(_reg_mu);
+  _rate_diffs = diffs;
```
josephnef: Companion to the `TxPower.h` range comment — this is the natural place for the clamp (or validation-reject), before the table is stored and walked. Everything downstream (`apply_rate_diffs_8822e`, the `GetTxPowerState` ref+diff estimate) then only ever sees in-range values.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -1568,10 +1590,31 @@ devourer::TxPowerState RtlJaguar3Device::GetTxPowerState() {
...
+    } else {
+      /* Either no custom table configured, or a flat override is currently
+       * live: apply_tx_power_current's flat branch zeroes the chip's
+       * per-rate diffs while flat >= 0, so the registers already read the
+       * honest flat truth here — fall through to the plain summary. */
+      s.ofdm_index = static_cast<int16_t>(ofdm);
+      s.mcs7_index = static_cast<int16_t>(ofdm);
+      s.cck_index = static_cast<int16_t>(cck);
+    }
+    s.rate_diffs_custom = _rate_diffs.has_value();
```
josephnef: This sits inside `if (_brought_up && !_cw_active)`, but the flag means "a table is *configured*" — knowable regardless of bring-up state. As written, a snapshot taken pre-bring-up or during a CW tone reads `rate_diffs_custom=false` with a table set. Move it outside the readback block, next to the other cached (non-readback) fields.

examples/txpower/main.cpp
```diff
@@ -257,6 +276,44 @@ int main(int argc, char **argv) {
...
+      int n = 0;
+      const char* p = a.rate_diffs.c_str();
+      char* end = nullptr;
+      while (n < 10) {
+        v[n++] = static_cast<int>(std::strtol(p, &end, 10));
+        if (*end != ',') break;
+        p = end + 1;
+      }
+      if (n != 10 || *end != '\0') {
+        std::fprintf(stderr, "devourer [E] --rate-diffs wants 10 comma ints "
```
josephnef: This validation runs in `main` *after* the device is opened and `InitWrite` has completed — a typo in the diff string costs a full chip bring-up before exiting 2. Every other flag validates in `parse_args`; parse the 10 ints there too and fail before touching the adapter.

(Also: the `strtol → int8_t` cast is a second silent truncation layer — `200` → `−56`. Moot once the library clamps, but a range check here gives the CLI user an actual error message.)

src/TxPower.h
```diff
@@ -97,6 +108,22 @@ inline int txpkt_pwr_db_for_step(uint8_t step) {
   return step < 6 ? db[step] : 0;
 }
 
+/* Caller-supplied per-rate TXAGC diffs (signed qdB vs the reference anchor).
+ * v1 consumer: mabur's wall-equalized ladder — each rate parked a uniform
```
josephnef: The *why* here is genuinely useful (each rate parked a uniform margin below its measured PA-compression wall — keep that). The person's name is provenance clutter in a library header; the PR/git history carries attribution.

tests/txpwr_rate_diffs_regcheck.sh
```diff
@@ -0,0 +1,226 @@
...
+    pc_rd="$(state_field "$ov_log" "$post_clear_idx" rate_diffs)"
+    pc_ofdm="$(state_field "$ov_log" "$post_clear_idx" ofdm)"
+    pc_mcs7="$(state_field "$ov_log" "$post_clear_idx" mcs7)"
+
+    if [ -z "$pd_ofdm" ] || [ -z "$mp_ofdm" ] || [ -z "$pc_ofdm" ]; then
+        fail "$name override(flat-pulse): missing state field(s) in last 3 events"
+    else
+        pd_want_mcs7=$((pd_ofdm - 16))
+        pc_want_mcs7=$((pc_ofdm - 16))
+        ok=1
```
josephnef: `ok` is assigned (`ok=1`, then `ok=0` in the failure branches) but never read — dead code, drop it.

tests/txpwr_rate_diffs_regcheck.sh
```diff
@@ -0,0 +1,226 @@
...
+# The run emits 4 txpwr.state events total (baseline, post-diffs, mid-pulse,
+# post-clear); we read the LAST THREE:
+#
+#   post-diffs (3rd-from-last): rate_diffs=1, mcs7-ofdm=-16 — the table just
+#     landed (same assertion as check (b), confirms the pulse starts clean).
+#   mid-pulse (2nd-from-last): rate_diffs=1 (a table is still CONFIGURED) BUT
+#     mcs7-ofdm=0 (mcs7 == ofdm == the flat index, 40) — the honest flat
+#     truth: apply_tx_power_current's flat branch zeroed the chip's per-rate
+#     diffs, and GetTxPowerState must report that chip truth rather than
+#     ref+diff during the override (this is the Issue 1 fix under test: a
```
josephnef: "Issue 1" is review-thread context that won't mean anything in-tree. The surrounding sentence already states the invariant (GetTxPowerState must report chip truth, not ref+diff, during an override) — just end it there.

josephnef (changes requested):
## Review: `SetTxPowerRateDiffs` (8822E per-rate TXAGC diffs)

Well-built PR — this extends the Runtime-TX-power family the way it was designed to be extended: `apply_rate_diffs_8822e` mirrors the proven pg-walk byte-for-byte (same 0x1c90[15] gate pattern, ref clamps, `MRateToHwRate` → `0x3a00 + (hw & 0xfc)` addressing, `set_tx_power_ref`'s zero range), routing through `apply_tx_power_current(full=true)` reuses the existing `_diffs_zeroed` machinery instead of inventing new state, and the one-process flat-pulse regcheck is better test design than most contributions.

**Hardware-validated on the bench (2026-07-21):**
- `tests/txpwr_rate_diffs_regcheck.sh` on an 8822EU (0bda:a81a): **6/6 PASS** — apply (mcs7−ofdm = −16 exact), sticky across ch36→ch149 re-fold, and the flat-pulse cell: mid-pulse honest flat truth (mcs7=ofdm=40, `rate_diffs=1` still flagged), spread restored post-clear.
- Regression guard `tests/txpwr_offset_regcheck.sh` over 4 plugged DUTs (8812AU/J1, 8812BU/J2, 8812CU/J3, 8822EU/J3): **36/36 PASS**, canary register dumps **byte-identical to master on every DUT** — with no diff table configured this PR changes nothing at register level.
- CI 18/18 green incl. `8822e-only` / `8822c-only` config legs.

Requesting changes for one real footgun (the ±64 diff aliasing, inline) plus two small correctness items; nits below.

**Nits not worth inline threads:**
- "8822E-only **in v1**" phrasing (`IRtlDevice.h`, `TxPower.h`, the warn string, regcheck header) — this repo keeps docs/comments current-state only; plain "8822E-only" reads the same today and doesn't promise a future. Git is the changelog.
- `--flat-pulse N ` help line has a stray extra space breaking the column alignment.

Not covered by bench validation (and fine to leave open): whether the chip clamps or wraps when ref+diff exceeds 127 — needs an SDR power sweep. The inline clamp makes the question moot for sane callers either way.

src/TxPower.h
gilankpam: Clamped to [-64, 63] in `SetTxPowerRateDiffs` before the table is stored, and documented the range on the `TxRateDiffsQdb` fields + struct comment. With in-range input the `& 0x7f` mask now encodes 7-bit two's-complement without aliasing the sign. 3d1687b

src/jaguar3/RtlJaguar3Device.cpp
gilankpam: Done — the clamp lands here in `SetTxPowerRateDiffs`, before `_rate_diffs` is stored and walked, so `apply_rate_diffs_8822e` and the `GetTxPowerState` ref+diff estimate only ever see in-range values. 3d1687b

src/jaguar3/RtlJaguar3Device.cpp
gilankpam: Moved `rate_diffs_custom = _rate_diffs.has_value()` out of the readback block — it's reported regardless of bring-up / CW state now. Hoisted `_reg_mu` over both the flag read and the readback so the non-atomic `_rate_diffs` read stays race-free (removed the now-redundant inner lock). 3d1687b

examples/txpower/main.cpp
gilankpam: Parsing + range-check [-64, 63] moved into `parse_args`; a malformed or out-of-range `--rate-diffs` now fails with an explicit error and exit 2 before the adapter is opened. The range check also replaces the silent `strtol`->`int8_t` truncation with a real CLI error. 3d1687b

src/TxPower.h
gilankpam: Dropped the name; kept the wall-equalized-ladder / PA-compression-margin rationale. 3d1687b

tests/txpwr_rate_diffs_regcheck.sh
gilankpam: Removed. 3d1687b

tests/txpwr_rate_diffs_regcheck.sh
gilankpam: Trimmed to the invariant sentence; dropped the 'Issue 1' thread reference. 3d1687b

gilankpam:
Pushed 3d1687b addressing the review. The two nits from the summary (no inline thread):

- **"in v1" phrasing** — removed across the feature: `IRtlDevice.h`, `TxPower.h`, the warn string, the regcheck header, and the two `RadioManagementJaguar3` comments. `grep -rn 'in v1'` over src/tests is now empty.
- **`--flat-pulse` help space** — removed the stray space; the description column realigns with the rest of the block.

Local `ctest` 37/37; the register-level regcheck assertions are unchanged (the clamp only affects out-of-range input, which the bench cells don't exercise).

josephnef (approved):
Follow-up commit 3d1687b3 addresses all seven review threads cleanly, and I re-validated the fixed head on hardware.

**Resolutions**
- **[blocking] ±64 clamp** — `SetTxPowerRateDiffs` now clamps to the 7-bit two's-complement field range `[-64, 63]` before storing, at the top of the setter, so both `apply_rate_diffs_8822e` and the `GetTxPowerState` ref+diff estimate only ever see in-range values. The `& 0x7f` sign-aliasing footgun is closed.
- `rate_diffs_custom` reported regardless of bring-up/CW state, under a hoisted `_reg_mu` (redundant inner lock removed).
- `--rate-diffs` parse + `[-64, 63]` range-check moved into `parse_args`, failing before the adapter opens rather than silently truncating.
- Dropped the provenance name, the dead `ok=` var, the review-thread reference, and the "in v1" future-promising phrasing.

**Validation on `3d1687b3`**
- Build clean; **ctest 37/37** green.
- **Hardware regcheck 6/6 PASS** on the 8822EU (`0bda:a81a`): default walk, apply `mcs7−ofdm=−16`, sticky across `ch36→ch149`, flat-pulse honest-truth + override-clear restore.
- CLI range-check confirmed: out-of-range and wrong-count inputs are rejected before bring-up.

LGTM — thanks for the clean turnaround.

PR #335, opened 2026-07-24
Add a WPA2/CCMP station-mode client (APFPV)

GingerFluffyCat:
## Summary

devourer currently has no way to *be* a WPA2 client — it injects/receives in
monitor mode, and the AP-mode test harnesses under `tests/` play the AP role
against a real station. This adds the missing piece: a full station-mode
client that scans, authenticates, associates, runs the WPA2 4-way handshake,
gets an IP over DHCP, and streams data — usable as a real Wi-Fi uplink for an
RTL8812AU-family dongle, not just a monitor/injection tool.

New library surface (mirrors the existing `devourer`/chip-family target
pattern): two new CMake targets, `apfpv` (driver-independent: WPA2
supplicant, 802.11 frame builders, CCMP crypto, beacon/scan parsing) and
`apfpv_station` (the driver-dependent orchestrator + RX pipeline).

## What's included

- **`src/apfpv/`** — `Wpa2Supplicant` (4-way handshake, client side),
  `Dot11Frames` (auth/assoc/probe request builders), `Wpa2Crypto` +
  `crypto/{AesCcm,AesCmac,Sha1Hmac}` (own CCMP/PBKDF2/PRF implementation, no
  external crypto dependency), `ScanProbe` (beacon parsing for AP
  discovery), `StationMode` (the arming sequence: MACID → MSR=STATION →
  BSSID → RCR filter, ported from the kernel's `hw_var_set_opmode` STATION
  path), `StationTxDesc` (the unicast TX descriptor — BMC=0 enables the
  ACK handshake, the fix that made auth/assoc frames actually get ACKed).
- **`ApfpvStation`** — the top-level orchestrator: `connect()` runs the
  gated scan→arm→auth→assoc→4-way→DHCP→stream chain; a driver-level
  supervisor thread auto-reconnects on link loss (deauth or an RX/beacon
  timeout), re-arming on the known channel rather than a full rescan.
  Also exposes a general-IP bridge (arbitrary IPv4 over the CCMP link, not
  just an RTP/video assumption) and a minimal in-order TCP stack over it.
- **`RxDeframe`** — the station RX pipeline: CCMP decrypt, A-MPDU RX
  reorder (kernel `recv_indicatepkt_reorder` equivalent — required once the
  AP aggregates; without it, out-of-order subframes are dropped), BlockAck
  request handling, and RTP-sequence reordering for the video use case this
  was built for.
- **`ApfpvDhcp`** / **`LqFeedback`** — a DHCP client and a link-quality
  telemetry feed.
- **Driver hooks needed for the above** (`RtlAdapter`): `fillH2CCmd`/
  `sendH2CPacket` (firmware mailbox), `setSecCamKey`/`enableHwSec` (security
  CAM programming for the PTK/GTK), `sendStationFrameSync` (concurrent
  IN/OUT unicast TX with an ACK-drain check, validated against the
  worst-case USB/IP transport), `reset_device`.
- **Two hardware-validated fixes** (`RadioManagementModule`/
  `PhydmWatchdog`): `SetStationRxFilter` — the kernel-matched RCR/RXFLTMAP
  configuration station mode needs (the prior permissive filter dropped
  management frames the auth/assoc handshake needs); and a close-range
  DIG-saturation fix — once linked, track the receiver's gain floor/ceiling
  off the live beacon RSSI instead of the always-monitor-mode bounds, or a
  strong/close signal storms the false-alarm counter and the front-end goes
  deaf (no RX at all) even though the link should be trivially strong.

## Caveats / what's *not* claimed

- Single client only (no multi-station bookkeeping beyond what's needed for
  the one active connection).
- `ApfpvStation.cpp` also contains AP-mode (SoftAP) scaffolding —
  beaconing, RSSI tracking, and a WPA2 4-way *authenticator* (the AP side) —
  wired onto this repo's own `StartBeacon`/`StopBeacon`. It's explicitly
  marked work-in-progress in its own comments: association completion and
  the data plane (ARP/ICMP/DHCP responding to an associated client) aren't
  there yet, unlike this repo's own `tests/ap_wpa2.cpp`/`ap_responder.cpp`
  harnesses, which do have a working data plane. Happy to split it out of
  this PR if you'd rather review the station-mode work on its own — say the
  word and I'll drop those commits/functions.
- Bench-validated on RTL8812AU (Jaguar1) against a real OpenIPC AP; not
  tested against Jaguar2/3/Kestrel chips.

## Testing

Built cleanly as a standalone CMake target (`apfpv_station`) against this
branch; also integrated and building end-to-end in a downstream Android app
(PixelPilot) alongside the existing WFB-ng monitor-mode path.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -221,6 +221,7 @@ add_library(apfpv
     src/apfpv/crypto/AesCcm.cpp      src/apfpv/crypto/AesCcm.h
     src/apfpv/crypto/AesCmac.cpp     src/apfpv/crypto/AesCmac.h
     src/apfpv/crypto/Sha1Hmac.cpp    src/apfpv/crypto/Sha1Hmac.h
+    src/apfpv/ScanProbe.cpp          src/apfpv/ScanProbe.h
     src/ApfpvDhcp.cpp                src/ApfpvDhcp.h
     src/LqFeedback.cpp               src/LqFeedback.h
     src/Wpa2Authenticator.cpp        src/Wpa2Authenticator.h)
@@ -228,6 +229,20 @@ target_compile_features(apfpv PUBLIC cxx_std_20)
 target_include_directories(apfpv PUBLIC src src/apfpv src/apfpv/crypto)
 if(WIN32)
     target_link_libraries(apfpv PUBLIC ws2_32 bcrypt)
+endif()
+
+# ApfpvStation/RxDeframe DO need the driver layer (RtlAdapter/RtlJaguarDevice/
+# RadioManagementModule) -- separate target so `apfpv` itself stays
+# driver-independent (it's also used as-is on hosts with no libusb at all).
+add_library(apfpv_station
+    src/ApfpvStation.cpp     src/ApfpvStation.h
+    src/RxDeframe.cpp        src/RxDeframe.h
+    src/apfpv/StationMode.cpp src/apfpv/StationMode.h
+    src/apfpv/StationTxDesc.cpp src/apfpv/StationTxDesc.h)
+target_link_libraries(apfpv_station PUBLIC apfpv devourer)
+target_compile_features(apfpv_station PUBLIC cxx_std_20)
+if(NOT ANDROID)
+    target_include_directories(apfpv_station PUBLIC src/apfpv/compat)
 endif()
 # Host (non-Android) builds: Wpa2Supplicant.cpp etc. call __android_log_print
 # unconditionally (ported as-is from WiFiDriver) -- same compat shim already
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,35 @@ target_include_directories(devourer PUBLIC hal/phydm/rtl8822b)
 target_include_directories(devourer PUBLIC hal/phydm/rtl8821c)
 target_include_directories(devourer PUBLIC src)
 
+# ---------------------------------------------------------------------------
+#  APFPV additions -- station-mode WPA2/CCMP + the APFPV IP/FPV/session layer,
+#  ported onto upstream's rebuilt tree. Only WPA2Supplicant/Dot11Frames/
+#  Wpa2Crypto/crypto/*/ApfpvDhcp/LqFeedback/Wpa2Authenticator so far (zero
+#  driver-layer dependency, confirmed by direct #include inspection); RxDeframe
+#  and ApfpvStation come later once their driver-layer API updates land.
+# ---------------------------------------------------------------------------
+add_library(apfpv
+    src/apfpv/Wpa2Supplicant.cpp     src/apfpv/Wpa2Supplicant.h
+    src/apfpv/Dot11Frames.cpp        src/apfpv/Dot11Frames.h
+    src/apfpv/Wpa2Crypto.cpp         src/apfpv/Wpa2Crypto.h
+    src/apfpv/crypto/AesCcm.cpp      src/apfpv/crypto/AesCcm.h
+    src/apfpv/crypto/AesCmac.cpp     src/apfpv/crypto/AesCmac.h
+    src/apfpv/crypto/Sha1Hmac.cpp    src/apfpv/crypto/Sha1Hmac.h
+    src/ApfpvDhcp.cpp                src/ApfpvDhcp.h
+    src/LqFeedback.cpp               src/LqFeedback.h
+    src/Wpa2Authenticator.cpp        src/Wpa2Authenticator.h)
+target_compile_features(apfpv PUBLIC cxx_std_20)
+target_include_directories(apfpv PUBLIC src src/apfpv src/apfpv/crypto)
+if(WIN32)
+    target_link_libraries(apfpv PUBLIC ws2_32 bcrypt)
+endif()
+# Host (non-Android) builds: Wpa2Supplicant.cpp etc. call __android_log_print
+# unconditionally (ported as-is from WiFiDriver) -- same compat shim already
+# proven there, reused verbatim rather than reinvented.
+if(NOT ANDROID)
+    target_include_directories(apfpv PUBLIC src/apfpv/compat)
+endif()
+
 # --- per-chip source groups (see options above) ---
 if(DEVOURER_JAGUAR1)
     target_sources(devourer PRIVATE
--- a/src/ApfpvStation.cpp
+++ b/src/ApfpvStation.cpp
@@ -1,0 +1,1768 @@
+// ============================================================================
+//  ApfpvStation.cpp — orchestrator + DRIVER-LEVEL persistent reconnect
+//  Mirrors the VRX's wpa_supplicant -B daemon behavior at the driver level:
+//  a supervisor thread re-establishes the link on loss (deauth or RX timeout),
+//  with no app/UI involvement. Faster than the VRX on recovery because we
+//  re-arm on the known channel (no full rescan).
+// ============================================================================
+#include "ApfpvStation.h"
+#include "StationMode.h"
+#include "Wpa2Supplicant.h"
+#include "Wpa2Authenticator.h"
+#include "hal_com_reg.h"     // MSR / REG_BSSID / REG_RCR / RCR_* / FORCEACK for AP-mode HW config
+#include "Wpa2Crypto.h"
+#include "ApfpvDhcp.h"
+#include "RxDeframe.h"
+#include "StationTxDesc.h"
+#include "Dot11Frames.h"
+#include "FrameParser.h"
+#include "ScanProbe.h"
+#include "LqFeedback.h"
+#if defined(__ANDROID__)
+#include <sys/system_properties.h>
+#endif
+#include "PhydmWatchdog.h"
+#include "RtlUsbAdapter.h"
+#include "RtlJaguarDevice.h"
+#include "SelectedChannel.h"
+#include "RadioManagementModule.h"
+#include <algorithm>
+#include <cstring>
+#if defined(__ANDROID__)
+#include <sys/system_properties.h>
+#endif
+
+namespace {
+// Android has no settable process env, so DEVOURER_* getenv knobs never fire in the app. Mirror the
+// key streaming knobs to `debug.pixelpilot.*` props (adb setprop, no root). Returns true if the prop
+// exists and is not "0". Host builds (no __ANDROID__) always return false → env-only there.
+static bool apfpvProp(const char* name) {
+#if defined(__ANDROID__)
+    char v[PROP_VALUE_MAX] = {0};
+    if (__system_property_get(name, v) > 0 && v[0] != '0') return true;
+#else
+    (void)name;
+#endif
+    return false;
+}
+// ACCEPT the AP's BlockAck (arm RX-BA regs + ADDBA-init + StatusCode 0) instead of declining.
+// Env DEVOURER_ENABLE_BA (host) or `debug.pixelpilot.ba=1` (adb setprop, Android). THE fundamental-fix
+// test: now that FW_RA makes the AP actively aggregate A-MPDU to us, does accepting BA make the chip's
+// MAC auto-emit the SIFS compressed BlockAck so the AP retransmits lost subframes → losses recovered
+// (no more reference-corruption freeze)? Prior sessions tested this only when the AP WASN'T aggregating.
+static bool enableBaOn() {
+    return std::getenv("DEVOURER_ENABLE_BA") != nullptr || apfpvProp("debug.pixelpilot.ba");
+}
+// Buffer size (in MPDUs) to negotiate in OUR ADDBA Response, overriding the AP's request (see
+// handleAddbaRequest). 0/unset = echo the AP's requested value unchanged. DEVOURER_BA_BUFSZ (host)
+// / debug.pixelpilot.babufsz (Android).
+static int baRespBufSizeOverride() {
+    if (const char* e = std::getenv("DEVOURER_BA_BUFSZ")) { int n = std::atoi(e); if (n > 0) return n; }
+#if defined(__ANDROID__)
+    char v[PROP_VALUE_MAX] = {0};
+    if (__system_property_get("debug.pixelpilot.babufsz", v) > 0) {
+        int n = std::atoi(v);
+        if (n > 0) return n;
+    }
+#endif
+    return 0;
+}
+} // namespace
+// Platform-agnostic: <android/log.h> is the NDK header on Android and the
+// compat stderr shim on native host builds (see WiFiDriver/compat), so the same
+// diagnostics surface on Windows/Linux too — essential for chasing parity.
+#include <android/log.h>
+#define SCANLOG(...) __android_log_print(ANDROID_LOG_INFO, "apfpv-scan", __VA_ARGS__)
+
+namespace apfpv {
+
+// Map a primary channel + desired bandwidth (MHz) to a SelectedChannel kept
+// INSIDE the legal DE APFPV band 5170-5250 MHz (5.2 GHz UNII-1; 200 mW @ 20 MHz
+// under the 10 mW/MHz PSD cap). For 40 MHz we pick the HT40 secondary side that
+// never crosses 5170/5250 — primary 36/44 extend ABOVE (offset LOWER = primary
+// is the lower 20 MHz), primary 40/48 extend BELOW (offset UPPER); both land on
+// centres 38 (5170-5210) or 46 (5210-5250). 20 MHz is pass-through; 80 MHz
+// centres on 42 and fills the band exactly. NB: WFB's 5.8 GHz / 25 mW band is a
+// separate code path and is unaffected by this helper.
+static SelectedChannel legalApfpvChannel(int ch, int bwMHz) {
+    uint8_t c = (uint8_t)(ch > 0 ? ch : 40);
+    if (bwMHz >= 80)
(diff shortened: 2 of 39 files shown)


src/ApfpvStation.cpp
```diff
@@ -0,0 +1,1780 @@
...
+void ApfpvStation::startBeaconCal(const std::string& ssid, int channel, int txIndex) {
+    stopBeaconCal();
+    disconnect();                       // can't be a station and beacon at once
+    if (!_rtl || !_dev) return;
+    auto* rtl = reinterpret_cast<RtlJaguarDevice*>(_rtl);
+    auto* dev = reinterpret_cast<RtlUsbAdapter*>(_dev);
+    try {
+        rtl->StopRxLoop(); safeJoinThread(_rxThread);   // no RX needed for a TX-only beacon
+        rtl->InitWrite(legalApfpvChannel(channel, 20));   // tune + bring up (no RX loop)
+        rtl->SetTxPowerIndexOverride(std::max(0, std::min(63, txIndex)));
```
josephnef, line 139: **Blocker — this is the red CI (MSVC).** `build (windows-latest, Release, cl)` fails here:
```
ApfpvStation.cpp(139,43): error C2589: '(': illegal token on right side of '::'
ApfpvStation.cpp(1675,32): error C2589: ...
```
`bcrypt`/`ws2_32` pull in `<windows.h>`, which defines function-like `min(`/`max(` macros, so `std::min(63, txIndex)` (and line 1675) won't compile. Fix: `#define NOMINMAX` before any Windows header (or wrap as `(std::min)(...)`). `<algorithm>` is already included — the macro is the problem.

src/apfpv/Wpa2Supplicant.cpp
```diff
@@ -0,0 +1,418 @@
...
+                bool sent = _send(buildMsg2());
+                _state = State::WaitMsg3;
+                fprintf(stderr, "[wpa] M1 -> built+sent M2 (sent=%d) -> WaitMsg3\n", (int)sent);
+                WPALOG("[wpa] M1 rx -> M2 sent=%d -> WaitMsg3", (int)sent);
+            } else {
+                WPALOG("[wpa] WaitMsg1 but frame not M1 (mic=%d pairwise=%d) — ignored", (int)hasMic, (int)isPairwise);
+            }
+            return false;
+        case State::WaitMsg3:
+            if (hasMic && isPairwise) {
```
josephnef, line 258: **Blocker (KRACK-class) — the M3 MIC is never verified.** This case installs the GTK and goes `Done` without ever computing `_c.eapol_mic(_kck, ...)` over the frame (MIC field zeroed) and comparing it to the received MIC. Same omission on the group-rekey and PTK-rekey paths below. MIC verification is the step that proves the AP holds the PMK — without it the STA cannot reject a forged/rogue M3. Verify the MIC before deriving/installing keys or sending M4; reject on mismatch. (GTK unwrap's RFC-3394 check and TK-decrypt failure are *not* substitutes.)

src/apfpv/Wpa2Supplicant.cpp
```diff
@@ -0,0 +1,418 @@
...
+}
+
+bool Wpa2Supplicant::onEapolKey(const uint8_t* body, size_t len) {
+    WPALOG("[wpa] onEapolKey len=%zu state=%d", len, (int)_state);
+    if (len < 95) { WPALOG("[wpa] EAPOL too short (len=%zu <95) — dropped", len); return false; }
+    // EAPOL anti-replay: the AP's 8-byte replay counter (offset 9) must not go backwards once
+    // the handshake is up — rejects replayed EAPOL-Key frames (KRACK-class). Only enforced past
+    // Done so a reconnect's fresh (possibly lower) M1 counter is never rejected; begin() resets.
+    uint64_t rc = 0; for (int i = 0; i < 8; ++i) rc = (rc << 8) | body[9 + i];
+    if (_state == State::Done && rc < _replay) {
```
josephnef, line 230: **Blocker (KRACK-class) — off-by-one in the replay gate.** `rc < _replay` accepts an *equal* counter. A replayed post-`Done` group-rekey carries `rc == _replay`, passes here, and `installGtkFromKeyData()` reinstalls the GTK and resets its RX PN window to 0 (line ~170) — group-key reinstallation. Combined with the missing MIC check above, an attacker can replay a captured rekey frame to reset the broadcast replay window. Post-handshake rekeys must require a **strictly greater** counter *and* a valid MIC.

src/apfpv/Wpa2Supplicant.cpp
```diff
@@ -0,0 +1,418 @@
...
+        if (t == 0 || l == 0) break;
+        if (t == 0xdd && (size_t)(i + 2 + l) <= plen &&
+            kp[i+2]==0x00 && kp[i+3]==0x0f && kp[i+4]==0xac) {
+            if (kp[i+5]==0x01 && l >= 22) {        // GTK KDE: <keyid(low2)|tx><rsv> <GTK16..>
+                uint8_t newId = kp[i+6] & 0x03;
+                if (_gtkKeyId != 0xff && _gtkKeyId != newId) { _gtkPrev = _gtk; _gtkPrevId = _gtkKeyId; }
+                _gtkKeyId = newId;
+                std::memcpy(_gtk.data(), &kp[i+8], 16);
+                _rxPnGtk[newId & 3] = 0; winClear(_rxPnGtkWin[newId & 3]);  // fresh key: reset RX replay window
+                fprintf(stderr, "[wpa] GTK installed keyid=%d (%02x%02x..)\n", _gtkKeyId, _gtk[0], _gtk[1]);
```
josephnef, line 171: Key material must never hit logs — this prints the first two GTK bytes to stderr. Drop it (and route any diagnostics through the repo `Logger`, not `fprintf`).

src/apfpv/crypto/AesCcm.cpp
```diff
@@ -0,0 +1,351 @@
...
+// Data RX only: hardware-AES CCMP decrypt (ARM-CE or x86 AES-NI). Returns false
+// to fall back to the SW path (e.g. an x86 CPU lacking AES-NI). Only aes_enc_ce
+// differs per-arch; the CCM (CTR + CBC-MAC) logic below is shared.
+bool aes_ccm_decrypt_ce(const uint8_t key[16],const uint8_t* nonce,size_t nlen,
+                        const uint8_t* aad,size_t aadlen,
+                        const uint8_t* ct,size_t ctlen,uint8_t* out){
+    if(ctlen<8) return false;
+    // A/B knob: DEVOURER_FORCE_SW_AES=1 forces the software-AES fallback (the pre-fix path)
+    // so the AES-NI win can be measured back-to-back on the same link.
+    static const bool kForceSw = std::getenv("DEVOURER_FORCE_SW_AES") != nullptr;
```
josephnef, line 231: The library must not read the environment (`DEVOURER_FORCE_SW_AES`). This is a bench A/B knob — either delete it now the AES-NI path is validated, or lift it to `DeviceConfig`. Same rule applies to the `getenv` sites in `RxDeframe.cpp`, `StationMode.cpp`, `RadioManagementModule.cpp`, and the 40+ in `ApfpvStation.cpp`.

src/apfpv/crypto/AesCcm.cpp
```diff
@@ -0,0 +1,351 @@
...
+                     const uint8_t* aad,size_t aadlen,const uint8_t* ct,size_t ctlen,uint8_t* out){
+    if(ctlen<8) return false; size_t ptlen=ctlen-8;
+    Aes a; aes_key(a,key);
+    // decrypt: CTR first to recover plaintext, then MAC-verify
+    uint8_t A[16]={0}; A[0]=1; memcpy(A+1,nonce,nlen);
+    uint16_t ctr=1; size_t i=0;
+    while(i<ptlen){ A[14]=ctr>>8;A[15]=ctr&0xff; uint8_t Sx[16]; aes_enc(a,A,Sx);
+        size_t k=ptlen-i<16?ptlen-i:16; for(size_t j=0;j<k;j++) out[i+j]=ct[i+j]^Sx[j]; i+=k; ctr++; }
+    uint8_t mic[8],calc[16]; (void)calc;
+    uint8_t tmp[2048]; if(ptlen>sizeof(tmp)) return false;
```
josephnef, line 107: The fixed `tmp[2048]` scratch silently fails decrypt for any plaintext > 2048 bytes (large/A-MSDU MPDUs). The scratch is also pointless — it holds a re-encryption nobody reads; compute the CBC-MAC without emitting output. (Also: the MIC compare at line 109/125 is `memcmp`, not constant-time — minor, but easy to fix here.)

src/RxDeframe.cpp
```diff
@@ -0,0 +1,672 @@
...
+        lastPkt = nowp;
+        if (gapMs > 12)
+            __android_log_print(ANDROID_LOG_WARN, "rx-gap", "#%d RX gap %ldms (>12ms = RX blind this long)", ++gapN, gapMs);
+    }
+    if (_station) _station->notifyRxAlive();   // any RX = link alive (supervisor)
+    if (_lq) _lq->update(toDbm(pkt.RxAtrib.rssi[0]), toDbm(pkt.RxAtrib.rssi[1]));
+    // Feed live RX RSSI to DIG so it uses phydm *connected-mode* boundaries
+    // (IGI floor tracks RSSI ~0x37 @ -45 dBm) instead of monitor coverage
+    // bounds (capped 0x2a → over-gained → FA storm → AP rate-caps us).
+    PhydmWatchdog::SetLinkRssi(toDbm(pkt.RxAtrib.rssi[0]));
```
josephnef, line 90: Layering: this is generation-agnostic `src/` core reaching into `jaguar1/PhydmWatchdog` (include at line 7). Together with the 8812AU-hardcoded registers in `ApfpvStation`/`StationMode`, this station code is effectively jaguar1-only despite living in the neutral tree — scope it jaguar1-side and gate the build on `DEVOURER_JAGUAR1`.

src/jaguar1/RadioManagementModule.cpp
```diff
@@ -2979,3 +3011,86 @@ void RadioManagementModule::PHY_TxPowerTrainingByPath_8812(RfPath rfPath) {
...
+          // retransmit storm that collapsed paced throughput.
+          RCR_VHT_DACK;
+    // RCR_APP_MIC (BIT30) + RCR_APP_ICV (BIT29) -- the MAC retains the CCMP
+    // MIC/ICV at the bottom of each RX frame. Only needed when HW decrypt is
+    // on (RxDeframe's HW-decrypt path strips CCMP-hdr(8)+MIC(8)=16 bytes,
+    // which is only correct if the HW kept the MIC -- without this the -16
+    // eats 8 bytes of real payload off every frame -> corrupt NALUs -> black
+    // screen). Default off (the proven SW-CCMP path). DEVOURER_RCR_APP or
+    // DEVOURER_HW_DECRYPT opts in.
+    bool rcrApp = std::getenv("DEVOURER_RCR_APP") != nullptr
```
josephnef, line 3059: Two problems here: (1) library reads env (`DEVOURER_RCR_APP`/`DEVOURER_HW_DECRYPT`/…) — move to `DeviceConfig`; (2) **NDK build blocker** — the `__ANDROID__` branch just below calls `__system_property_get`/`PROP_VALUE_MAX` (and `__android_log_print` at line 616) but this file includes neither `<sys/system_properties.h>` nor `<android/log.h>`. Host builds preprocess the block away, so it compiles here but breaks on the actual Android target.

src/apfpv/Dot11Frames.cpp
```diff
@@ -0,0 +1,252 @@
...
+                                       uint32_t pairwiseCipher, uint32_t groupCipher,
+                                       uint16_t rsnCaps) {
+    // Reuse the base frame, then append an RSN IE with the negotiated ciphers.
+    std::vector<uint8_t> f = BuildAssocRequest(self, bssid, ssid, rsnCaps);
+    // strip the trailing hardcoded RSN (last sizeof(rsn)+2 bytes) and re-add.
+    // Simpler: the base already appended a CCMP/PSK RSN; if negotiated == CCMP
+    // it is identical, so only rebuild when different.
+    if (pairwiseCipher == 0x000FAC04 && groupCipher == 0x000FAC04) return f;
+    // remove last RSN IE (0x30 ... ) — find last 0x30 tag
+    for (size_t i = f.size(); i-- > 0; ) {
```
josephnef, line 239: The negotiated-RSN overload strips the RSN IE by scanning backwards for byte `0x30` and `f.resize(i)` — but the base builder appends RSN *before* the WMM/HT/VHT/ExtCap tail, so this truncates the entire tail (the AP then sees an assoc-req with no HT/VHT caps — the exact 'never sends M1' failure your own comment describes). It also mis-triggers on an `rsnCaps` low byte of `0x30`. Track the RSN offset when building and splice in place, or walk IEs properly.

CMakeLists.txt
```diff
@@ -207,6 +207,50 @@ target_include_directories(devourer PUBLIC hal/phydm/rtl8822b)
...
+target_compile_features(apfpv PUBLIC cxx_std_20)
+target_include_directories(apfpv PUBLIC src src/apfpv src/apfpv/crypto)
+if(WIN32)
+    target_link_libraries(apfpv PUBLIC ws2_32 bcrypt)
+endif()
+
+# ApfpvStation/RxDeframe DO need the driver layer (RtlAdapter/RtlJaguarDevice/
+# RadioManagementModule) -- separate target so `apfpv` itself stays
+# driver-independent (it's also used as-is on hosts with no libusb at all).
+add_library(apfpv_station
```
josephnef, line 237: `apfpv_station` links `RadioManagementModule`/`PhydmWatchdog`/`RtlJaguarDevice` (jaguar1) but is built unconditionally. It's a static archive so CI's `JAGUAR1=OFF` cells don't fail today, but any executable linking it with jaguar1 off gets undefined refs. Gate this target on `DEVOURER_JAGUAR1` like `DEVOURER_8814` does.

josephnef (changes requested):
## Request changes — WPA2/CCMP station mode (APFPV)

Substantial, genuinely useful work, and the crypto core is mostly correct — but this can't merge as-is. **CI is red** (see below), there are two KRACK-class handshake defects, and the PR broadly violates this repo's standing conventions (the library reads no environment; two-plane logging; generation-agnostic core). Inline comments mark the specific spots; the must-fix summary is below.

### CI is red 🔴
`build (windows-latest, Release, cl)` **fails** — the rest of the matrix is green. Root cause is a real MSVC portability break in `ApfpvStation.cpp` (`min`/`max` macro collision), flagged inline. It must build on MSVC — Windows is first-class here.

### Blockers
1. **EAPOL-Key MIC is never verified** — M3, group-rekey, and PTK-rekey all act on unauthenticated frames (`Wpa2Supplicant.cpp`). This is *the* authentication step of the 4-way; skipping it is a KRACK-class defect.
2. **EAPOL replay gate uses `<` not `<=`** — an equal-counter replayed group-rekey passes and resets the GTK PN window (KRACK group-key reinstallation).
3. **MSVC build broken** (CI red) — `std::min`/`std::max` vs the `<windows.h>` `min`/`max` macros.
4. **Android NDK build broken** — `RadioManagementModule.cpp` uses `__android_log_print` / `__system_property_get` / `PROP_VALUE_MAX` under `#if __ANDROID__` without including `<android/log.h>` / `<sys/system_properties.h>` (host builds preprocess the block away, hiding it — ironically the PR's own target platform).

### Major
5. **The library reads the environment pervasively** — 40+ `getenv`/Android-prop sites in `ApfpvStation.cpp`, plus `RxDeframe.cpp`, `AesCcm.cpp`, `StationMode.cpp`, `RadioManagementModule.cpp`. CLAUDE.md: *"The library reads no environment."* These belong in `DeviceConfig` (env mapped in `examples/common/env_config.cpp`); the one-shot `*_OLD` / `FORCE_SW_AES` A/B knobs should just be deleted now the fixes are validated.
6. **Direct logging bypasses the two-plane contract** — `fprintf(stderr,...)` and `__android_log_print(...)` throughout library sources, including one that prints GTK bytes to stderr. Route through the repo `Logger`.
7. **Layering** — generation-agnostic core (`RxDeframe`, `ApfpvStation`) reaches into `jaguar1/` and hardcodes 8812AU registers; nothing in the API says "8812 only." `apfpv_station` isn't gated on `DEVOURER_JAGUAR1` (builds as a static archive today so CI doesn't catch it, but any consumer linking it with JAGUAR1=OFF gets undefined refs).
8. **Negotiated-cipher assoc-request truncates the HT/VHT/ExtCap tail** (`Dot11Frames.cpp`) — latent (only the non-default cipher path), but a real frame-corruption bug.
9. **No tests** — hand-rolled AES/CCM/PBKDF2/PRF/RFC-3394 with *zero* KATs, in a repo with a strong ctest culture. Highest-leverage thing to add; the code comments themselves record several already-fixed crypto bugs.
10. **WIP AP-mode/authenticator scaffolding** ships in the same PR (admittedly untested). Please take up your own offer to split it out — it's a second unverified 4-way implementation and enlarges the review/security surface.

### Not regressed (good)
The jaguar1 **monitor-mode** path is correctly fenced off: the DIG fix and `SetStationRxFilter` gate on station-only state, so the primary use case is safe.

Minor items (fixed 2048-byte CCM scratch caps decrypt; non-constant-time MIC `memcmp`; no key zeroization; `#if 0` dead code; dated archaeology comments; `chooseCipher` can pick unsupported suites) can ride the same pass.

Happy to re-review once the blockers + CI are addressed.

GingerFluffyCat:
Closing this — the approach here isn't right for upstream; withdrawing it. Thanks.

PR #346, opened 2026-07-27
Share the length-prefixed record reader across the stdin demos

josephnef:
Follow-up to #345, which noted this as a known duplication rather than folding it in.

`stream_stdin.h` centralised `set_stdin_binary()` and `read_exact()` so there would be one source of truth instead of one copy per demo — but it stopped one level too low. The thing every caller actually wanted, *read one `<u32_le len><body>` record*, stayed replicated in four demos, and the copies had already drifted: the little-endian assembly re-typed each time, three different range checks, and four different answers to what an EOF part-way through a record means.

## What's shared now

`read_length()`, `read_body()`, and the composed `read_record()`.

The result states are deliberately finer than "it failed", because that distinction is precisely what the callers differ on:

| state | meaning |
|---|---|
| `Ok` | a complete record |
| `Eof` | clean close — nothing where a record would have started |
| `Short` | the stream ended part-way through a record |
| `EofMidBody` | the length word arrived, then the stream closed before a byte of its body |
| `BadLength` | zero, or larger than the caller's bound |

Each demo maps those onto its own policy and none re-derives them: `streamtx` warns on one and exits 2 on another, `txdemo` counts the lost shard and stops, `svctx` treats any incomplete record as the end of the clip, `duplex` stops its TX thread and lets RX run on.

`duplex` keeps composing `read_length()` + `read_body()` rather than taking `read_record()`: its top length bit escapes to a control TLV with its own much smaller bound, so it must see the length before the body may be read. That's the case the primitives exist for.

## Testing

The selftest now drives `read_record()` — the path the demos actually run — and pins what each state means, through `tmpfile()` for the cases a well-formed stream never produces: truncated prefix, length with no body, body cut short, zero length, length past the bound.

Validated on air, not only headlessly:

- **txdemo** reader through the adaptive FEC harness: 0.869 → 0.935 FEC delivery across an exclusion, in line with the four runs before this change.
- **streamtx** reader through the jammer-resilience matrix: 6132 bodies aired, 0.917 clean / 0.975 jammed.
- **duplex** and **svctx** are covered by the shared selftest and the build only — neither has a bench harness here.

47/47 ctest green.

### DIFF
--- a/examples/common/stream_stdin.h
+++ b/examples/common/stream_stdin.h
@@ -14,10 +14,22 @@
 //      truncates the first PSDU ("short read on stdin (76/269)") before a
 //      single frame is transmitted.
 //
-//   2. read_exact() — the length-prefixed record reader, returning a tri-state
-//      so each caller keeps its own short-read policy (the TX demo aborts on a
-//      truncated record; the duplex demo just stops its TX thread and lets RX
-//      run on).
+//   2. read_exact() — the byte reader, returning a tri-state so each caller
+//      keeps its own short-read policy (the TX demo aborts on a truncated
+//      record; the duplex demo just stops its TX thread and lets RX run on).
+//
+//   3. read_record() — one whole <u32_le len><body> record, which is what
+//      every caller actually wanted. The little-endian assembly and the
+//      zero/oversize check were re-typed in four demos before this existed,
+//      and they had already drifted. Callers that need the length word before
+//      the body — the duplex demo escapes to a control TLV on its top bit —
+//      compose read_length() and read_body() instead.
+//
+// The result states are deliberately finer than "it failed": a producer that
+// closed cleanly between records is an ordinary end of run, one that closed
+// with a length word already written has lost a record, and a length out of
+// range is a producer bug rather than a stream end. Each demo maps them to its
+// own policy; none of them has to re-derive them.
 //
 // StreamStdinSelftest + tests/stream_stdin_test.cmake exercise this header
 // headlessly (no libusb, no hardware), so a regression in the _WIN32 gate
@@ -27,6 +39,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstdio>
+#include <vector>
 
 #if defined(_WIN32)
   #include <io.h>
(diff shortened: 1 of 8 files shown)


josephnef:
End-to-end bench coverage added for the two demos that had none here (`bca86ba`), which closes the gap this PR's description called out.

**duplex** — `tests/duplex_ctl_onair.sh`. `tests/adaptive_onair.sh` does exercise the control escape, but `adaptive_link.py` consumes duplex's event stream on a pipe and drops what it isn't looking for, so nothing could observe whether a TLV was read or acted on. The new script scripts the stream instead and witnesses the air:

```
control TLVs read: 2   {"ev":"stream.ctl","op":2,"len":3}  {"ev":"stream.ctl","op":1,"len":2}
PSDUs aired: 1500 of 1500
witnessed DESC_RATE values: 4(x980) 13(x452)
```

452 frames at MCS1 (segment 1) then 980 at 6M (segments 2–3) — the TLVs were read, SET_RATE reached the air, and every PSDU aired, so the record stream stayed in sync across both escapes.

**svctx** — `tests/svctx_uep_witness.sh`, using rxdemo as the witness rather than an 8814AU in kernel monitor mode (`rx.frame` already carries the decoded rate), so it runs on any two adapters:

```
witnessed frames: 1837
DESC_RATE histogram: 12(x509) 13(x816) 16(x512)
```

Three ladder rungs on air (MCS0 critical / MCS1 T0 / MCS4 T1), which also proves the NAL records were read with their temporal_ids intact.

**Refactor equivalence, measured rather than argued:** I rebuilt duplex from the pre-refactor source against the same library and ran the same script — identical outcome (2 TLVs, 1500/1500 aired). So the reader change is behaviour-preserving on hardware, not just by inspection.

One trap is now documented in the script, because it cost me a wrong conclusion first: the TLV must switch **down** to a more robust rate. At this bench's ~13 dB margin an MCS1 → MCS5 switch is indistinguishable from no switch at all — the frames fly at the new rate, the witness can't decode them, and you get one rate in the histogram plus a quietly lower frame count. I initially reported that as a pre-existing SET_RATE defect; it wasn't. The frame counts gave it away (each run decoded exactly the segments flown at the robust rate).

Issue #348, opened 2026-07-28
TX degrades across warm sessions: high-order constellations die, only a VBUS power cycle recovers (a drone cannot power-cycle in flight)


josephnef:
## Summary

Repeated devourer sessions on one adapter progressively degrade **high-order
constellation TX** while leaving the robust rates untouched. Only a **VBUS
power cycle** recovers it — `libusb_reset`, teardown/re-init, and idling do
not.

This is a real deployment bug, not just a bench annoyance: **an airborne link
cannot power-cycle its adapter.** A drone that restarts its TX process, or runs
a long session, silently loses its top rates with nothing in any log to say so.

## Why it is dangerous

It is not a wedge, and that is what makes it easy to misdiagnose. Throughout
the degradation the adapter:

- keeps enumerating and accepting frames;
- reports **every frame submitted**, zero failures;
- keeps delivering the robust rates at full ratio;
- reports strong RSSI and clean EVM *on the frames that do decode* (because
  only the low-order frames decode, and those are clean).

The signature is indistinguishable from "the link is too weak to carry a denser
constellation". In the investigation that surfaced this, it produced a
completely coherent but wrong conclusion — a peaked delivery-vs-power curve,
the same ceiling on an empty channel (ch14), and the same ceiling across three
different receivers — all of it an artifact of accumulated chip state.

## Reproduction

`tests/warm_tx_degradation_repro.sh` (added alongside this report). It holds one
ground receiver up for the whole run so the reference never moves, then
alternates {probe} and {N warm devourer sessions} on the DUT, and finishes with
two controls.

```sh
sudo REGRESS_VBUS_MAP="0bda:8812=3-2.3.4,3;2357:012d=10,2" \
     tests/warm_tx_degradation_repro.sh
```

Two independent runs, RTL8812AU TX -> RTL8822BU ground, ch6, 20 MHz, constant
TX config:

| stage | MCS7 (64-QAM 5/6) | MCS1 control |
| --- | --- | --- |
| cold baseline | 63.3% / 62.6% | 97.7% |
| after 3 warm sessions | 58.1% / 58.1% | 97.2% |
| after 6 warm sessions | 56.3% / 56.6% | 97.3% |
| after 9 warm sessions | 57.4% / 55.2% | 97.1% |
| after 12 warm sessions | 56.3% / 55.1% | 97.2% |
| after 60 s idle, **no** power cycle | 57.6% / 59.6% | 97.0% |
[...]

PR #350, opened 2026-07-28
Jaguar1: power the chip down at teardown (mechanism open; thermal explanation retracted)

josephnef:
Addresses the reproducible half of #348. **Note the title is now inaccurate** —
see "Mechanism: open" below; the fix stands, the explanation did not.

## What is fixed

A Jaguar1 session left the chip in the **ACT power state with the RF front end
live, indefinitely**, long after the owning process exited. Nothing else in the
tree does this: Jaguar2 resets at init, Jaguar3 has run a full card-disable at
teardown for a long time.

`Stop()` and the destructor now run `HalModule::rtw_hal_deinit()` — halt the MAC
engines (`REG_CR`, `REG_RCR`), then the die's card-disable power sequence,
through the same `HalPwrSeqCmdParsing` and the same three-way chip dispatch
`InitPowerOn` already uses. The tables (`rtl8812_card_disable_flow` and the
8814A/8821A equivalents) had been sitting in `hal/Hal88*PwrSeq.c` since the port
began with nothing calling them.

**Measured effect** — same repro, before and after:

| stage | before | after |
| --- | --- | --- |
| 60 s idle, **no** power cycle | thermal 48, MCS7 73.0% | thermal 40, MCS7 **85.8%** |

~13 points, about 7 sd against the measured noise floor. Before the fix, no
amount of idling recovered anything and only pulling VBUS helped.

The change is also justified independently of any performance argument: leaving
a radio powered and transmitting-capable after its owning process has exited is
wrong, it is what every other generation already avoids, and it is what makes an
autonomously-airing beacon stop at session end.

## Mechanism: open (a correction)

An earlier revision of this PR claimed the cause was chip temperature. **That
was correlation across sessions presented as causation, and it is retracted.**
Every recovery that works removes power, and removing power resets chip state
*and* lets the die cool — no power-cycle experiment separates them. Two
harnesses added here try to, and both argue against cooling:

- **`thermal_offtime_sweep.sh`** — degrade, then power down for 5/15/30/60/120 s
  before an otherwise identical probe. Reset bit-identical in every arm; only
  cooling time varies. Temperature fell cleanly with off-time (47 → 40) but
  delivery scattered 63–83% with **no relation to off-time or temperature** —
  best result from the *shortest* off-time and the *hottest* chip.
- **`thermal_causation_probe.sh`** — delivery inside a single uninterrupted
  session (one bring-up, no re-init, no state change). Thermal stayed **pinned**
  through 240 s of max-duty TX while MCS7 drifted ~10% and the control held flat.

Heat remains a real correlate within one probe sequence (thermal 39 → 45 costing
~5 points, control flat, RSSI unchanged) and is demonstrably not the driver
above. The docs now say the mechanism is open rather than asserting it.

## Know the noise floor

**`probe_repeatability.sh`** is the measurement that should have come first: ten
identical back-to-back probes, nothing changed between them.

| | mean | sd | range |
| --- | --- | --- | --- |
| MCS7/20 | 67.3% | **1.8** | 65.0 – 70.7 |
| MCS1/20 control | 98.0% | 0.2 | 97.5 – 98.3 |

A single probe is worth ±3 points, so **effects under ~4 points are not
findings** — and several plausible-looking decay curves in this investigation sat
inside that band. This is now documented in `CLAUDE.md` as a bench rule.

## Behaviour change

A beacon in the MAC's reserved page now stops when the session ends instead of
airing until re-enumeration. Matches Jaguar3 and the vendor driver; `CLAUDE.md`
updated.

## Testing

- 8812AU, 8814AU, 8821AU each re-init cleanly across three consecutive sessions
  after the power-down.
- `teardown_gen_sanity.sh` 10/10 across every plugged generation.
- `tx_teardown_asan.sh` 6/6 including max-duty kills — the case that matters now
  that teardown writes registers while URBs may be in flight.
- `ctest` 48/48.

## Not included, deliberately

- **No in-flight recovery API.** Until the mechanism is understood there is
  nothing principled to implement.
- **No init-side power-off.** It runs immediately before power-on, so it buys
  nothing while adding risk to a working bring-up path.
- **Jaguar2 unchanged.** Audited and inconclusive — thermal did not move and the
  available link was too marginal at MCS7 to resolve a small effect. No
  measurement, no speculative change.

## Still open in #348

The severe form (MCS5/6/7 at 0% while MCS0–4 held ~88%) is unreproduced, the
mechanism is unresolved, and `SIGKILL` still leaves the chip powered.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -612,17 +612,23 @@ generators, never the output files.
   only helps when firmware state is intact.
 - **The chip retains state across soft re-init** — cold-bisect hardware
   problems with a VBUS power-cycle, not a re-run.
-- **Chip temperature silently caps the usable modulation.** A hot 8812AU loses
-  the dense constellations first while the robust rates carry on: measured on
-  ch6/20 MHz, thermal meter 43 → 53 costs MCS7 (64-QAM) ~83% → ~71% delivery
-  with RSSI *flat*, i.e. no power is lost, only signal quality. Nothing logs an
-  error and the signature is indistinguishable from a link too weak to carry
-  the rate. Any rate-ceiling measurement must therefore control temperature —
-  VBUS-cycle per cell, and read the `thermal` event
-  (`DEVOURER_THERMAL_POLL_MS`) beside the delivery number.
-  `tests/warm_tx_degradation_repro.sh` is the reference method. Removing power
-  is what cools the part: on a chip left in ACT, 120 s of idle bought 4 thermal
-  units where a 12 s VBUS cycle bought 9.
+- **A single delivery probe is worth ±3 points, so small effects are not
+  findings.** Ten identical back-to-back MCS7/20 probes on an 8812AU (nothing
+  changed between them) gave sd 1.8 and a 5.7-point spread; the MCS1 control
+  over the same run held 98.0 ± 0.2. Anything under ~4 points needs repetition
+  before it means anything — several plausible-looking "decay curves" have
+  turned out to sit inside that band. `tests/probe_repeatability.sh` measures
+  the floor for a given pair; run it before believing a delivery difference.
+- **A hot 8812AU tends to lose the dense constellations while the robust rates
+  carry on**, and it is easy to over-read. Delivery does correlate with the
+  `thermal` meter within one probe sequence (39 → 45 costing MCS7 ~5 points,
+  control flat) — but a sweep that varied *only* cooling time, by powering the
+  chip down for 5–120 s before an otherwise identical probe, found delivery
+  scattered 63–83% with no relation to either off-time or temperature. So heat
+  is a real correlate and not a demonstrated cause. Read the `thermal` event
+  (`DEVOURER_THERMAL_POLL_MS`) beside any rate-ceiling number, VBUS-cycle per
+  cell, and repeat the cell — do not attribute a difference to temperature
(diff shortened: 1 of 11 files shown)


josephnef:
Closing while the investigation continues, per review.

The teardown power-down here has a measured effect but does **not** fix #348, and I got its justification wrong twice (latched state, then thermal — both asserted from correlation, both retracted). Shipping it now would mean shipping a change whose reason I cannot state.

The branch stays so the harnesses and measurements are not lost:
- `probe_repeatability.sh` — the noise floor (MCS7/20 sd 1.8, 5.7-point spread; effects under ~4 points are not findings)
- `thermal_offtime_sweep.sh` — identical reset, only cooling time varies; refuted the thermal story
- `thermal_causation_probe.sh` — delivery inside one uninterrupted session
- `warm_tx_degradation_repro.sh` — the mild-form repro

Investigation continues on the actual lead: TX IQ correction (`0xcc4/0xcc8/0xccc/0xcd4`, `0xec4/…`) and RF `0x58` LOK are written per session, are **not** reset by the BB/RF init tables, and are **not** in any dump we have — so they are both invisible and potentially persistent, and they govern exactly the EVM/phase-noise damage that kills dense constellations while leaving robust rates intact.

Everything will come back as a single PR once the state is demonstrated and the fix is validated.

PR #384, opened 2026-08-06
jaguar3: EFUSE walk assumed ordered sections and quit early (8822C read rfe_type=0)

snokvist:
## The bug

`HalJaguar3::read_efuse_logical_map` stopped walking as soon as a section's
logical base passed the byte the caller asked for:

```c
if (base > upto + 8)
  break; /* past the byte we need */
```

That is only valid if sections appear in **ascending base order**. They do not.

Physical EFUSE dumped off an RTL8822CU (`0bda:c812`), decoded by hand:

```
phys 0x00  hdr=0x00         -> base 0x000   ok
phys 0x09  hdr=0x10         -> base 0x008   ok
phys 0x12  hdr=0x0F ext=48  -> base 0x100   <- early exit fires here
phys 0x2C  hdr=0x4F ext=5D  -> base 0x150   never reached
phys 0xD5  hdr=0x4F ext=4E  -> base 0x110   never reached
phys 0xDA  hdr=0x4F ext=5E  -> base 0x150   never reached
```

The third section on the chip jumps to base 0x100, so **any** request below that
— including `EEPROM_RFE_OPTION_8822C` at logical 0xCA, which is the whole reason
`read_efuse_rfe_type()` calls this — ended the walk after three sections and
returned a map that was 0xFF almost everywhere.

## Why it matters

On the affected adapter `read_efuse_rfe_type()` returned **0**, while the vendor
kernel driver reads **0x03** from the same chip
(`/proc/net/rtl88x2cu/<iface>/efuse_map`, logical 0xCA). The RFE type gates BB /
RFE configuration, so those units were being brought up against an unprogrammed
default rather than their actual front-end.

It is silent: nothing errors, the map just reads unprogrammed.

## The fix

Walk the whole programmed area (the existing 0xFF-header terminator and
`kPhysMax` bound already stop it). The `upto` parameter is removed rather than
left unused — a parameter that still *looks* like it bounds the walk is how this
comes back.

The 8822E branch is untouched: it never used `upto`, terminating on a long 0xFF
run instead, which is why only the C path was affected.

## Hardware verification

| adapter | before | after | kernel (`efuse_map` 0xCA) |
|---|---|---|---|
| RTL8822CU (`0bda:c812`, C8822C) | `rfe_type=0x00` | `rfe_type=0x03` | **0x03** |
| RTL8822EU (`0bda:a81a`, C8822E) | `rfe_type=0x15` | `rfe_type=0x15` | **0x15** |

The EU is the regression check — unchanged, and its `efuse decoded (0x22=46
0x4c=51 0xca=15)` line is identical before and after. The 8822C EFUSE stability
probe also now reports a valid `0x8129` EEPROM ID.

Found while implementing #383 (EFUSE MAC as a per-unit identity), which could not
read the MAC on 8822C for this reason. With this fix that adapter's MAC decodes
correctly — `40:a5:ef:2f:23:08`, matching its netdev exactly. The two changes are
independent; this one stands on its own regardless of what happens to #383.

### DIFF
--- a/src/jaguar3/HalJaguar3.cpp
+++ b/src/jaguar3/HalJaguar3.cpp
@@ -601,28 +601,6 @@ void HalJaguar3::config_pa_bias_8822e() {
                 pg2a & 0xf, pg2b & 0xf, pg5a & 0xf, pg5b & 0xf);
 }
 
-/* Decode the packed (extended-header) EFUSE into a logical map, up to (and
- * including the block holding) every programmed logical offset. Shared by
- * read_efuse_rfe_type and read_efuse_txpwr_base_8822e. `map` must be zero-init'd
- * by the caller (this fills 0xFF for gaps). Standard Realtek section format:
- * header (or header+ext) gives a logical block offset + 4-bit word-enable; each
- * enabled 2-byte word follows.
- *
- * The walk runs to the end of the programmed area. It used to stop early once a
- * section's base passed the byte the caller asked for, which silently assumed
- * the sections appear in ascending base order — they do not. Measured on an
- * RTL8822CU, the third section on the chip jumps to base 0x100:
- *
- *   phys 0x00  hdr=0x00         -> base 0x000
- *   phys 0x09  hdr=0x10         -> base 0x008
- *   phys 0x12  hdr=0x0F ext=48  -> base 0x100   <- early exit fired here
- *   phys 0x2C  hdr=0x4F ext=5D  -> base 0x150   (never reached)
- *
- * so asking for anything below 0x100 — including EEPROM_RFE_OPTION at 0xCA —
- * ended the walk after three sections and returned a map that was 0xFF almost
- * everywhere. On that adapter read_efuse_rfe_type() therefore returned 0 while
- * the kernel driver read 0x15 from the same chip, i.e. the BB/RFE config was
- * being chosen from an unprogrammed default. */
 bool HalJaguar3::probe_efuse_map(uint8_t *map, size_t len) {
   /* 8822E OTP reads are not reliable after TX/coex bring-up (by design — see
    * cache_efuse_8822e); probing there would flag healthy units. 8822C only. */
@@ -634,6 +612,26 @@ bool HalJaguar3::probe_efuse_map(uint8_t *map, size_t len) {
   return true;
 }
 
+/* Decode the packed (extended-header) EFUSE into a logical map. Shared by
+ * read_efuse_rfe_type and read_efuse_txpwr_base_8822e. `map` must be zero-init'd
+ * by the caller (this fills 0xFF for gaps). Standard Realtek section format:
+ * header (or header+ext) gives a logical block offset + 4-bit word-enable; each
+ * enabled 2-byte word follows.
+ *
+ * The walk decodes the whole programmed area. Sections are not in ascending base
+ * order, so it must not stop at any requested offset — measured on an RTL8822CU,
+ * the third section on the chip jumps to base 0x100:
+ *
+ *   phys 0x00  hdr=0x00         -> base 0x000
+ *   phys 0x09  hdr=0x10         -> base 0x008
+ *   phys 0x12  hdr=0x0F ext=48  -> base 0x100
+ *   phys 0x2C  hdr=0x4F ext=5D  -> base 0x150
+ *
+ * A walk bounded by the byte the caller asked for ends after those first three
+ * sections for anything below 0x100 — including EEPROM_RFE_OPTION at logical
+ * 0xCA — and returns a map that is 0xFF almost everywhere. On that adapter it
+ * made read_efuse_rfe_type() return 0 while the kernel driver reads 0x03 from
+ * the same chip, i.e. BB/RFE config chosen from an unprogrammed default. */
 void HalJaguar3::read_efuse_logical_map(uint8_t *map, size_t len) {
   constexpr uint16_t kPhysMax = 1024; /* EFUSE_REAL_CONTENT_LEN_8822C */
   for (size_t i = 0; i < len; ++i) map[i] = 0xFF;
--- a/src/jaguar3/HalJaguar3.cpp
+++ b/src/jaguar3/HalJaguar3.cpp
@@ -602,11 +602,27 @@ void HalJaguar3::config_pa_bias_8822e() {
 }
 
 /* Decode the packed (extended-header) EFUSE into a logical map, up to (and
- * including the block holding) logical offset `upto`. Shared by read_efuse_rfe_type
- * and read_efuse_txpwr_base_8822e. `map` must be zero-init'd by the caller (this
- * fills 0xFF for gaps). Standard Realtek section format: header (or header+ext)
- * gives a logical block offset + 4-bit word-enable; each enabled 2-byte word
- * follows. */
+ * including the block holding) every programmed logical offset. Shared by
+ * read_efuse_rfe_type and read_efuse_txpwr_base_8822e. `map` must be zero-init'd
+ * by the caller (this fills 0xFF for gaps). Standard Realtek section format:
+ * header (or header+ext) gives a logical block offset + 4-bit word-enable; each
+ * enabled 2-byte word follows.
+ *
+ * The walk runs to the end of the programmed area. It used to stop early once a
+ * section's base passed the byte the caller asked for, which silently assumed
+ * the sections appear in ascending base order — they do not. Measured on an
+ * RTL8822CU, the third section on the chip jumps to base 0x100:
+ *
+ *   phys 0x00  hdr=0x00         -> base 0x000
+ *   phys 0x09  hdr=0x10         -> base 0x008
+ *   phys 0x12  hdr=0x0F ext=48  -> base 0x100   <- early exit fired here
+ *   phys 0x2C  hdr=0x4F ext=5D  -> base 0x150   (never reached)
+ *
+ * so asking for anything below 0x100 — including EEPROM_RFE_OPTION at 0xCA —
+ * ended the walk after three sections and returned a map that was 0xFF almost
+ * everywhere. On that adapter read_efuse_rfe_type() therefore returned 0 while
+ * the kernel driver read 0x15 from the same chip, i.e. the BB/RFE config was
+ * being chosen from an unprogrammed default. */
 bool HalJaguar3::probe_efuse_map(uint8_t *map, size_t len) {
   /* 8822E OTP reads are not reliable after TX/coex bring-up (by design — see
    * cache_efuse_8822e); probing there would flag healthy units. 8822C only. */
@@ -614,11 +630,11 @@ bool HalJaguar3::probe_efuse_map(uint8_t *map, size_t len) {
     return false;
   if (map == nullptr || len != sizeof(_efuse_cache))
     return false;
-  read_efuse_logical_map(map, len, 0xFA);
+  read_efuse_logical_map(map, len);
   return true;
 }
 
-void HalJaguar3::read_efuse_logical_map(uint8_t *map, size_t len, uint16_t upto) {
+void HalJaguar3::read_efuse_logical_map(uint8_t *map, size_t len) {
   constexpr uint16_t kPhysMax = 1024; /* EFUSE_REAL_CONTENT_LEN_8822C */
   for (size_t i = 0; i < len; ++i) map[i] = 0xFF;
 
@@ -704,15 +720,13 @@ void HalJaguar3::read_efuse_logical_map(uint8_t *map, size_t len, uint16_t upto)
           map[idx] = d;
       }
     }
-    if (base > upto + 8)
-      break; /* past the byte we need */
   }
 }
 
 void HalJaguar3::cache_efuse_8822e() {
   if (_variant != ChipVariant::C8822E)
     return;
-  read_efuse_logical_map(_efuse_cache, sizeof(_efuse_cache), 0xFA);
+  read_efuse_logical_map(_efuse_cache, sizeof(_efuse_cache));
   _efuse_cache_valid = true;
   _logger->info("Jaguar3(8822e): efuse decoded (0x22={:x} 0x4c={:x} 0xca={:x})",
                 _efuse_cache[0x22], _efuse_cache[0x4c], _efuse_cache[0xca]);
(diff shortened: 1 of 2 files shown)


snokvist:
Both review findings addressed.

**1. Read past `kPhysMax` — real, and this PR did make it reachable.** The loop head checks `phys < kPhysMax` once per section, but a header + ext header + four enabled data words advance it up to ten bytes further, and `efuse_OneByteRead` masks the address to 10 bits — so a read at 1024 aliases to 0 and would decode the *start* of the EFUSE into whatever logical base the truncated section named. Guarded in the shared `rd` lambda, which covers both the 8822C and 8822E walks in one place:

```c
if (a >= kPhysMax)
  return 0xFF;
```

`0xFF` is already what both walks treat as end-of-map/skip, so the truncated section terminates the walk exactly as an unprogrammed area does.

**2. Stale `upto` in the header comment** — my error; I appended to the comment instead of rewriting its first sentence. Reworded to describe the actual behaviour.

Re-verified on hardware after both changes, unchanged from the original results:

| adapter | rfe_type | kernel `efuse_map` 0xCA |
|---|---|---|
| RTL8822CU (C8822C) | `0x03` | **0x03** |
| RTL8822EU (C8822E) | `0x15` | **0x15** |

src/jaguar3/HalJaguar3.cpp
```diff
@@ -602,23 +602,39 @@ void HalJaguar3::config_pa_bias_8822e() {
...
+ *
+ *   phys 0x00  hdr=0x00         -> base 0x000
+ *   phys 0x09  hdr=0x10         -> base 0x008
+ *   phys 0x12  hdr=0x0F ext=48  -> base 0x100   <- early exit fired here
+ *   phys 0x2C  hdr=0x4F ext=5D  -> base 0x150   (never reached)
+ *
+ * so asking for anything below 0x100 — including EEPROM_RFE_OPTION at 0xCA —
+ * ended the walk after three sections and returned a map that was 0xFF almost
+ * everywhere. On that adapter read_efuse_rfe_type() therefore returned 0 while
+ * the kernel driver read 0x15 from the same chip, i.e. the BB/RFE config was
```
josephnef: Factual slip: per the PR's own hardware table, the kernel value on the affected **8822CU is 0x03** — **0x15 is the 8822EU's** (the regression-check adapter). The follow-up commit says "correct the stale doc" but this one survived it. Worth fixing: this exact number is what a future debugging session will trust.

```suggestion
 * the kernel driver read 0x03 from the same chip, i.e. the BB/RFE config was
```

src/jaguar3/HalJaguar3.cpp
```diff
@@ -602,23 +602,39 @@ void HalJaguar3::config_pa_bias_8822e() {
 }
 
 /* Decode the packed (extended-header) EFUSE into a logical map, up to (and
- * including the block holding) logical offset `upto`. Shared by read_efuse_rfe_type
- * and read_efuse_txpwr_base_8822e. `map` must be zero-init'd by the caller (this
- * fills 0xFF for gaps). Standard Realtek section format: header (or header+ext)
- * gives a logical block offset + 4-bit word-enable; each enabled 2-byte word
- * follows. */
+ * including the block holding) every programmed logical offset. Shared by
```
josephnef: Splice leftover from the old "up to offset `upto`" text — "up to (and including the block holding) every programmed logical offset" is grammatically broken now. Simpler to just say it decodes the whole programmed area.

Also (pre-existing, but this PR grows the block 4×): this comment documents `read_efuse_logical_map` yet sits above `probe_efuse_map`. Consider moving it down to the function it describes while touching it.

src/jaguar3/HalJaguar3.cpp
```diff
@@ -602,23 +602,39 @@ void HalJaguar3::config_pa_bias_8822e() {
...
+ * read_efuse_rfe_type and read_efuse_txpwr_base_8822e. `map` must be zero-init'd
+ * by the caller (this fills 0xFF for gaps). Standard Realtek section format:
+ * header (or header+ext) gives a logical block offset + 4-bit word-enable; each
+ * enabled 2-byte word follows.
+ *
+ * The walk runs to the end of the programmed area. It used to stop early once a
```
josephnef: Style nit: the "It used to stop early…" framing is changelog-ish — git carries the history. The measured phys dump is the valuable part (it proves the unordered-sections invariant) and should stay; the framing could be present-tense: "sections are not in ascending base order (measured on an RTL8822CU: …), so the walk must not stop at any requested offset."

josephnef (commented):
Reviewed with the branch checked out; decode math verified against the dump, callers audited. The fix is correct and the hardware evidence is exactly the right shape — before/after on the affected 8822CU cross-validated against the vendor kernel `efuse_map`, with the 8822EU as an explicit no-change regression check. CI fully green.

**Correctness — verified:**
- The decode math checks out against the dump: `hdr=0x0F ext=0x48` → offset `0x20` → base `0x100`; `hdr=0x4F ext=0x5D` → `0x2A` → base `0x150`. Both match the hand-decode, so the unordered-sections claim is substantiated, not inferred.
- Removing `upto` instead of leaving it unused is the right call — a parameter that looks like a bound but isn't would reintroduce exactly this bug class.
- Smaller callers stay safe: `read_efuse_rfe_type`'s `0x140`-byte stack map vs the section at base `0x150` is handled by the `idx < len` guard on every data write.
- The `kPhysMax` guard in `rd()` is load-bearing, not belt-and-braces: `efuse_OneByteRead` masks to 10 bits, so a truncated straddling section would alias back to phys 0 and decode the start of the EFUSE into a bogus logical base. Previously the `upto` break made that unreachable; with the full walk it matters.
- 8822E untouched as claimed: the EU branch `return`s before the removed break, and neither of its callers depended on `upto`.

**Performance:** the 8822C non-cached paths now walk the full programmed area (per-byte USB control reads), but the walk still stops at the first 0xFF header and both callers are one-shot at bring-up/probe time — no meaningful cost.

One factual error in the shipped comment (inline, worth fixing before merge — it's the kind of number a future debugging session will trust) plus two comment-style nits inline.

src/jaguar3/HalJaguar3.cpp
snokvist: Fixed in 38790b1 — the number is now 0x03 (the affected 8822CU), matching the PR's hardware table and the before/after. Thanks for catching it; 0x15 was the 8822EU regression-check value bleeding into the wrong sentence.

src/jaguar3/HalJaguar3.cpp
snokvist: Both done in 38790b1 — the block now just says it decodes the whole programmed area, and it moved down onto `read_efuse_logical_map` where it belongs. `probe_efuse_map` keeps its own 8822C-only note.

src/jaguar3/HalJaguar3.cpp
snokvist: Reframed in 38790b1 to present tense: the invariant (sections are not in ascending base order) with the measured phys dump as its evidence. The "it used to" framing is gone — git has it.

snokvist:
Thanks for the review — all three comments are addressed in 38790b1 (comment-only; build clean, 49/49 ctest):

- **the 0x15 slip** → now `0x03`, the affected 8822CU's value. You were right that it is the kind of number a later session trusts; `0x15` was the 8822EU regression-check value in the wrong sentence. The same slip was in the PR body of #383 and is fixed there too.
- **the splice + placement** → the block no longer tries to say "up to … every offset"; it just says it decodes the whole programmed area, and it moved down onto `read_efuse_logical_map`. `probe_efuse_map` keeps its own 8822C-only note.
- **present tense** → the invariant with the measured dump as evidence; the "it used to" framing is gone.

## Independent cross-check against the vendor parser

Since the argument for removing `upto` is "sections are not ordered", it is worth having the vendor's own walk on record. `eeprom_parser_88xx()` (rtl88x2cu 20230728, `hal/halmac/halmac_88xx/halmac_efuse_88xx.c:1139`) is a `do { … } while (1)` with **no logical-offset bound at all** — it terminates only on `hdr == 0xff` (or `hdr2 == 0xff`) and on physical-size guards. So the post-fix walk is the vendor's, and the pre-fix `upto` exit was devourer-local.

Two things fall out of that:

- The **`kPhysMax` guard is the right shape** — the vendor bounds-checks `efuse_idx` after *every* advance (header, ext, and each of the 8 data bytes), and is stricter still: it fails the whole parse rather than continuing. Returning `0xFF` from `rd()` is the gentler equivalent.
- The **header decode matches bit-for-bit**: vendor `blk_idx = ((hdr2 & 0xF0) >> 1) | ((hdr >> 5) & 0x07)` vs devourer `((ext & 0xF0) >> 1) | ((hdr & 0xE0) >> 5)`, so the hand-decode in the comment (`0x0F/0x48 → 0x100`, `0x4F/0x5D → 0x150`) is the vendor's arithmetic, not a re-derivation.

It also answers the "previously-read bytes *moving*" half of #385: the vendor writes `log_map[eeprom_idx] = value8` unconditionally as it walks, so a later section overriding an earlier one is the designed re-burn semantics — and the measured unit does carry two sections at base `0x150` (phys `0x2C` and `0xDA`).

## #385's validation asks

**"check that `read_efuse_txpwr_base` output changes are the intended ones"** — there is nothing to check, and that is verifiable rather than hopeful. `read_efuse_txpwr_base_8822e()` returns early unless `_variant == C8822E`, and on 8822E the walk takes the EU branch, which `return`s before the removed exit ever ran. So that reader's output is bit-identical on both variants. The complete consumer set of the changed (non-EU) walk is `read_efuse_rfe_type`, `probe_efuse_map`, and #383's MAC read. `probe_efuse_map`'s compare surface does grow — it now sees the whole map instead of the first three sections — which strictly improves what the stability probe is checking.

**The RFE blast radius, measured offline.** I walked all five 8822C tables twice through `PhyTableLoaderJaguar3::Load` with `rfe_type = 0` and `= 3` and diffed the register-write streams:

| table | writes rfe0 / rfe3 | distinct addrs | changed | added/removed |
|---|---|---|---|---|
| phy_reg | 1289 / 1289 | 923 / 923 | 0 | 0 |
| agc_tab | 450 / 450 | 2 / 2 | 0 | 0 |
| radioa | 789 / 789 | 55 / 55 | **5** | 0 |
| radiob | 697 / 697 | 37 / 37 | **2** | 0 |
| cal_init | 2464 / 2464 | 44 / 44 | 0 | 0 |

Identical counts and identical address sets everywhere — no row appears or disappears. The entire delta is 7 RF register values:

```
radioa  0x52: 0x000942ca -> 0x000902ca
        0x63: 0x00000c02 -> 0x00000002
        0xb3: 0x0007c760 -> 0x000fc760
        0xb6: 0x000187f8 -> 0x000387f8
        0xdd: 0x00000500 -> 0x00000540
radiob  0x52: 0x000942c0 -> 0x000902ca
        0x63: 0x00000c02 -> 0x00000002
```

Cut-independent (identical for `cut_version` 0–3). BB, AGC and RFK cal-init are untouched, so the on-air surface is RF front-end path config only — which is exactly what an RFE type is supposed to select. Happy to land the harness as a ctest if you want it as a standing check; it needs no hardware.

**On-air A/B** — I don't have an SDR here, so I can't produce the per-rate measurement you asked for. What I can run on the affected 8822CU is a link-level A/B: fixed TX power and channel, N frames per MCS, before/after, comparing per-MCS PDR and receiver RSSI. Say the word and I'll post it. Worth noting the direction of travel independently of any measurement: `0x03` is what the vendor driver programs on that same unit, so the fix moves devourer's RF config onto the kernel's rather than away from it — today those units run the `rfe_type=0` column by accident.

## Interaction with #383

They conflict, in both directions — GitHub reports each as mergeable against master because it only tests them pairwise with the base. #383 calls `read_efuse_logical_map(map, sizeof(map), kMacLogicalOff + 6)` in two places, and this PR removes that parameter, so whichever lands second needs the arg dropped at both call sites (`cache_efuse_8822e` and `perm_mac`) — a 2-token change, nothing semantic: with `upto` gone the 8822C on-demand walk simply becomes the full walk it wanted.

I merged both onto master locally and resolved it that way: build clean, 49/49. Merge them in whatever order suits you and I'll push the fixup to the trailing one immediately — or say the word and I'll rebase #383 onto this branch so it lands conflict-free.

josephnef (approved):
Approving with the on-air arm #385 asked for, run on this rig's affected RTL8822CU (the second of the 2/2 units that reproduce the truncated walk).

## On-air TX A/B (master vs this fix, txdemo MCS7/20 flood, SDR duty)

| band | master (`rfe_type=0`) | fixed (`rfe_type=0x03`) |
|---|---|---|
| ch36 (5 GHz) | 49.2% duty (~32.0 Mbps on-air) | 48.7% (~31.7 Mbps) — tie |
| ch6 (2.4 GHz) | 78.2–81.8% over 4 interleaved reps | **92.5–94.2%** |

The 2.4 GHz gap is reproducible (interleaved A/B ×3, per-arm sd < 1 point, far above this rig's ±3-point single-probe floor): the correct front-end rows are worth ~+12 points of duty at MCS7/ch6 on this unit — master was flying the unprogrammed-default tables. 5 GHz is unchanged.

## RX + identity regression checks

- RX smoke A/B on the same unit, ch6 ambient: 345 vs 408 clean frames / 8 s, 0 corrupt both arms, HEALTHY verdicts — the table change does not deafen RX.
- 8822CU efuse: stability probe id `0x8129` valid, `rfe_type=0x03` (kernel agrees), and the #383 MAC decodes to `a8:b5:8e:6a:94:ea` — the exact value the #385 worktree experiment predicted on this unit.
- 8822EU (`98:03:cf:cf:a4:49`), 8814AU (`20:0d:b0:c7:e4:b3`): unchanged, stable across runs.

The kPhysMax guard against the 10-bit address aliasing is a good catch — a truncated tail section would otherwise silently decode the start of the EFUSE into a wrong logical base.

CI note: the pushes landed during today's GitHub Actions incident and the `pull_request` events were dropped (no runs existed for either head SHA); both PRs were close/reopened to re-trigger. Merge order: this one first, then #383 rides clean.

src/jaguar3/HalJaguar3.cpp
```diff
@@ -631,6 +645,15 @@ void HalJaguar3::read_efuse_logical_map(uint8_t *map, size_t len, uint16_t upto)
...
   auto rd = [this, eu](uint16_t a) -> uint8_t {
+    /* A section straddling the end of the physical area would otherwise run
+     * `phys` past kPhysMax: the loop head checks it once per section, but a
+     * header + ext + four data words advance it up to ten more bytes. That
+     * matters because efuse_OneByteRead masks the address to 10 bits, so a
+     * read at 1024 aliases to 0 and would silently decode the START of the
+     * EFUSE into whatever logical base the truncated section named. 0xFF is
+     * what both walks already treat as end-of-map / skip. */
+    if (a >= kPhysMax)
+      return 0xFF;
```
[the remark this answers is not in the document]
josephnef: Dismissed — not a bug. `kPhysMax` is a `constexpr uint16_t` local; `a >= kPhysMax` only performs an lvalue-to-rvalue conversion yielding a constant expression, which is not an odr-use ([basic.def.odr]/5), and non-odr-used locals may be named in a lambda without capture ([expr.prim.lambda.capture]/3). Empirically: this exact code compiles clean across the full CI matrix — GCC, Clang, MSVC and MinGW.

Issue #233, opened 2026-07-10
RTL8733BU support — new halmac 87xx HAL (USB-first)


josephnef:
## Summary

Add userspace support for the **RTL8733BU** (USB) — a 1×1 (2T2R-capable) dual-band 802.11ac + BT5 combo that is a current sweet-spot FPV adapter for OpenIPC air/ground links. The motivating feature is the same one devourer already serves elsewhere: **5/10 MHz narrowband long-range video**.

This needs a **new chip HAL generation**, not a variant of an existing one.

## Why a new HAL (not a Jaguar2/3 variant)

Investigation of the vendor tree (`libc0607/rtl8733bu-20240806`) established that 8733B is **halmac `87xx`** (`hal/halmac-rs/halmac-rs.mk`: `series:=87xx ic:=8733b`; siblings 8812F/8814B/8197F — none in devourer). Devourer's Jaguar2 (8822B/8821C) *and* Jaguar3 (8822C/8822E) are all halmac **`88xx`-common**. Different MAC generation ⇒ it cannot slot in as a `ChipVariant`; it gets its own HAL dir dispatched by chip-id like every generation.

**De-risking discovery:** 8733B narrowband uses the exact **`R_0x9b0` / `R_0x9b4`** re-clock dividers as devourer's Jaguar3 (`hal/phydm/rtl8733b/phydm_hal_api8733b.c:1316–1400`), *not* the Jaguar2 `0x8ac` path. So although the MAC layer is new (87xx), the RF/narrowband layer is a Jaguar3 twin — `RadioManagementJaguar3` is a direct port reference for the feature that matters most.

## Scope

**In:** RTL8733BU (USB) RX + TX on 2.4/5 GHz at 20/40/80 MHz **and 5/10 MHz narrowband**; BT-combo coex runtime; TX-power knobs; FastRetune; AdapterCaps; demos + CI configs. Neutral naming `src/rtl8733b/`, CMake option `DEVOURER_8733B` / define `DEVOURER_HAVE_8733B`.

**Out (follow-ups):** SDIO transport + 8733BS (needs a brand-new SDIO `IRtlTransport` — separate linked issue); PCIe (no 8733 PCIe part); WoWLAN firmware.

## Confirmed vendor facts

- RF type: `hal/rtl8733b/rtl8733b_ops.c:196` — `BIT_RF_TYPE_ID_8733B ? 2T2R : 1T1R` (runtime, from a reg bit).
[...]

PR #388, opened 2026-08-10
rtl8733b: add RTL8731BU/RTL8733BU USB RX and TX

snokvist:
## Summary

Adds a dedicated userspace HALMAC 87xx backend for the RTL8731BU/RTL8733BU Wi-Fi function.

- Recognizes vendor USB IDs `0bda:f72b` and `0bda:b733`, while requiring live RTL8733B chip ID `0x16`.
- Implements card power sequencing, cut-selected firmware boot, HALMAC/MAC and EFUSE setup, generated BB/RF/RFK tables, dual-band channel setup, monitor RX, raw TX, thermal telemetry and safe teardown.
- Advertises the hardware-tested surface only: 1x1 legacy OFDM + HT MCS0-7, BCC, 20/40 MHz, 2.4/5 GHz; long-preamble CCK is accepted only on 2.4 GHz at 20 MHz.
- Adds reproducible, hash-pinned firmware/table extraction, parser/descriptor/EFUSE/table self-tests, a bounded lifecycle harness, `rtl8733bprobe`, adapter-doctor integration and capability documentation.
- Integrates RTL8733B with the cross-generation `rx.parse_abort` telemetry added in #387.

This progresses the USB-first portion of #233. It does not auto-close that issue because narrowband RF validation and combo-module/coexistence work remain outside this PR.

## Why a separate backend

RTL8733B is HALMAC 87xx, not a Jaguar2/3 variant. Its power, firmware, MAC, RX/TX descriptor and PHY paths therefore remain isolated under `src/rtl8733b/`.

Both reviewed vendor trees describe this device as 1SS 802.11b/g/n with a 20/40 MHz capability mask. The 2024 tree also contains a later `Fix VHT flags` commit that removes the accidentally enabled VHT flag. This PR consequently does not adopt #233's provisional VHT/80 MHz/2T2R assumptions.

Firmware and tables are pinned to [libc0607/rtl8733bu-20230626](https://github.com/libc0607/rtl8733bu-20230626) commit `9e5f6845b4e3393207720fc1886668bc80b6af1c`. The implementation was compared with [libc0607/rtl8733bu-20240806](https://github.com/libc0607/rtl8733bu-20240806) commit `2ec19e154cffbc2abd98d43d59278dffa6e50d49`, including its active-path TSSI, SRRC band-edge, GPIO/coexistence, power-saving and C2H/TX-tasklet changes. The already hardware-validated artifacts were not silently replaced.

## Hardware tested

Test device: one `0bda:f72b`, RTL8733B cut D, USB high speed. Independent witness: RTL8812AU in monitor mode.

Verified through the normal `WiFiDriver` / `IRtlDevice` path:

- USB discovery/endpoints, repeated power/firmware/MAC/BB/RF bring-up, stable EFUSE reads and permanent MAC extraction.
- Ambient monitor RX on channels 6 and 36, descriptor aggregation, PHY status, CRC reporting and clean stop.
- A 60-second high-traffic RX run exposed a 20 KiB device-aggregate versus 16 KiB host-URB split. Capping RX aggregation at 12 KiB fixed it; five repeated high-traffic channel-36 runs then had zero malformed descriptors, aggregate mismatches or parser aborts.
- Raw TX independently captured on both bands: legacy OFDM and HT MCS0-7 at 20 MHz, plus HT 40 MHz. BCC was forced throughout.
- Long-preamble CCK independently captured at 1, 2, 5.5 and 11 Mbps on channel 6. A mixed CCK/OFDM run submitted 650/650 frames; 636 were captured with valid FCS across all requested rates. A 1 Mbps request on channel 36 was rejected before USB TX.
- A rapid transition run submitted 1,000/1,000 frames through 118 CCK/OFDM TSSI table changes with zero USB failures. The witness captured 660 canonical frames across both rates, all with valid FCS; the thermal code stayed at 32-33 against EFUSE baseline 32.
- Ten bounded warm lifecycle cycles rotating channels 6/36 passed EFUSE, firmware, initialization, RX parsing, thermal, rollback, card-disable and USB-persistence gates.
- Concurrent RX/TX, bounded sustained TX, malformed-input rejection, teardown during active work, SIGINT during initialization and explicit failure cleanup passed.
- Correctly targeted post-stress adapter doctor: HEALTHY; 4/4 identical EFUSE reads, firmware ready, 91 clean + 34 corrupt ambient frames, 125 frames total with zero malformed/aggregate errors.
- Final explicit card-disable readback: `RF_CTRL=0x00`, `CR=0xea`.

## Checks

- Full default build: passed.
- Full CTest matrix: **53/53 passed**.
- RTL8733B-only build (all other chip options disabled): passed.
- RTL8733B-only CTest matrix: **49/49 passed**.
- RTL8733B-only ASan/UBSan build and CTest matrix: **49/49 passed**.
- `bash -n`, ShellCheck and failure-cleanup exercise for `tests/rtl8733b_lifecycle_soak.sh`: passed.
- Firmware extractor `--check`: reproduced both checked-in NIC images byte-for-byte.
- Table extractor `--check`: reproduced all checked-in MAC/BB/AGC/RF/RFK arrays byte-for-byte.
- Based on current `OpenIPC/devourer:master` (`da06ccc`) for the final local builds and hardware checks.

## Explicitly not tested or claimed

- **No SDR equipment was available.** Occupied bandwidth, spectral mask, EVM, absolute output power and SDR duty measurements were not performed.
- Experimental 5/10 MHz has register-readback and normal-path RX evidence only. Narrowband TX and independent narrowband-peer decode are deferred; `narrowband_ok` remains false.
- No physical `0bda:b733` combo module was available. Its ID is from the vendor table; Bluetooth/coexistence is not implemented or claimed.
- SGI remains disabled: descriptors with the short-GI bit set submitted, but the RTL8812AU witness decoded both broadcast and unicast probes as long GI.
- STBC and LDPC are rejected. The vendor reports TX-LDPC but not RX-LDPC; this raw-injection path has only been validated with global BCC forced, and the known injection defect makes enabling LDPC here unjustified.
- CCK short preamble, CCK outside 2.4 GHz/20 MHz, VHT, 80 MHz, ACK/BlockAck, A-MPDU, fast retune and throughput are not claimed.
- The hub cannot switch VBUS, so automated repeated true-cold boots were not possible; physical replug and warm reinitialization were tested.
- A vendor-kernel-driver cross-test was attempted with the 2024 tree on Linux 6.14. Its build first omitted required include paths, then failed on incompatible `MODULE_IMPORT_NS` syntax and missing `platform_ops.h`. No vendor module was loaded, so no vendor-driver parity result is claimed.
- An RTL8731BU-labelled `f72b` passed early identity/EFUSE/firmware checks, but later stopped enumerating and overheated while attached to a custom power supply. It is excluded from all radio-validation claims.

The full tested/deferred matrix and provenance are in `docs/rtl8733b.md`.


src/AdapterCaps.h
```diff
@@ -77,7 +80,8 @@ inline uint8_t bw_mask_for_generation(ChipGeneration g) {
    * bw_sup declares BW_CAP_5M|10M); 160 MHz is 8852C-only (rtl8852c_halinit.c
    * bw_sup has BW_CAP_160M, rtl8852b_halinit.c tops at 80) and is OR'd in by
    * the device layer per variant. */
-  return g == ChipGeneration::Jaguar1  ? ac
+  return g == ChipGeneration::Rtl8733b ? (kBw20 | kBw40)
+         : g == ChipGeneration::Jaguar1  ? ac
          : g == ChipGeneration::Unknown ? 0
                                         : (ac | kBw5 | kBw10);
```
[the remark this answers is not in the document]
snokvist: Intentionally not changed. This PR advertises only capabilities that were validated: 20/40 MHz. We have no SDR equipment or independent narrowband peer available, and the pinned/newer RTL8733BU vendor HALs declare BW20/BW40 for this 1SS HT part. Advertising 5/10 MHz or FastRetune now would turn an unverified future experiment into a false API promise. Narrowband TX and peer-decode validation are explicitly deferred in the PR description and docs.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -0,0 +1,608 @@
...
+  caps.chip_name = "RTL8733B";
+  caps.marketing_names = "RTL8731BU/RTL8733BU";
+  caps.chip_id = rtl8733b::kChipId;
+  caps.generation = devourer::ChipGeneration::Rtl8733b;
+  caps.variant = "cut-selected";
+  caps.transport = _device.is_usb() ? "usb" : "unknown";
+  caps.tx_chains = 1;
+  caps.rx_chains = 1;
+  caps.tx = GetTxCaps();
+  caps.bw_mask = devourer::bw_mask_for_generation(caps.generation);
```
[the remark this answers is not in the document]
snokvist: Intentionally not changed. narrowband_ok=false and fastretune_ok=false are conservative, truthful capability reporting for the tested surface. The 1x1 chain count is also deliberate: both the pinned and newer vendor HAL identify RTL8733B as 1SS, and the tested f72b hardware exposes one RF path. Runtime inference is not needed to avoid misreporting here; reporting 2T2R would contradict the vendor implementation and observed device.

src/rtl8733b/TxDescriptor8733b.h
```diff
@@ -0,0 +1,132 @@
...
+  p[0] = static_cast<uint8_t>(value);
+  p[1] = static_cast<uint8_t>(value >> 8);
+  p[2] = static_cast<uint8_t>(value >> 16);
+  p[3] = static_cast<uint8_t>(value >> 24);
+}
+
+inline void txdesc_set_bits(uint8_t *p, unsigned bit, unsigned width,
+                            uint32_t value) {
+  const uint32_t field = ((1u << width) - 1u) << bit;
+  txdesc_write_le32(p, (txdesc_le32(p) & ~field) |
```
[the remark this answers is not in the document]
snokvist: Fixed in c94daba. txdesc_set_bits now treats width 0 and out-of-range fields as no-ops and special-cases width 32 without shifting by the type width. The selftest covers full-width writes plus zero-width and overflowing field requests.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -0,0 +1,608 @@
...
+    uint8_t advertised = 0;
+    uint32_t packets = 0;
+    while (offset + rtl8733b::kRxDescSize <= static_cast<size_t>(length)) {
+      rtl8733b::RxFrame8733b frame{};
+      if (!rtl8733b::parse_rx_8733b(data + offset,
+                                    static_cast<size_t>(length) - offset,
+                                    frame)) {
+        ++malformed;
+        devourer::emit_rx_parse_abort(
+            _logger->events(), data + offset,
```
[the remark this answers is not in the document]
snokvist: Fixed in c94daba. Zero-padding classification is now a sink-independent helper; RTL8733B increments malformed and emits rx.parse_abort only for a non-zero failed remainder. The RX selftest covers both all-zero padding and a non-zero malformed remainder.

src/rtl8733b/Halmac8733bMac.cpp
```diff
@@ -0,0 +1,791 @@
...
+}
+
+void Halmac8733bMac::stop() {
+  _device.rtw_write32(kRegRcr, 0);
+  _device.rtw_write8(
+      kRegTxdmaPqMap,
+      static_cast<uint8_t>(_device.rtw_read8(kRegTxdmaPqMap) & ~(1u << 2)));
+  _device.rtw_write16(kRegTxPause, 0xffff);
+  _device.rtw_write8(kRegCr, 0);
+}
```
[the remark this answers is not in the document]
snokvist: Fixed in c374834. Halmac8733bMac::stop now clears the same 16-bit CR width used by monitor enable. The existing rtl8733bprobe MAC lifecycle gate now reads the full 16-bit CR before reinitialization. Verified with the full 53-test CTest matrix and on the cut-D 0bda:f72b hardware: stopped_cr=0x0000, immediate reinit=true, then card-disable reached CR=0xea and RF_CTRL=0x00.

examples/rx/main.cpp
```diff
@@ -1416,36 +1439,51 @@ int main(int argc, char **argv) {
...
+          ev.f("baseline", nullptr);
+        ev.f("status", ThermalBucket(t));
+        if (t.valid && t.delta >= g_thermal_warn_delta && !warned) {
+          logger->warn("thermal: raw={} baseline={} delta={} status={}", t.raw,
+                       t.baseline, t.delta, ThermalBucket(t));
+          warned = true;
+        } else if (!t.valid || t.delta < g_thermal_warn_delta) {
+          warned = false;
+        }
+        if (t.valid && t.delta >= 25) {
```
josephnef: **[major] Hardcoded thermal kill-switch now applies to every generation.** On master the Jaguar1-only poller just warned; this rewrite makes any adapter running with `DEVOURER_THERMAL_POLL_MS` self-terminate RX at delta ≥ 25, with the threshold hardcoded (`DEVOURER_THERMAL_WARN_DELTA` configures the warn, not the kill). CLAUDE.md documents this var as pure telemetry (“emit `thermal` events”), and `docs/warm-tx-degradation.md` warns the meter is not a validated degradation predictor. It's extra sharp on Jaguar3 8822C, where the baseline is a first-read cold reference rather than an EFUSE constant, so a legitimately hot long session can cross 25.

Please drop the cross-generation abort, or gate it behind an explicit opt-in env with a configurable threshold.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -0,0 +1,647 @@
...
+    _channel = channel;
+    _rx_configured_bw = channel.ChannelWidth == CHANNEL_WIDTH_40 ? 1 : 0;
+    if (!configure_tx_power(channel))
+      throw std::runtime_error("RTL8733B safe TX power setup failed");
+    if (_cfg.rx.enable_with_tx &&
+        !_mac.configure_monitor_rx(_cfg.rx.keep_corrupted))
+      throw std::runtime_error("RTL8733B monitor RX configuration failed");
+    const auto thermal = GetThermalStatus();
+    if (thermal.valid && thermal.delta >= 25)
+      throw std::runtime_error("RTL8733B thermal gate refused TX");
```
josephnef: **[major] Hardcoded, unconfigurable thermal TX-refusal knob.** The `delta >= 25` threshold appears here and in `send_packet` as a magic number with no `DeviceConfig` field — a per-chip behavioral sentinel no other generation has (project rule: knobs mean the same thing on every chip). As written, a warm chip refuses `InitWrite` entirely and the caller has no way to turn it off. If the PA-protection intent is real, make it a `DeviceConfig` field (with the refusal an explicit opt-in or at least tunable), and consider degrading the `send_packet` path to a rate-limited warn.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -0,0 +1,647 @@
...
+    return false;
+  }
+  if (packet == nullptr)
+    return false;
+  const auto thermal = GetThermalStatus();
+  if (thermal.valid && thermal.delta >= 25) {
+    _logger->error("RTL8733B TX thermal gate: raw={} baseline={} delta={}",
+                   thermal.raw, thermal.baseline, thermal.delta);
+    return false;
+  }
```
josephnef: **[major] Thermal poll on every `send_packet`.** `GetThermalStatus()` → `read_thermal()` is 3 RF writes + a 15 µs sleep + 1 RF read — each RF access itself several USB control transfers — per frame, under `_reg_mu`, on a USB-HS part. No other generation does register I/O on the send hot path; at injection rates this dominates the per-frame budget. Rate-limit the poll (every N frames or a wall-clock deadline).

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -0,0 +1,647 @@
...
+
+  const size_t frame_offset =
+      rtl8733b::kTxDescSize + static_cast<size_t>(packet_offset) * 8u;
+  std::memcpy(out + frame_offset, dot11, frame_len);
+  return frame_offset + frame_len;
+}
+
+void Rtl8733bDevice::SetCcaMode(bool disabled) {
+  (void)disabled;
+  radio_operation_unavailable("CCA control");
```
josephnef: **[major] `SetCcaMode(false)` powers the device off and throws.** `radio_operation_unavailable` calls `Stop()` (full MAC stop + card power-off) before throwing — but `false` requests the *universal default* (carrier-sense enabled). Refusing it destructively breaks real callers: `examples/timesync/main.cpp` calls `SetCcaMode` unconditionally, so timesync on this backend card-disables the chip mid-setup. `false` should succeed as a no-op; only `true` may legitimately fail — and arguably without tearing down the session (the pure-virtual interface exists to prevent silent no-ops, not to mandate teardown).

tools/extract_8733b_tables.py
```diff
@@ -0,0 +1,187 @@
...
+from __future__ import annotations
+
+import argparse
+import hashlib
+import re
+import struct
+from pathlib import Path
+
+UPSTREAM = "libc0607/rtl8733bu-20230626 commit 9e5f684"
+DEFAULT_ROOT = "reference/rtl8733bu-20230626"
```
josephnef, line 18: **[major] Vendor tree is not a `reference/` submodule — `--check` is irreproducible from a fresh checkout.** Both extractors read `reference/rtl8733bu-20230626/`, but the PR adds no `.gitmodules` entry and `reference/README.md` isn't updated; the repo convention is pinned submodules (“`git submodule update --init --recursive` before running `tools/extract_*.py`”). Without it, neither the extractor check nor the many hand-transcribed `halmac_reg2.h` constants in `Halmac8733bMac.cpp` can be re-verified by a reviewer or CI — and the missing-input error message doesn't name the pinned commit. Please add the submodule pinned at `9e5f684` (and the 2024 tree if you want the comparison on record), or document the deviation with the exact commit in `reference/README.md`.

(Credit where due: the per-source and per-array SHA-256 pinning with byte-for-byte `--check` and zero network access is the best extractor hygiene in `tools/` — I recomputed all 9 array hashes against the checked-in files and they match.)

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -0,0 +1,647 @@
...
+      if (advertised != 0 && packets >= advertised)
+        break;
+    }
+    if (advertised != 0 && packets != advertised)
+      ++aggregate_mismatch;
+  };
+
+  int urb_bytes = _cfg.rx.urb_bytes.value_or(16 * 1024);
+  if (urb_bytes < 4096)
+    urb_bytes = 4096;
```
josephnef: **[minor] `urb_bytes` floor undercuts the 12 KiB aggregate cap.** The floor is 4096 while the device-side cap is a fixed `kUsbRxAgg8733b` (12 KiB) — a configured `DEVOURER_RX_URB_BYTES` in 4096..12287 lets an aggregate span two URBs, recreating exactly the straddle failure this PR fixed (the walk won't read out of bounds, but every split aggregate becomes a malformed abort + lost frames). Clamp the floor to 12288, or derive the aggregation threshold from `urb_bytes`.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -0,0 +1,647 @@
...
+  const uint8_t endpoint = _device.first_bulk_out_ep();
+  const uint8_t rate_hw = usb_frame[0x10] & 0x7f;
+  const uint8_t rate_id = usb_frame[0x06] & 0x1f;
+  const uint8_t bandwidth = (usb_frame[0x14] >> 5) & 0x3;
+  const uint8_t short_gi = (usb_frame[0x14] >> 4) & 0x1;
+  const uint8_t ldpc = (usb_frame[0x14] >> 7) & 0x1;
+  if (!select_tssi_rate_table(rate_hw <= 3)) {
+    _logger->error("RTL8733B TX rejected: TSSI rate-table transition failed");
+    Stop();
+    return false;
```
josephnef, line 398: **[minor] Fatal path returns plain `false` after tearing the session down.** A TSSI table-switch failure calls `Stop()` (MAC stop + card disable) but returns an ordinary `false` — indistinguishable from a transient short bulk write — and every later send just logs “TX rejected before InitWrite”. The PA-safety intent is plausible but undocumented at the `Stop()` site; retry loops will be confused. Comment the why, and consider throwing like the other fatal paths.

src/rtl8733b/Halmac8733bMac.cpp
```diff
@@ -0,0 +1,791 @@
...
+  const size_t calibration_end = kPhysicalEfuseSize - kReservedEfuseSize;
+  for (size_t address = calibration_begin; address < calibration_end;
+       ++address) {
+    if (!read_physical_byte(static_cast<uint16_t>(address), map[address]))
+      return false;
+  }
+  return true;
+}
+
+bool Halmac8733bMac::parse_physical_efuse(const uint8_t *physical,
```
josephnef, line 227: **[minor] `parse_physical_efuse` hard-fails a fully-used map.** The walk only returns `true` via an explicit `0xff` terminator; running off the end of the readable span returns `false`, which `read_efuse` escalates to a bring-up abort (and leaves `*used_bytes` unset on that path). Vendor EFUSE walkers treat address-space exhaustion as success-with-whatever-parsed — a heavily reprogrammed but valid EFUSE bricks the backend here.

src/rtl8733b/Rtl8733bBringup.cpp
```diff
@@ -0,0 +1,558 @@
...
+    w8(0x0077, 0x00);
+    w8(0x001f, 0x87);
+    w8(0x0077, 0x87);
+    w8(R8733_SYS_STATUS1 + 1,
+       static_cast<uint8_t>(r8(R8733_SYS_STATUS1 + 1) & ~0x01u));
+  }
+
+  init_system_cfg();
+  _powered = r8(R8733_CR) != 0xea && (r8(R8733_SYS_STATUS1 + 1) & 1u) == 0;
+  _logger->info("RTL8733B power: card active, CR=0x{:02x} SYS_STATUS1=0x{:08x}",
```
josephnef: **[minor] `power_on()` logs “card active” unconditionally, then may return false.** On a unit where the readback check fails, the diagnostic stream asserts “card active” immediately before the caller throws “card-enable sequence failed” — misleading during exactly the triage this log exists for. Log the computed `_powered` (or split success/failure lines).

Related: if `power_on()` fails mid-sequence (several card-enable writes already landed), `_power_ready` stays false so `Stop()` skips `power_off()` — no rollback, and the next attempt relies on the sequence being re-enterable. Worth a comment at minimum.

docs/rtl8733b.md
```diff
@@ -0,0 +1,136 @@
...
+  20 MHz and HT 40 MHz. BCC was used throughout. On 2.4 GHz, an RTL8812AU
+  witness also decoded long-preamble 1, 2, 5.5 and 11 Mbps CCK with valid FCS.
+- Repeated warm initialization, concurrent RX/TX, bounded sustained TX,
+  teardown during active RX/TX, malformed-input rejection and post-test
+  adapter-doctor checks.
+
+A 60-second high-traffic receive run exposed a real USB aggregation boundary
+fault: the vendor-default 20 KiB aggregate could straddle the backend's 16 KiB
+URBs, leaving a descriptor tail and body in separate completions. RX aggregation
+is now capped at 12 KiB. Five repeated high-traffic channel-36 runs after that
```
josephnef: **[minor] Changelog-style phrasing — docs are current-state only.** “is **now** capped at 12 KiB” here, “the capability claims **in this PR**” (line 131) and the #233 framing (line 134) are PR-relative; the repo convention is present-tense facts (“RX aggregation is capped at 12 KiB because …”) — git is the changelog. Content is otherwise exemplary: every favourable measurement is paired with its adversarial counterpart.

README.md
```diff
@@ -101,6 +102,7 @@ Bandwidth cells are devourer's measured on-air TX throughput (Mbps, HT MCS7,
 | **RTL8822CU**                 | 2T2R + BT         | —             | —             | —                | —                | not benchmarked (`0bda:c82c`). 5/10 MHz capable |
 | **RTL8812EU**                 | 2T2R              | ‡             | 51            | 47               | —                | LB-LINK BL-M8812EU2 (`0bda:a81a`); bare 5 GHz FPV module. 5/10 MHz capable. ‡ 2.4 GHz TX airs energy but no receiver decodes it — the vendor kernel driver behaves identically on this module ([quirks](docs/8822e-quirks.md)) |
 | **RTL8822EU**                 | 2T2R + BT         | —             | —             | —                | —                | not benchmarked. 5/10 MHz capable |
+| **RTL8731BU / RTL8733BU**     | 1T1R              | —             | —             | —                | —                | `0bda:f72b` cut D: monitor RX and CCK/legacy/HT raw TX validated on 2.4 GHz; legacy/HT validated on 5 GHz, at 20/40 MHz where legal; not benchmarked. `0bda:b733` is recognized from the vendor ID table but lacks a physical test. [Status and test limits](docs/rtl8733b.md) |
```
josephnef: **[request] Please split this row into separate `RTL8731BU` and `RTL8733BU` rows to match the table's schema.** The rest of the table keeps one chip per row even when they share a code path (e.g. RTL8812EU vs RTL8822EU, RTL8852BU vs RTL8832BU), with each row stating what was physically validated on *that* chip and naming the DUT adapter model used — e.g. “LB-LINK BL-M8812EU2 (`0bda:a81a`); bare 5 GHz FPV module”.

Concretely: which chip was the tested `0bda:f72b` cut-D unit — and what adapter/module model is it (brand + product, as with the other rows)? That row carries the validated surface; the other chip's row should say it's recognized from the vendor ID table only, no physical test (as you already state for `0bda:b733`). The RTL8731BU-labelled unit that overheated and stopped enumerating is also worth a word in `docs/rtl8733b.md`'s matrix about which row it would have covered.

tests/rtl8733b_lifecycle_soak.sh
```diff
@@ -0,0 +1,180 @@
...
+#   COOLDOWN_SECS         delay between cycles (default: 1)
+#   MAX_THERMAL_DELTA     maximum raw delta from EFUSE baseline (default: 20)
+#   MAX_THERMAL_SPREAD    maximum raw spread over all cycles (default: 12)
+#   OUT                   output directory (default: timestamp under /tmp)
+#
+# A SUSPECT doctor verdict is accepted only for reason 0x40 (no clean ambient
+# traffic without --expect-traffic). All hardware/firmware/EFUSE invariants
+# must still pass. Exit 0 means every requested lifecycle passed.
+
+set -u
```
josephnef: **[minor] Script hardening.** Three small ones: (1) `set -u` only — 27 existing `tests/*.sh` use `set -euo pipefail`; the rc handling here is manual and mostly sound, but e.g. `mkdir -p`/`sha256sum` failures fall through. (2) Line 137: `grep -qF '"efuse_reads":'"$READS"` is an unanchored substring match — `READS=4` also matches `"efuse_reads":40`; anchor the delimiter or use `tests/devourer_events.py`, which exists for exactly this field scraping. (3) Lines 88-89: the sysfs VID/PID compare is case-sensitive — sysfs is lowercase, so `VID=0x0BDA` fails verification of a present, correct device.

Otherwise solid: cleanup trap with explicit card-disable, `timeout -k` on every hardware invocation, bounded per-rep wall clock.

tests/rtl8733b_tx_desc_selftest.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+int main() {
+  using rtl8733b::TxDescConfig;
+  std::array<uint8_t, rtl8733b::kTxDescSize> desc{};
+
+  for (uint8_t rate = 0; rate <= 3; ++rate) {
+    expect("2G CCK rate ID",
+           rtl8733b::tx_rate_id_8733b(rate, 0, false) == 8);
+    expect("5G CCK rejected",
+           rtl8733b::tx_rate_id_8733b(rate, 0, true) == 0xff);
```
josephnef, line 28: **[minor] CCK-on-5 GHz rejection chain isn't closed end-to-end.** This asserts `tx_rate_id_8733b(...) == 0xff`, but nothing asserts that `fill_tx_desc_8733b` rejects `rate_id = 0xff` (the actual enforcement is the `cfg.rate_id <= 0x1f` clause in `valid_tx_desc_config`) — one extra `fill` call with the sentinel makes the band gate a tested property instead of two disconnected halves. Related coverage gap: STBC/VHT rejection lives only in the device radiotap path (`Rtl8733bDevice.cpp`) and has no headless test; the doc narrative is currently the only evidence.

josephnef (commented):
## Review

**Overall: high-quality backend — the code reads like the existing HALs, the RX parse is genuinely bounds-safe, caps are honest, and the extractors are the best hash-pinning in `tools/`. A handful of items below are worth resolving before merge; three of them cluster around one design decision (the hardcoded thermal-policy layer), plus the destructive `SetCcaMode` refusal.**

### Verified during review (not just read)

- Both ON and 8733B-only builds compiled here; ctest 53/53 and 49/49 pass, matching the PR claims.
- `git diff master...HEAD -- src/jaguar1 src/jaguar2 src/jaguar3 src/kestrel` is empty — no existing HAL touched. The `RxParseAbort.h` refactor is behavior-preserving for the four existing generations (checked combination-by-combination).
- Dispatch is conservative in exactly the right way: chip-id `0x16` required, a known 8733B PID with a failed/wrong chip-id read is refused rather than falling through to Jaguar1, and `DEVOURER_8733B=OFF` returns nullptr+log on both paths.
- All 9 generated-array SHA-256 pins in the extractors were recomputed against the checked-in `hal/hal8733b_*.c` — all match, and the tools do no network access.
- The 12 KiB RX-aggregate cap is real, enforced at init and readback-verified via `MacState::matches_normal_usb3out()`.
- EFUSE packed-map decoding (extended-header block math, active-low word-enable, bounds on truncated headers) was hand-checked against vendor semantics; the TX golden descriptors were hand-decoded field-by-field.
- The `examples/rx` rework fixes a real pre-existing bug (the qd/thermal emitter threads were never stopped/joined — a `std::terminate` on early-return paths).

### Findings not tied to a single diff line

- **8733B PIDs missing from the demo discovery lists**: `examples/doctor` gained `0xf72b`/`0xb733` in `kRealtekPids`, but neither `examples/rx` nor `examples/tx` `kRealtekProductIds` did — a plugged 8733B is invisible to rxdemo/txdemo without `DEVOURER_PID=0xf72b`. Add to both or note why not.
- **`DEVOURER_TX_FRAMES`** (new txdemo demo-local var) is referenced nowhere — README/docs need a line.
- The first-8 RX completion/frame breadcrumbs log at `info`; convention puts per-frame diagnostics at debug/trace (bounded to 8, so cosmetic).
- `rtl8733bprobe`'s `add_executable` sits mid-Kestrel section in CMakeLists.txt, far from the other example targets — works, but breaks the file's grouping.
- Extractors use `read_text()`/`write_text()` without explicit `encoding=`/`newline=`: on Windows `--check` won't be byte-identical (CRLF). Pre-existing pattern in older extractors, but these otherwise raise the bar — worth the two keywords.
- Headless test gaps: STBC/VHT radiotap rejection lives only in `Rtl8733bDevice.cpp` and has no selftest; the RX walk's tail case (last frame's 8-byte-rounded `next_offset` past `buflen`) is untested.

The hardware-validation narrative in the PR description and `docs/rtl8733b.md` is unusually honest (SGI decode failure, the dead second unit, the failed vendor-module cross-test all disclosed) — appreciated.

examples/rx/main.cpp
snokvist: Dropped the abort entirely — the poller emits and warns, nothing more (`37a30ba`).

You were right that generalizing it onto `IRtlDevice::GetThermalStatus` was the point and the kill-switch just rode along. I also took your `docs/warm-tx-degradation.md` note as the general verdict rather than patching one call site, so the whole policy layer is gone:

- `InitWrite` no longer refuses TX; it keeps a one-shot bring-up snapshot, same shape as the Kestrel one.
- `send_packet` reads no register at all.
- `Phy8733b::enable_tssi_tracking` had a live `read_thermal()` inside its readback verdict — and that runs on every CCK↔OFDM rate-table transition, i.e. *inside* `send_packet`. Its verdict is register readback only now.
- `rtl8733bprobe` reports the "critical" bucket instead of aborting.

The one remaining threshold is `kTssiAuditThermalCeiling`, named and confined to the probe-only `audit_tssi_*` stages.

I also corrected `CLAUDE.md`, which called a rising delta "the early TX-degradation warning" — that claim is what the policy layer was built on and `docs/warm-tx-degradation.md` contradicts it.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Removed rather than made configurable (`37a30ba`). A `DeviceConfig` field would have kept a per-chip behavioural sentinel no other generation has; the honest fix was to match them. `InitWrite` keeps a one-shot snapshot for the log and nothing gates on it.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Removed, not rate-limited (`37a30ba`). Rate-limiting would have kept register I/O on a send path where no other generation has any.

Worth flagging: `send_packet` was not the only per-frame reader. `Phy8733b::enable_tssi_tracking` folded a `read_thermal()` into its readback verdict, and `select_tssi_rate_table` calls it on every CCK↔OFDM transition — so a mixed-rate run was paying for it there too. Both are gone; the meter tracks PA bias, not whether those BB writes landed.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Fixed (`37a30ba`). `false` returns as a no-op — it is the state the HALMAC bring-up already leaves programmed — and `true` no longer calls `Stop()`. `radio_operation_unavailable` had no other caller and is deleted, so nothing in this backend card-disables on a refusal any more.

`true` still throws, per your "only `true` may legitimately fail" and the pure-virtual contract. One correction to the example though: both `timesync` call sites are `SetCcaMode(true)`, not `false`, and both are behind `c.hwbeacon` — so timesync still raises there. What is fixed is that it no longer powers the chip down first.

tools/extract_8733b_tables.py
snokvist: Added as a pinned submodule (`9b18a28`): `reference/rtl8733bu-20230626` at `9e5f684`, branch `v5.13.0.1`, `shallow = true` to match the other pins, with `reference/README.md` updated — including why it matters beyond the extractors, since the `halmac_reg2.h` and power-sequence constants transcribed into `src/rtl8733b/` are now re-verifiable too.

Both extractors' missing-input errors name the pinned commit and print the exact `git submodule update --init` line.

Verified against the fresh submodule: `extract_8733b_fw.py --check` and `extract_8733b_tables.py --check` reproduce every checked-in artifact byte-for-byte.

The 2024 tree stays out of the submodule set and is recorded as a deliberate deviation in `reference/README.md` — nothing in `hal/` is generated from it.

(And thanks for recomputing all 9 array hashes.)

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Fixed by deriving the floor from the cap (`37a30ba`). The device-side ceiling is now `rtl8733b::kRxAggregateBytes8733b` (12 KiB) in `Halmac8733bMac.h`, the RX loop floors `urb_bytes` at it and warns when it raises a configured value, and a `static_assert` ties the constant to the programmed `RXDMA_AGG` page count so the two cannot drift:

```cpp
static_assert(kRxAggregateBytes8733b == ((kUsbRxAgg8733b & 0x0f) * 4096),
              "RX URB floor must match the programmed RXDMA_AGG page count");
```

`DEVOURER_RX_URB_BYTES` can still raise it, never lower it below one aggregate.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Documented, and the reason is now visible to the caller (`37a30ba`) — but I kept the `bool`.

Throwing would have made `send_packet` the only generation that throws per frame, and `txdemo`'s send loop does not catch. Instead the fatal reason is latched in `_tx_fatal` and cleared by `InitWrite`, so subsequent sends log "session stopped by a failed TSSI rate-table transition — re-run InitWrite to recover" instead of "TX rejected before InitWrite". A retry loop can now tell the two apart, which was the actual confusion.

The `Stop()` site carries the why: the table switch disables closed-loop tracking, rewrites the thermal table and re-enables it, so a failure means the readback no longer describes the loop driving the PA.

Happy to switch it to a throw if you would rather have the consistency with the other fatal paths than with the other generations' `send_packet`.

src/rtl8733b/Halmac8733bMac.cpp
snokvist: Fixed (`37a30ba`). Address-space exhaustion returns `true` with `*used_bytes = physical_len`; the early returns still reject truncated headers and out-of-range block targets. Added a selftest case for a terminator-less map in `tests/rtl8733b_efuse_selftest.cpp`.

src/rtl8733b/Rtl8733bBringup.cpp
snokvist: Both fixed (`37a30ba`).

The log now reports the computed `_powered` — `info` "card active" or `error` "card NOT active", same registers either way.

For the rollback I went further than a comment: `_power_ready = true` is now set *before* `power_on()` in `bring_up_to_phy`, with the reasoning at the assignment. A mid-sequence failure therefore rolls back through `power_off()` (the full HALMAC card-disable flow, safe against a partially enabled card) instead of leaving the next attempt to depend on the sequence being re-enterable from an undefined state.

docs/rtl8733b.md
snokvist: Fixed (`dc33c45`). "is now capped" → "is capped at 12 KiB because …", and the "in this PR" / #233 framing is gone. A grep for changelog phrasing across `docs/rtl8733b.md` and `README.md` is clean.

The thermal paragraph was rewritten anyway, since the behaviour it described no longer exists, and I added the `SetCcaMode` contract and pointed the extraction commands at the new submodule.

README.md
snokvist: Split, and thanks for pushing on the identity question — the answer changed what the rows say (`dc33c45`).

The tested `0bda:f72b` cut-D unit is an **RTL8731BU**, and it is a bare unbranded 1T1R module, not a retail adapter. So:

- **RTL8731BU** carries the validated surface, with the DUT named the way the other rows name theirs.
- **RTL8733BU** is now its own row: rides the RTL8731BU code path, `0bda:b733` recognized from the vendor ID table only, no physical unit, Bluetooth/coexistence not implemented.

The unit that overheated and stopped enumerating was the **same chip** as the survivor — so it would have covered the RTL8731BU row, and `docs/rtl8733b.md` now says exactly that: the row rests on a single surviving sample, a one-unit result rather than a population. The validation section also states up front that nothing was measured on an RTL8733BU.

tests/rtl8733b_lifecycle_soak.sh
snokvist: All three fixed (`68ea20f`), plus one your list implies.

1. `set -euo pipefail`. That surfaced a latent bug: `grep -qF '"ev":"rx.parse_abort"' "$log" && failed=1` exits non-zero on the *expected* no-match case, so under `set -e` it would have aborted every clean cycle. Rewritten as `! grep … || failed=1`. The doctor exit is captured with `|| doctor_rc=$?` and the `verdict_line` capture takes `|| true`, so a missing verdict stays a graded failure rather than an abort. `mkdir`/`sha256sum` now fail loudly.
2. Anchored: `grep -qE '"efuse_reads":'"$READS"'[,}]'`, and the same for the other numeric gates.
3. Canonicalised `VID`/`PID` to lowercase `0xNNNN` at startup and lowercase the sysfs read, so `VID=0x0BDA` (or a decimal `PID`) verifies correctly.

`bash -n` and `shellcheck` clean. Still only lint-checked — it has not been re-run against the DUT since these edits.

tests/rtl8733b_tx_desc_selftest.cpp
snokvist: Both closed (`68ea20f`, `03502d3`).

The sentinel chain is now end-to-end: for each CCK rate the test asserts `tx_rate_id_8733b` returns `0xff` (5 GHz and 40 MHz) **and** that `fill_tx_desc_8733b` refuses that config, plus an explicit out-of-range `rate_id = 0x20`.

For STBC/VHT I factored the admission contract out of `Rtl8733bDevice::build_tx_block` into pure predicates in `TxDescriptor8733b.h` — `ht_request_supported_8733b`, `legacy_request_supported_8733b`, `tx_mode_supported_8733b` — shared by the `SetTxMode` and radiotap branches so they cannot drift. The selftest asserts the second-stream, 80 MHz, SGI, LDPC, STBC and VHT/HE refusals directly. The descriptor encoder has no STBC field to reject on, so a predicate was the only way to make that half a tested property rather than a doc claim.

Remaining honest gap: the two `IEEE80211_RADIOTAP_VHT`/`_HE` case labels in `build_tx_block` still have no headless coverage — they sit inside the iterator switch — but they enforce the same contract `tx_mode_supported_8733b` states and tests.

The RX walk's tail case is covered too (`tests/rtl8733b_rx_parse_selftest.cpp`): a last frame whose 8-byte-rounded `next_offset` lands past `buflen` must still parse, and must terminate the walk.

snokvist:
Thanks — that was a genuinely useful review, and the thermal cluster was the right thing to lead with. Five commits pushed (`c374834..03502d3`), replies on each thread.

### The thermal layer is gone, not made configurable

Your three thermal findings shared one root cause, so rather than gate each one behind a knob I checked what the other backends actually do: thermal is pure telemetry everywhere — `GetThermalStatus()` on `IRtlDevice`, Kestrel logs one snapshot at `InitWrite`, nothing gates TX. This backend had grown a policy layer no other generation has, and the honest fix was to match them:

- rxdemo's poller emits and warns, never stops RX.
- `InitWrite` keeps a one-shot bring-up snapshot and refuses nothing.
- `send_packet` reads no register.
- **`Phy8733b::enable_tssi_tracking` folded a live `read_thermal()` into its readback verdict** — and `select_tssi_rate_table` calls it on every CCK↔OFDM transition, i.e. inside `send_packet`. So a mixed-rate run was paying the RF-read cost there as well as in the two places you flagged. Its verdict is register readback only now.
- `rtl8733bprobe` reports the "critical" bucket instead of aborting.

The only threshold left is `kTssiAuditThermalCeiling`, named and confined to the probe-only `audit_tssi_*` stages.

I also corrected `CLAUDE.md`, which described a rising delta as "the early TX-degradation warning". That claim is what the policy layer was built on, and `docs/warm-tx-degradation.md` contradicts it — delivery scattered 63–83% with no relation to the meter.

For the record on where the layer came from: an adapter overheated and died early in this bring-up (almost certainly a faulty bench supply, not the chip), and the guards were added to protect the hardware while new features were being written. They were development scaffolding that should not have shipped in the library.

### Non-line findings

| Finding | Status |
|---|---|
| 8733B PIDs missing from rxdemo/txdemo | Added `0xf72b` + `0xb733` to both `kRealtekProductIds` |
| `DEVOURER_TX_FRAMES` undocumented | Documented in `CLAUDE.md` beside `DEVOURER_TX_GAP_US` |
| First-8 RX breadcrumbs at `info` | Both the completion and frame breadcrumbs moved to `debug` |
| `rtl8733bprobe` mid-Kestrel in CMakeLists | Moved to sit with `pcieprobe`/`kestrelprobe` |
| Extractors' `read_text`/`write_text` | `encoding="utf-8", newline=""` on both, with the CRLF reasoning in a comment |
| Headless test gaps | Closed — see the two selftest threads |

### Two places I did not do what you suggested

- **The fatal TSSI path** kept its `bool` return rather than throwing: `send_packet` throws on no other generation and `txdemo`'s loop does not catch. The reason is latched in `_tx_fatal` instead, so later sends say the session was stopped rather than "TX rejected before InitWrite" — which was the actual confusion. Say the word if you'd rather have the consistency with the other fatal paths.
- **`SetCcaMode(true)`** still throws (non-destructively). Minor correction to the finding: both `timesync` call sites are `SetCcaMode(true)`, not `false`, and both sit behind `c.hwbeacon` — so timesync still raises there. What's fixed is that it no longer powers the chip down first.

### Verification

Headless only: full build **53/53**, RTL8733B-only **49/49**, RTL8733B-only ASan+UBSan **49/49**. Both extractors' `--check` reproduce every checked-in artifact byte-for-byte **from the newly pinned submodule**, and their missing-input paths were exercised. `bash -n` + `shellcheck` clean on the soak script.

**Nothing here has been re-tested on hardware.** The changes with real on-air exposure are the removal of the `enable_tssi_tracking` thermal term (which had been part of the pass criteria for the CCK/OFDM transition runs quoted in the PR body) and the RX URB floor. The soak script's rewrite is lint-checked only — it has not been run against the DUT since. I'll re-run the transition-churn and lifecycle soak on the RTL8731BU unit and post the results before you merge, unless you'd rather take it as-is.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Re-ran this on the DUT, and your wording turned out to be literally correct — the per-frame read was **93% of the send budget**.

Max-duty injection, 2000 frames, MCS4, `DEVOURER_TX_GAP_US=0`, RTL8731BU `0bda:f72b` cut D on USB-HS. Same binary and options either side; wall clock minus the measured `txdemo.first_tx_submit` mark, so chip bring-up is excluded. Two runs each:

| build | ms/frame | submission rate |
|---|---|---|
| `c374834` (pre-fix) | 2.707 / 2.708 | 369 fps |
| `03502d3` (fixed) | 0.200 / 0.196 | ~5060 fps |

**2.51 ms of a 2.71 ms per-frame budget**, so ~13.7x on submission rate.

The `enable_tssi_tracking` read costs separately, once per CCK<->OFDM table switch. 500 frames with every frame flipping the table:

| build | ms/transition |
|---|---|
| `c374834` (pre-fix) | 91.96 / 92.19 |
| `03502d3` (fixed) | 86.71 / 86.72 |

5.36 ms per transition.

Two caveats on those numbers. They are **submission** rate — bulk-OUT accepted by the chip, `failed=0` on all 2000 — not airtime; no witness adapter was plugged, so nothing here claims frames aired. And the residual ~84 ms per table switch is the switch itself (disable tracking -> rewrite the thermal table -> re-enable, all USB control transfers), which the thermal read was only ever a small part of. That caps a mixed CCK/OFDM stream at ~11.5 fps on this backend and is unrelated to this PR — worth its own issue if anyone wants to stream mixed rates.

Unlike a delivery percentage these are deterministic register-sequence timings, so the run-to-run spread is 0.001-0.04 ms/frame rather than the +/-3 points `CLAUDE.md` warns about.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Verified on the DUT: `DEVOURER_RX_URB_BYTES=4096` — the exact 4096..12287 window you identified — is now clamped rather than allowed to straddle.

```
devourer [W] RTL8733B RX: raising urb_bytes 4096 -> 12288 (device RX aggregate cap)
devourer [I] RTL8733B RX: entering async loop ch=36 urb=12288x8
devourer [I] RTL8733B RX: stopped completions=691 frames=1117 malformed=0 agg_mismatch=0
```

Zero `rx.parse_abort` events. `rtl8733bprobe mac` also confirms the device side still programs `RXAGG=0x2003` on hardware, which is what the `static_assert` pins the floor to.

tests/rtl8733b_lifecycle_soak.sh
snokvist: Re-ran the rewritten script on the DUT — 10/10 cycles, all HEALTHY, channels 6/36 alternating, thermal 32-33 against baseline 32 (the same range as the original run).

Two things worth reporting beyond "it passes":

**The anchoring change was load-bearing.** Real doctor output contains both `"efuse_reads":4,` and `"init":1}` — so the `[,}]` character class was necessary; anchoring on a comma alone would have broken the `init` gate.

**I ran a negative control**, because `set -euo pipefail` could plausibly have converted a graded failure into a silent abort. Forcing a doctor timeout (`LISTEN_SECS=8 TIMEOUT_SECS=10`):

```
[rtl8733b-soak] FAIL rep=1 rc=124 verdict=HEALTHY reasons=0x0 thermal=32/32
[rtl8733b-soak] failure cleanup: requesting explicit card-disable
```

exit 1, log tail printed, cleanup trap fired and the card-disable landed (`ready=1 CR=0xea RF_CTRL=0x00`). So the non-zero doctor exit is captured rather than aborting, and the failure path still works — note the verdict line itself read HEALTHY there, and the gate correctly failed on the exit code anyway.

Post-stress doctor after all of it: HEALTHY, 4/4 identical EFUSE reads, 686 clean + 161 corrupt ambient frames, clean card-disable.

snokvist:
## Hardware re-validation of `03502d3`

Ran on the RTL8731BU DUT as promised. DUT: bare unbranded 1T1R module, `0bda:f72b`, die `0x16` cut D, USB high speed, bus 1 port 1.2, no kernel driver bound (`rtw88` does not claim `f72b`). Everything below is against `03502d3`, with `c374834` — the pre-review PR head — built into a throwaway worktree as the before-side.

### Lifecycle soak — 10/10 PASS

`REPS=10 CHANNELS="6 36"`. All HEALTHY, thermal 32–33 vs baseline 32, the same range as the original run. Per-rep gates (EFUSE stability, fw-ready, init, TSSI rollback, `power-off ready=1 CR=0xea`, `RF_CTRL=0x00`, `malformed=0 agg_mismatch=0`) all matched real output — re-checked independently against every rep log, and zero `rx.parse_abort` across all ten.

The rewritten script got a **negative control** as well, since `set -euo pipefail` could have turned a graded failure into a silent abort: forcing a doctor timeout produced `FAIL rep=1 rc=124`, exit 1, and a cleanup-trap card-disable that landed. Details on that thread.

### Transition churn — 1000/1000, zero failures

Alternating `1M,6M,2M,MCS0,5.5M,MCS4,11M,MCS7` on ch6, so **every frame** forced a CCK↔OFDM table switch — 1000 transitions (500 each way) rather than the 118 in the original run, i.e. ~8.5× the churn. `submitted=1000 failed=0`, and zero of: `TSSI rate-table transition failed`, `TSSI tracking: enable readback failed`, `unsafe enable precondition`, `TX failed/short`, `TX rejected`.

That is the specific thing the change needed to prove: `enable_tssi_tracking`'s readback verdict, with the thermal term removed, held across 1000 consecutive transitions.

### What removing the reads actually cost — measured

Wall clock minus the measured `txdemo.first_tx_submit` mark (so bring-up is excluded; it was 2.22–2.26 s on both builds). Two runs each, interleaved.

**Per-frame read** — max duty, 2000 frames, MCS4, `DEVOURER_TX_GAP_US=0`:

| build | ms/frame | submission rate |
|---|---|---|
| `c374834` | 2.707 / 2.708 | 369 fps |
| `03502d3` | 0.200 / 0.196 | ~5060 fps |

**Per-transition read** — 500 frames, every frame a table switch:

| build | ms/transition |
|---|---|
| `c374834` | 91.96 / 92.19 |
| `03502d3` | 86.71 / 86.72 |

So the per-frame read was 2.51 ms of a 2.71 ms budget — @josephnef's "at injection rates this dominates the per-frame budget" was literally right, at 93% — and the per-transition read was a further 5.36 ms.

### RX URB floor — clamps on hardware

`DEVOURER_RX_URB_BYTES=4096`, the exact window that previously let an aggregate straddle two URBs:

```
devourer [W] RTL8733B RX: raising urb_bytes 4096 -> 12288 (device RX aggregate cap)
devourer [I] RTL8733B RX: stopped completions=691 frames=1117 malformed=0 agg_mismatch=0
```

### Post-stress health

Doctor after everything: HEALTHY, 4/4 identical EFUSE reads, 686 clean + 161 corrupt ambient frames on ch36, clean card-disable (`CR=0xea`, `RF_CTRL=0x00`).

### Not claimed

- **No witness adapter was plugged**, so nothing above is an on-air result. The TX numbers are *submission* rate — bulk-OUT accepted by the chip, `failed=0` — not airtime and not delivery. The independently-witnessed CCK/OFDM captures in the PR description still stand on the earlier runs; I have not re-witnessed them since these commits.
- No SDR, so the occupied-bandwidth / mask / EVM gaps are unchanged.
- Still warm-only — the hub cannot switch VBUS.

### One finding that is not about this PR

The residual ~84 ms per CCK↔OFDM table switch is the switch itself — disable tracking → rewrite the thermal table → re-enable, all USB control transfers — and it caps a mixed CCK/OFDM stream at ~11.5 fps on this backend. It predates this PR and I have not touched it here; happy to open a separate issue if that is worth pursuing.

snokvist:
## Witnessed on-air results — closing the gap from my last comment

A working RTL8812AU turned up, so the "submission rate, not airtime" caveat on my previous comment no longer applies. Recorded in `docs/rtl8733b.md` (`c516d5d`).

Witness: RTL8812AU `0bda:8812`, monitor mode, channel 6. DUT unchanged (bare RTL8731BU module, `0bda:f72b` cut D). 300 frames submitted per rate. The witness's hit counter reports the first ten hits then every hundredth, so these are floors at 100-frame granularity:

| requested | captured | FCS errors |
|---|---|---|
| 1 Mbps CCK | 300 / 300 | 0 |
| 6 Mbps OFDM | 300 / 300 | 0 |
| HT MCS7 | 300 / 300 | 0 |
| 11 Mbps CCK | >= 900 / 1000 | 0 |

11 Mbps first landed exactly on a counter boundary, so I re-ran it at 1,000 frames; both repetitions landed in the same bucket. The small deficit relative to the other three is repeatable rather than noise, and expected — it is the least robust CCK rate.

**Mixed run:** 250 frames alternating all eight rates, i.e. a CCK/OFDM TSSI table switch on *every* frame. The witness decoded all eight `rate_hw` values on air — `0/1/2/3` (1/2/5.5/11 Mbps CCK), `4` (6 Mbps OFDM), `12/16/19` (MCS0/4/7) — every one with valid FCS, `bw=0`, `stbc=0`, `ldpc=0`, matching the forced-BCC contract this backend advertises.

Two more commits since the review, both from things the profiling turned up:

- `2b88b1c` — `prepare_tssi_thermal` logged `offsets[0]`/`offsets[15]` as cold/hot edge. Those two words are `00000000` in **both** the CCK and OFDM plans (they differ only in words 4..11), so the line read `edge=00000000/00000000` on every churn log regardless of rate and could never show whether a swap took effect. Now reports the words that differ plus a fold over all 16. Same commit corrects the `0x881a` comment I added to the demo PID lists: that PID also ships on a Jaguar1 RTL8812AU-VS, observed here reporting chip-id `0x04`. Dispatch was always right — it reads SYS_CFG2, not the PID — only the comment was wrong.
- `c516d5d` — the results above.

Full matrix re-run on every commit: 53/53, RTL8733B-only 49/49, RTL8733B-only ASan+UBSan 49/49.

### Still not claimed

No SDR, so occupied bandwidth, spectral mask, EVM and absolute power remain unmeasured — the witness gives delivery and relative RSSI, not calibrated power. Narrowband TX, the `0bda:b733` combo module and true VBUS-cold boots are unchanged from the PR description.

### Two questions from my earlier replies still open for you

Whether the fatal TSSI path should throw rather than return `bool` (I kept `bool` for consistency with the other generations' `send_packet`, and latched the reason in `_tx_fatal`), and whether `SetCcaMode(true)` throwing non-destructively is acceptable. Happy to change either.

josephnef (commented):
## Follow-up review — all 13 comments addressed

Re-reviewed at `c516d5d`. **Every item from the previous round is resolved**, several beyond what I asked for. Verified independently rather than taken on faith:

- Fresh worktree at PR head + `git submodule update --init reference/rtl8733bu-20230626` → checked out exactly `9e5f6845b4e3393207720fc1886668bc80b6af1c`, then both extractors' `--check` **verified all 9 arrays byte-for-byte from that clone** — which is precisely the property that was missing before. Full build clean (no new warnings), **ctest 53/53**, `bash -n` clean.
- `git grep -n 'delta >= 25'` now returns only `Phy8733b.cpp`'s `kTssiAuditThermalCeiling`, correctly scoped to the `audit_tssi_*` probe stages with the comment explaining why a probe verdict is not a runtime gate. The device send path, `InitWrite`, the rxdemo poller and `rtl8733bprobe` are all clear.

### On the measurements you added

The per-frame thermal read at **2.51 ms of a 2.71 ms budget** is a better result than I expected — and thank you for labelling it *submission* rate with no witness plugged rather than letting 13.7× read as an on-air number. That distinction is exactly the house style. Same for the negative control on the soak script (`LISTEN_SECS=8 TIMEOUT_SECS=10` → rc=124 captured, gates ran, cleanup trap fired) — proving `set -euo pipefail` didn't convert a graded failure into a silent abort is the check that actually matters when adding `-e` to a script whose whole job is grading non-zero exits. And the note that the verdict line itself read HEALTHY there while the gate still failed on the exit code is the kind of detail that usually gets left out.

The `[,}]` anchoring turning out to be load-bearing (`"init":1}` at end-of-object) is a good catch on my behalf — I'd have suggested a comma and broken that gate.

Two things you volunteered that weren't asked for and are worth keeping: the **TSSI settling observation** (CCK airing ~5 dB above its settled level when alternating at ~9 ms/frame, converging at 86 ms/frame) paired with the favourable delivery table, and the **11 Mbps deficit** reported as repeatable-across-two-runs rather than smoothed away. Both are the adversarial counterpart the docs convention asks for. Filing the 84 ms table-switch cost as #389 instead of burying it in this PR is the right call.

### Fixes that went past the ask

- `kRxAggregateBytes8733b` with a `static_assert` tying the URB floor to the programmed `RXDMA_AGG` page count — I asked for a clamp; pinning the two constants to each other so they can't drift is better, and the `warn` on raise means a user who set `DEVOURER_RX_URB_BYTES=4096` learns why it moved.
- `_power_ready` armed **before** `power_on()` with the reasoning spelled out (`power_off()` is the full card-disable flow and is safe against a partially-enabled card) — that's the correct inversion, not just a rollback bolt-on.
- `tx_mode_supported_8733b` / `ht_request_supported_8733b` factored out so the `SetTxMode` and radiotap branches share one predicate. That converts "these two branches happen to agree today" into a structural property, and makes the STBC refusal testable on a descriptor encoder that has no STBC field to reject on.
- `_tx_fatal` latching a static reason string so a retry loop can distinguish "InitWrite hasn't run" from "the session was deliberately stopped" — cleaner than the throw I floated.
- The `881a` PID collision comment (Jaguar3 8812EU vs Jaguar1 8812AU-VS, chip-id dispatch makes it harmless) is a genuinely useful drive-by that has nothing to do with this backend.

### README

The split is exactly right, and separating **RTL8731BU** (1T1R, the validated row) from **RTL8733BU** (1T1R + BT, vendor-table-only, BT/coex not implemented) matches how RTL8812EU/RTL8822EU and RTL8852BU/RTL8832BU are handled. "bare unbranded 1T1R module" is an honest DUT description — it's the right answer when there's no brand to name. And `docs/rtl8733b.md` now stating that the second `f72b` unit *would have covered the same README row*, so the RTL8731BU row rests on one surviving sample — "a one-unit result, not a population" — is the sentence that makes the table readable a year from now.

### One remaining nit (non-blocking, your call)

`_cfg.tuning.disable_cca` is silently dropped on this backend. Every other generation applies the knob during bring-up (`if (_cfg.tuning.disable_cca) SetCcaMode(true);` — jaguar1:82/1349, jaguar2:251, jaguar3:102/868, kestrel:257), so `DEVOURER_DIS_CCA=1` is a no-op here while the direct setter throws. Given you deliberately made the setter refuse loudly rather than no-op, the config path is the one door left where the request vanishes without a word. Wiring it would surface as a hard throw at `InitWrite` — arguably correct but harsh for a knob CLAUDE.md describes as on-by-default for the streamtx FPV downlink; a one-line `warn` in bring-up when the knob is set would probably be the better trade. Either way it doesn't need to hold the PR.

**LGTM.** Nothing blocking from my side.

josephnef:
Deferred work from this PR is now tracked, all of it gated on the incoming **BL-M8733BU2** board so nothing here blocks the merge:

- **#390** — SDR RF characterization + on-air throughput benchmark (OBW, spectral mask, EVM, absolute power; fills the README bandwidth cells). Carries the TSSI-settling trap you measured: a mixed-rate stream at ~9 ms/frame reads ~5 dB high, so power measurements must be paced single-rate.
- **#391** — qualify **or remove** the experimental 5/10 MHz path. Depends on #390: the vendor's own workaround documents a no-RF-output failure mode, which register readback cannot distinguish from success. Three honest outcomes, deletion included.
- **#392** — BL-M8733BU2 / `0bda:b733`: identity confirmation, **second-sample** validation (today's numbers are a one-unit result), BT/coex scope, true VBUS cold boots, secure-firmware variants.
- **#393** — the missing vendor-driver A/B control. The 2024 tree failed to build on 6.14; the pinned-kernel VM (`tests/setup_vm.sh`) is the established answer, and the 2023 tree — the one the firmware and tables were actually extracted from — is the more faithful control.
- **#394** — the deliberately unadvertised surface: SGI root-cause (descriptor bit set, witness decoded long GI — the most interesting one), LDPC/STBC, ACK/A-MPDU, FastRetune, CCA disable, plus the `tuning.disable_cca` silent-drop asymmetry from my last review.

#233 stays open as the umbrella; its USB-first scope is delivered here and the remainder is the list above.

Merging. Thanks for the two review rounds — the follow-ups measured the things I could only assert, and the 93%-of-send-budget number turned a code-review opinion into a fact.

josephnef (approved):
Approving. Round-two verification: fresh worktree at PR head, submodule init checked out exactly 9e5f684, both extractors' `--check` verified all 9 arrays byte-for-byte from that clone, full build clean, ctest 53/53. All 13 review comments resolved, several beyond the ask. Deferred RF work tracked in #390-#394.

snokvist:
Thanks for the re-review, and for re-verifying the submodule property from a fresh clone rather than trusting my word for it — that was the one claim that couldn't be checked before.

## `disable_cca` — wired, as a warning (`412f149`)

You were right that the config path was the one door where the request vanished without a word, and right about the trade. Bring-up now emits:

```
devourer [W] RTL8733B: DEVOURER_DIS_CCA / tuning.disable_cca is not implemented
             by this backend — carrier-sense stays ENABLED for this session
```

and carries on. Verified on the DUT: warning fires, `InitWrite` completes, TX proceeds. A hard throw would have been the harsh answer you described, for a knob that is on by default on the streamtx downlink and that a caller may be setting only through an inherited environment. The comment at the site says why it is a warning and not the throw the setter raises, so nobody "fixes" it into a throw later.

## One more thing the same question turned up

Checking whether the TSSI path was opt-in — it wasn't — surfaced that this backend was the **only** one doing register I/O inside `send_packet`. All four existing HALs score zero on `sed -n '/::send_packet(/,/^}/p' | grep -c 'rtw_write|rtw_read|set_bb|write_rf'`; this one re-selected the CCK/OFDM thermal-compensation curve on every rate-class crossing, at the 84 ms from #389.

The vendor doesn't do that either: `_halrf_tssi_set_tmeter_tbl_8733b` is only ever called from full TSSI setup, keyed on `phydm_get_tx_rate` at that instant, and never re-selected at runtime. So the table is now chosen once per channel set and left alone, with per-frame switching behind `DEVOURER_TSSI_RATE_TABLE=1`. An alternating CCK/OFDM stream goes from 19.69 s to 2.82 s per 200 frames, 200 switches to 0.

**The first cut of this was wrong and the witness caught it.** I initially made TSSI itself opt-in, falling back to the flat `kSafeTxAgcIndex8733b`. On air that path cannot carry HT at all — MCS7, 300/300 submitted, **0 captured**, twice. On a TSSI-offset PG unit the loop *is* the power control, so it stays on; only the table switching is opt-in. With the loop on and the table fixed, MCS7 delivers 277/300 and 1 Mbps CCK 300/300, zero FCS errors.

Fixing the table costs nothing measurable at room temperature — the two curves differ only in the swing ramp away from baseline, so near zero thermal delta they agree. 1 Mbps CCK: 300/300 on the OFDM table, 299/300 on the CCK table.

## Refreshed delivery table

Now exact per-frame counts (`DEVOURER_RX_DUMP_ALL` filtered to the injected frames) rather than the throttled hit-counter floors I reported before, all under the shipped default:

| requested | delivered | decoded as | FCS errors |
|---|---|---|---|
| 1 Mbps CCK | 298 / 300 | 1 Mbps | 0 |
| 2 Mbps CCK | 300 / 300 | 2 Mbps | 0 |
| 5.5 Mbps CCK | 300 / 300 | 5.5 Mbps | 0 |
| 11 Mbps CCK | 294 / 300 | 11 Mbps | 0 |
| 6 Mbps OFDM | 300 / 300 | 6 Mbps | 0 |
| HT MCS0 | 270 / 300 | MCS0 | 0 |
| HT MCS7 | 277 / 300 | MCS7 | 0 |

Single probes, so ~±3 points each. The HT column sitting under the legacy column is not a finding on this evidence — MCS0 reading *below* MCS7 is the giveaway, since MCS0 is the more robust rate. I've said so in the doc rather than presenting the ordering as a modulation effect.

Note this supersedes my earlier "11 Mbps deficit is repeatable" claim in the direction of less certainty: with exact counts, 11 Mbps at 294/300 sits inside the same band as 1 Mbps at 298/300, so I'd no longer call it a distinguishable deficit.

53/53, 49/49 RTL8733B-only, 49/49 ASan+UBSan on every commit.

## Still not claimed

No SDR — occupied bandwidth, spectral mask, EVM and absolute power remain unmeasured; the witness gives delivery and relative RSSI only. Narrowband TX, the `0bda:b733` combo module and true VBUS-cold boots are unchanged.

snokvist:
Small precision note on the send-path claim in my last comment, since it is the kind of thing someone will re-check.

Running that same grep against the merged `Rtl8733bDevice.cpp` returns **4**, not 0:

```
16:  /* No thermal read here. read_thermal() is 3 RF writes + a 15 us settle ...
75:  if (_tx_submits.fetch_add(1) == 0) {
82:        _device.rtw_read8(0x0207), _device.rtw_read16(0x041a),
83:        _device.rtw_read32(0x0210), _device.rtw_read8(0x0522),
84:        _device.rtw_read16(0x2de0), _device.rtw_read16(0x2de2));
```

Line 16 is a comment. Lines 82-84 are six reads guarded by `_tx_submits.fetch_add(1) == 0` — a deliberate one-shot on the very first submission that dumps FIFO / TXDMA / BB-TX-enable state, which is what made the original TX bring-up debuggable.

So the accurate statement is: **steady-state per-frame register I/O is zero**, matching the other four generations, with six reads on the first frame of a session only. The grep counts text, not executions. I did not want the parity claim resting on a number that does not reproduce.

josephnef:
Both of these landed after the merge, so replying here — and one of them collides with something I merged 20 minutes later.

## Heads-up: `disable_cca` is already on master, and your wording is better

I hit the same gap from the docs side and shipped it in **64d77b5** (#395) before reading your comment. `InitWrite` there emits:

```
RTL8733B: CCA disable (DEVOURER_DIS_CCA) is not implemented by this backend;
transmitting with carrier-sense enabled
```

Your `412f149` is not pushed anywhere — master is `cccb863` + `64d77b5` — so you will get a conflict in exactly that hunk on your next rebase. **Keep yours and overwrite mine:** naming both spellings (`DEVOURER_DIS_CCA / tuning.disable_cca`) is the more useful line, because the env var is what an operator sets and the struct field is what a library caller sets, and a grep for either should find the warning. Same reasoning at the site, same trade, so it is a one-line resolution.

Note #395 also added `src/rtl8733b/CLAUDE.md` — the subtree file the other HALs have. It currently documents the 84 ms table switch as a per-frame cost, which your change invalidates; that paragraph needs revising in the same PR.

## The TSSI finding is the real result here

This is the better catch of the two comments, and the method is why: you checked what the *vendor* does at the call site rather than reasoning about what the registers permit. `_halrf_tssi_set_tmeter_tbl_8733b` being reachable only from full TSSI setup, keyed on `phydm_get_tx_rate` at that instant, is decisive in a way no amount of register inspection would have been — the table was never designed to be re-selected at runtime, so the 84 ms was paying for a mechanism that does not exist upstream. 19.69 s → 2.82 s per 200 frames, 200 switches to 0, effectively closes #389.

**The wrong first cut is the part worth writing down.** Making TSSI itself opt-in and falling back to the flat AGC index looked like the more conservative choice and was the more destructive one — MCS7 at 0/300, twice, on air. "On a TSSI-offset PG unit the loop *is* the power control" is a transferable fact about this PG mode, not an RTL8733B curiosity, and it is exactly the class of thing a register-readback test would have called green. The witness earned its keep.

One review question on the shape, not the result: **what is `DEVOURER_TSSI_RATE_TABLE=1` for now?** Your own evidence says the two curves differ only in the swing ramp away from baseline, so at room temperature fixing the table costs nothing measurable (300/300 vs 299/300 is not a difference). If that is the whole story the knob is dead weight and the vendor's behaviour is simply correct. But if the curves genuinely diverge at elevated thermal delta — which is the regime nobody has measured on this part — then the knob is an opt-in for a case that might matter, and it should say so at its declaration in those terms: *default is validated at room temperature; whether it is right hot is unmeasured*. Either "delete it" or "scope it honestly" is fine; what I would avoid is shipping it as a neutral performance toggle, since that reads as though someone has characterized both settings.

## The retraction is the right call

> with exact counts, 11 Mbps at 294/300 sits inside the same band as 1 Mbps at 298/300, so I'd no longer call it a distinguishable deficit

Applying the ±3 rule against your own earlier claim, unprompted, is worth more than the original claim was. Same for refusing to read the HT-under-legacy ordering as a modulation effect — MCS0 at 270 sitting *below* MCS7 at 277 is indeed the giveaway, because it inverts the robustness ordering, which means the spread is measuring something other than the transmitter.

Worth naming what that something probably is: on this evidence the honest suspect is the receiver, not the DUT. `tests/ground_station_qualify.sh` sweeps the witness's own ladder and refuses the pairing when the test rate sits off its flat part — running it on that RTL8812AU would tell you whether the ~270-277 HT band is the transmitter or the witness's margin. Cheap, and it converts "not a finding" into "attributed".

## Grep precision note: accepted, and the claim survives

Good instinct to pre-empt the re-check. For the record the merged CLAUDE.md wording is scoped the way your correction requires — *"nothing reads a register per frame on the send path"* — so the one-shot first-submission dump guarded by `_tx_submits.fetch_add(1) == 0` is not a counterexample to it. Keep that dump; six reads once per session that made the original bring-up debuggable is a good trade, and it is the steady-state number that has to match the other four HALs.

---

None of the above is pushed yet, so it all wants a PR — happy to review it when it is up.

Issue #390, opened 2026-08-13
RTL8733B: SDR RF characterization + on-air throughput benchmark (blocked on BL-M8733BU2 arrival)


josephnef:
The RTL8733B backend merged in #388 has **no RF-domain measurement at all** — the DUT bench had no SDR. Every RF claim in `docs/rtl8733b.md` is therefore a delivery/readback claim, and the README bandwidth cells for RTL8731BU are `—`. This issue closes that gap once a second module (BL-M8733BU2) is on the bench.

Everything here is hardware-gated: it cannot start before the board arrives.

## Why it matters beyond bookkeeping

- **`AdapterCaps::narrowband_ok` is false purely for want of an SDR** (#TBD-narrowband depends on this issue). The 5/10 MHz register sequence already reads back on hardware; occupied-bandwidth proof is the missing evidence.
- The README table is the first thing an integrator reads. A `—` row next to benchmarked 8812EU/8852BU rows reads as "untested silicon" when what is actually missing is one instrument.
- A one-unit result is not a population (`docs/rtl8733b.md`, "Known gaps"). The second module makes every number below a two-sample claim.

## Work items

1. **Absolute output power, per band and per rate.** 1/2/5.5/11 Mbps CCK and 6 Mbps OFDM on 2.4 GHz; 6 Mbps OFDM and HT MCS0/4/7 on both bands. The backend has no TX-power API surface beyond its safe TSSI target (`kSafeTssiTargetQdbm8733b`) — establish what that target actually radiates before anyone proposes raising it.
2. **Occupied bandwidth + spectral mask** at 20 and 40 MHz, both bands. This is the gate for the narrowband work and the sanity check on the 40 MHz centre-channel table.
3. **EVM** at the top of the ladder (MCS7/20, MCS7/40). Per `docs/bench-testing-near-field.md`, EVM — not SNR — is the saturation tell; take the near-field precautions.
4. **On-air TX throughput** via `tests/bench_onair.py`: **Mbps = SDR duty × PHY rate**, never monitor-sniffer frame counts (a sensitive receiver decodes weak frames and masks a real drop). Fill the README RTL8731BU row's 2.4 GHz / 5 GHz / 40 MHz cells.
5. **TSSI target validation.** Confirm the closed-loop target is met and stable across a bounded run, and that the safe cap is neither leaving margin on the table nor overdriving the PA.

## Traps this specific chip has already demonstrated

[...]

PR #396, opened 2026-08-13
rtl8733b: keep the send path free of register I/O, scope the TSSI rate-table knob

### LINKED ISSUE #389: rtl8733b: CCK<->OFDM TSSI table switch costs 84 ms (136 USB register ops) inside send_packet
## Summary

On the RTL8733B backend, switching the TSSI thermal table between the CCK and OFDM/HT variants costs **84 ms**. `select_tssi_rate_table` runs inside `send_packet`, so a stream that alternates CCK and OFDM rates is capped at ~11.5 fps.

This is latent, not a live regression. `select_tssi_rate_table` early-returns when the rate class is unchanged, so a single-rate stream — the normal FPV/video case — pays nothing. It predates #388 and is not caused by it.

## Where the 84 ms goes

Measured on the DUT (bare unbranded RTL8731BU module, `0bda:f72b` cut D, USB high speed) with an instrumented build counting every BB register access and timing each stage. 60 consecutive switches, steady state:

| stage | ms | reads | writes | ops |
|---|---|---|---|---|
| `enable_tssi_tracking` | 44.99 | 35 | 35 | 70 |
| `disable_tssi_tracking` | 25.77 | 13 | 34 | 47 |
| `prepare_tssi_thermal` | 13.50 | 2 | 17 | 19 |
| **total** | **84.26** | 50 | 86 | **136** |

There are no sleeps in the path — it is 136 USB control transfers. `build/reglat` on the same adapter measures a register round trip at mean 375 us / p99 407 us, and the mixed read/write traffic here averages ~620 us/op.

The structural cause: to change 16 BB dwords, the backend tears down and rebuilds the entire TSSI front-end. 32 of the 136 ops are the 16-command ANAPAR sequence (`write_tssi_anapar`), written out by `restore_tssi_analog` and written straight back by `apply_tssi_analog`. The channel and band do not change across a rate-table switch, so that analog state is restored only to be re-applied microseconds later.

## The premise looks wrong

`Rtl8733bDevice::select_tssi_rate_table` says the table "cannot be changed while closed-loop tracking is enabled". The vendor's own writer disagrees: `_halrf_tssi_set_tmeter_tbl_8733b` (`hal/phydm/halrf/rtl8733b/halrf_tssi_8733b.c` in the pinned `reference/rtl8733bu-20230626`) writes `R_0x4380` and the 16 dwords at `0x4200..0x423c` directly and touches no enable bit.

Caveat, stated plainly: the vendor only ever calls it from the full `_halrf_tssi_8733b` setup sequence, so there is **no vendor precedent for a runtime swap** in either direction. The vendor picks a table once at TSSI setup from whatever `phydm_get_tx_rate` returns and leaves it.

## The hardware accepts an in-place rewrite

Throwaway build, writing the table with tracking live and reading it back:

- **101/101 swaps** readback-verified across all 16 words against the rate-specific plan.
- `0x4318` stayed at `7` (tracking enabled) across every swap.
- **13.77 ms / 20 ops**, against 84.26 ms / 136 ops — a **6.1x** reduction.

The two plans genuinely differ (8 of the 16 words), so the readback was tracking a real change rather than a no-op.

## Why this is not already a PR

TSSI is closed-loop PA control. Register readback proves the write lands; it does not prove the tracking loop behaves correctly across a live table change. Validating that needs an independent witness and ideally an SDR — neither was available — and the failure mode of getting it wrong is transmitting at an uncalibrated power setting.

## Suggested work

1. Add a table-swap-only path to `select_tssi_rate_table` that writes `0x4380` + `0x4200..0x423c` in place and readback-verifies, skipping the disable/enable pair.
2. Gate it behind a `DeviceConfig` field until validated, defaulting to the current conservative path.
3. Validate with a witness at both rate classes across repeated swaps, and an SDR power check either side of a swap to confirm the closed loop settles to the same output power as the teardown path.
4. If it holds, make it the default and drop the comment claiming the table cannot be changed while tracking is enabled.

Separately: `write_tssi_anapar` is 16 dwords written twice per switch. If the in-place path lands, that cost disappears with it; if it does not, the restore/re-apply pair is worth revisiting on its own.

## Reproducing

```sh
# alternating CCK/OFDM so every frame forces a table switch
sudo DEVOURER_PID=0xf72b DEVOURER_CHANNEL=6 DEVOURER_TX_RATE=1M \
     DEVOURER_TX_MCS_SWEEP="1M,6M,2M,MCS0,5.5M,MCS4,11M,MCS7" \
     DEVOURER_TX_MCS_STEP_MS=17 DEVOURER_TX_GAP_US=2000 \
     DEVOURER_TX_FRAMES=500 ./build/txdemo
```

500 frames takes ~45 s (~87 ms/frame). The single-rate control — same command without `DEVOURER_TX_MCS_SWEEP` — takes ~3.5 s (~2.6 ms/frame).

Note that on `master` the per-switch `rtl8733b.tssi_thermal_table` log cannot show whether the swap took effect: it prints words 0 and 15, which are identical in both tables. #388 changes it to report the words that differ plus a fold over all 16.

snokvist:
The two post-merge findings, rebased onto master and answering the review points on both. Fixes #389.

## The send path no longer touches a register

This backend was the only one doing register I/O inside `send_packet` — it re-selected the CCK/OFDM thermal-compensation curve on every rate-class crossing at 84 ms / 136 USB round trips (#389). `configure_tx_power` now picks the curve once per channel set and leaves it, which is what the vendor does: `_halrf_tssi_set_tmeter_tbl_8733b` is reachable only from full TSSI setup, keyed on `phydm_get_tx_rate` at that instant, never re-selected at runtime.

Measured on the DUT — alternating CCK/OFDM, 200 frames: **19.69 s → 2.82 s, 200 table switches → 0**, TSSI still enabled. `send_packet` now has no conditional register-I/O path at all.

The TSSI loop itself stays on wherever the EFUSE is TSSI-offset PG, because there it *is* the TX-power control. An early cut made TSSI opt-in with a flat-index fallback; witnessed, that could not carry HT at all — **MCS7, 300/300 submitted, 0 captured, twice**.

## No runtime table switch, and the measurement that closes it

The first revisions carried `DEVOURER_TSSI_RATE_TABLE` as an opt-in for per-crossing switching, on the argument that the hot thermal-delta regime is where the two curves are *designed* to differ. Both reviewers landed on "probably delete it". A heat soak turned that from a judgement call into a measurement, so **the knob is deleted** — it could never have done anything on this part.

**The two curves are bit-identical until thermal delta +18.** Reconstructing the plan against the vendor 64-entry layout: the swing ramp starts at index 18 (CCK swing 0 for 0..17, then 3,3,5…; the OFDM plan is all-zero), and every word either side of the baseline is zero in both tables. Below +18 the CCK table and the OFDM table are the same bits.

**The part plateaus at +8.** Five-minute max-duty MCS7 soak (`DEVOURER_TX_GAP_US=0`, ~5000 fps submission), thermal polled every 2 s:

```
raw 33 -> 40   delta +1 -> +8
trajectory: 33/1, 37/5, 38/6, 39/7, 39/7, 39/7, 40/8, 40/8, 40/8, 40/8, 40/8
```

It stopped climbing after about two minutes and held +8 for the remaining three — asymptotic, not still-rising-when-I-stopped. Less than half the delta needed for a single table entry to differ. A runtime switch would therefore spend 84 ms and 136 USB register round trips, inside `send_packet`, to install a table bit-identical to the one already loaded.

Deleted with it: `select_tssi_rate_table`, the `DeviceConfig` field, the env translation, and `_tx_fatal` — whose only producer was that function's failure path, so keeping it would have left a permanently-null member guarding a dead branch. Re-verified on air after the deletion: MCS7 279/300, 1 Mbps CCK 300/300 — unchanged from 277/300 and 300/300 before it.

**+8 is a property of this unit, not of the silicon.** Bare module, open air, room ambient, and submission-rate duty rather than measured radiated duty. A sealed BL-M8733BU2 at high ambient could plausibly reach +18 where this one cannot. The **+18 threshold** is the durable half — it comes from the vendor table contents, not the unit. So the code and docs state the reasoning rather than the conclusion, and both point at #389 for the validated in-place implementation (13.8 ms, tracking left enabled) as the thing to revisit if a board ever does get hot enough; #390's heat-soak work now has a specific number to test against.

## `disable_cca`

Took the reviewer's call — kept my wording naming both spellings, dropped theirs. One placement change on top: it sits in `bring_up_to_phy` rather than `InitWrite`, so an RX-only session that set the knob is told too, and so it fires exactly once per bring-up rather than twice now that both existed. Verified on hardware: fires once, bring-up completes, TX proceeds.

## Docs

`src/rtl8733b/CLAUDE.md` from #395 described the 84 ms as an unconditional per-frame cost and stated the table cannot be changed while tracking is enabled — both now wrong. Rewritten, pointing at #389.

Follow-up review pass also removed the drift this PR would otherwise have introduced: both CLAUDE.md sites still placed the `disable_cca` warning at `InitWrite`, and `docs/rtl8733b.md` still narrated two runs in terms of per-frame TSSI table switching — a path this tree deletes, so those runs are not reproducible as described. The rate-coverage and thermal evidence stays; the mechanism framing is gone.

## The witness qualification you suggested

Cheap and worth it. Sweeping the RTL8812AU across the HT ladder, 300 frames each:

| rate | delivered | witness RSSI |
|---|---|---|
| MCS0 | 270 / 300 | 57.8 |
| MCS1 | 300 / 300 | 65.0 |
| MCS3 | 300 / 300 | 65.3 |
| MCS5 | 300 / 300 | 65.2 |
| MCS6 | 296 / 300 | 65.1 |
| MCS7 | 277 / 300 | 60.3 |

Flat at 100% through MCS5, so the witness is not measuring itself — a receiver at its cliff falls off progressively at the top rather than dipping at both ends. Both low readings also carried 5-7 units less RSSI than the flat middle, which places the variation in per-run link conditions.

## Checks

53/53, RTL8733B-only 49/49, RTL8733B-only ASan+UBSan 49/49. Hardware verified on the same one-unit DUT: send-path switch count, the heat soak, the single `disable_cca` warning, and the ladder above.

Still no SDR on this part, so nothing here asserts an RF-domain magnitude.

### DIFF
--- a/src/rtl8733b/Rtl8733bDevice.cpp
+++ b/src/rtl8733b/Rtl8733bDevice.cpp
@@ -73,6 +73,26 @@ void Rtl8733bDevice::bring_up_to_phy() {
   _phy_ready = true;
   _logger->info("RTL8733B factory path reached PHY-ready (cut {})",
                 _chip.cut);
+  /* Every other generation applies tuning.disable_cca during bring-up
+   * (jaguar1/2/3, kestrel all call SetCcaMode there), so a caller setting
+   * DEVOURER_DIS_CCA=1 reasonably expects it to take effect. This backend has
+   * not located and measured the HALMAC 87xx carrier-sense gate, so it cannot
+   * honour the request — say so once, loudly, rather than dropping it in
+   * silence, which is the one way a refused knob can look like a granted one:
+   * the operator would otherwise believe carrier-sense was off and read the
+   * resulting deferral as a transmitter problem.
+   *
+   * Deliberately a warning and not the throw SetCcaMode(true) raises: the knob
+   * is on by default for the streamtx FPV downlink, and failing bring-up
+   * outright is a harsh answer to a request the caller may be making only via
+   * an inherited environment. The session runs with standard carrier-sense.
+   *
+   * Sited in bring_up_to_phy rather than InitWrite so an RX-only session that
+   * set the knob is told too, and so it fires exactly once per bring-up. */
+  if (_cfg.tuning.disable_cca)
+    _logger->warn(
+        "RTL8733B: DEVOURER_DIS_CCA / tuning.disable_cca is not implemented by "
+        "this backend — carrier-sense stays ENABLED for this session");
 }
 
 void Rtl8733bDevice::Init(Action_ParsedRadioPacket packetProcessor,
@@ -111,17 +131,6 @@ void Rtl8733bDevice::InitWrite(SelectedChannel channel) {
     _tx_ready = true;
     _tx_submits = 0;
     _tx_fatal = nullptr;
-    /* Say so rather than dropping it. SetCcaMode(true) refuses loudly, so the
-     * config path must not be the one door where the same request vanishes
-     * without a word — the operator would otherwise believe carrier-sense was
-     * off and read the resulting deferral as a transmitter problem. Warn
-     * rather than throw: the knob is on by default for the streamtx FPV
-     * downlink, and refusing to bring TX up over an unported optimisation is a
-     * worse trade than airing with standard carrier-sense. */
-    if (_cfg.tuning.disable_cca)
-      _logger->warn("RTL8733B: CCA disable (DEVOURER_DIS_CCA) is not "
-                    "implemented by this backend; transmitting with "
-                    "carrier-sense enabled");
     /* One-shot thermal snapshot at bring-up — the PA-heating baseline for the
      * session, same as the Kestrel InitWrite snapshot. Logged, never acted on:
      * the meter is a PA-bias tracking index, not a calibrated °C sensor, and
@@ -143,23 +152,44 @@ void Rtl8733bDevice::InitWrite(SelectedChannel channel) {
 bool Rtl8733bDevice::configure_tx_power(SelectedChannel channel) {
   _tssi_tracking = false;
   _tssi_cck = false;
+  /* Closed-loop TSSI is the TX-power control on a TSSI-offset PG unit, so it
+   * is not optional there: the flat fallback index below is a conservative
+   * bring-up value, and on-air it runs cold enough that HT rates do not
+   * survive the link (measured on the DUT: MCS7 undecodable by an RTL8812AU
+   * witness, 300/300 submitted, 0 captured). A unit whose EFUSE carries no
+   * TSSI calibration has nothing to drive the loop and takes the flat path. */
   if (_efuse.tx_power_mode != rtl8733b::TxPowerPgMode8733b::TssiOffset)
     return _phy.set_flat_tx_power(rtl8733b::kSafeTxAgcIndex8733b);
 
-  /* Start in the OFDM/HT thermal table. A first CCK submission switches the
-   * table through select_tssi_rate_table() before its descriptor reaches USB. */
+  /* Pick the thermal-compensation curve once, from the TX mode configured at
+   * this point, and leave it alone — the vendor's own setup keys the table on
+   * `phydm_get_tx_rate` at TSSI-setup time and never re-selects it at runtime.
+   * Re-selecting per frame is the opt-in DEVOURER_TSSI_RATE_TABLE path; the
+   * default keeps send_packet free of register I/O like every other
+   * generation. */
+  const bool cck_table = _tx_mode_default.has_value() &&
+                         _tx_mode_default->mode ==
+                             devourer::TxMode::Mode::Legacy &&
+                         rtl8733b::is_cck_rate_500kbps(
+                             _tx_mode_default->legacy_rate_500kbps);
   if (!_phy.prepare_tssi_bb(channel, _efuse) ||
-      !_phy.prepare_tssi_thermal(_efuse, false) ||
+      !_phy.prepare_tssi_thermal(_efuse, cck_table) ||
       !_phy.prepare_tssi_offsets(channel, _efuse) ||
       !_phy.enable_tssi_tracking(
           channel, _efuse, rtl8733b::kSafeTssiTargetQdbm8733b))
     return false;
   _tssi_tracking = true;
+  _tssi_cck = cck_table;
   return true;
 }
 
 bool Rtl8733bDevice::select_tssi_rate_table(bool cck) {
-  if (_efuse.tx_power_mode != rtl8733b::TxPowerPgMode8733b::TssiOffset)
+  /* Default: the table was chosen once in configure_tx_power and stays put, so
+   * the send path does no register I/O — the vendor behaves the same way.
+   * Both conditions must be checked: gating only on the EFUSE PG mode would
+   * return false here, which send_packet treats as fatal and stops the card. */
+  if (!_cfg.tuning.tssi_rate_table ||
+      _efuse.tx_power_mode != rtl8733b::TxPowerPgMode8733b::TssiOffset)
     return true;
   if (!_tssi_tracking)
     return false;
--- a/src/rtl8733b/Rtl8733bDevice.cpp
+++ b/src/rtl8733b/Rtl8733bDevice.cpp
@@ -131,6 +131,16 @@ void Rtl8733bDevice::InitWrite(SelectedChannel channel) {
     _tx_ready = true;
     _tx_submits = 0;
     _tx_fatal = nullptr;
+    /* An opt-in that re-introduces register I/O on the send path should not be
+     * silent about it: enabling this trades ~84 ms per CCK<->OFDM crossing for
+     * thermal-compensation accuracy nobody has yet shown to matter, and a
+     * caller who inherited the knob from an environment deserves to see the
+     * cost rather than discover it as a mysterious frame-rate ceiling. */
+    if (_cfg.tuning.tssi_rate_table && _tssi_tracking)
+      _logger->warn(
+          "RTL8733B: DEVOURER_TSSI_RATE_TABLE is on — each CCK<->OFDM rate "
+          "crossing re-selects the TSSI thermal table inside send_packet at "
+          "~84 ms / 136 USB register round trips (OpenIPC/devourer#389)");
     /* One-shot thermal snapshot at bring-up — the PA-heating baseline for the
      * session, same as the Kestrel InitWrite snapshot. Logged, never acted on:
      * the meter is a PA-bias tracking index, not a calibrated °C sensor, and
--- a/src/rtl8733b/Rtl8733bDevice.cpp
+++ b/src/rtl8733b/Rtl8733bDevice.cpp
@@ -150,7 +150,6 @@ void Rtl8733bDevice::InitWrite(SelectedChannel channel) {
 
 bool Rtl8733bDevice::configure_tx_power(SelectedChannel channel) {
   _tssi_tracking = false;
-  _tssi_cck = false;
   /* Closed-loop TSSI is the TX-power control on a TSSI-offset PG unit, so it
    * is not optional there: the flat fallback index below is a conservative
    * bring-up value, and on-air it runs cold enough that HT rates do not
@@ -171,7 +170,16 @@ bool Rtl8733bDevice::configure_tx_power(SelectedChannel channel) {
    * 0..+17 and first differ at +18, while a five-minute max-duty MCS7 soak
    * plateaued at +8 after two minutes and stopped climbing. It is set from the
    * rate class anyway because that costs nothing and is what the vendor does —
-   * but do not expect it to be measurable. */
+   * but do not expect it to be measurable.
+   *
+   * Which is just as well, because this is weaker than the vendor's keying in
+   * the canonical demo flow: txdemo calls InitWrite before SetTxMode, so
+   * _tx_mode_default is usually unset here and the OFDM/HT table is what a
+   * DEVOURER_TX_RATE=1M session loads — the CCK branch is reached only by a
+   * caller that configures the mode first, or by a later SetMonitorChannel.
+   * The vendor keys on the rate in flight at setup instead. Nothing chases
+   * that gap, because closing it would buy a table identical to the one
+   * already loaded at every delta this part reaches. */
   const bool cck_table = _tx_mode_default.has_value() &&
                          _tx_mode_default->mode ==
                              devourer::TxMode::Mode::Legacy &&
@@ -184,7 +192,6 @@ bool Rtl8733bDevice::configure_tx_power(SelectedChannel channel) {
           channel, _efuse, rtl8733b::kSafeTssiTargetQdbm8733b))
     return false;
   _tssi_tracking = true;
-  _tssi_cck = cck_table;
   return true;
 }
 
@@ -591,7 +598,6 @@ void Rtl8733bDevice::Stop() {
     }
     _tssi_tracking = false;
   }
-  _tssi_cck = false;
   _phy_ready = false;
   if (_mac_ready) {
     _mac.stop();
--- a/src/rtl8733b/Rtl8733bDevice.cpp
+++ b/src/rtl8733b/Rtl8733bDevice.cpp
@@ -130,17 +130,6 @@ void Rtl8733bDevice::InitWrite(SelectedChannel channel) {
       throw std::runtime_error("RTL8733B monitor RX configuration failed");
     _tx_ready = true;
     _tx_submits = 0;
-    _tx_fatal = nullptr;
-    /* An opt-in that re-introduces register I/O on the send path should not be
-     * silent about it: enabling this trades ~84 ms per CCK<->OFDM crossing for
-     * thermal-compensation accuracy nobody has yet shown to matter, and a
-     * caller who inherited the knob from an environment deserves to see the
-     * cost rather than discover it as a mysterious frame-rate ceiling. */
-    if (_cfg.tuning.tssi_rate_table && _tssi_tracking)
-      _logger->warn(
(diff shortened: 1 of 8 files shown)


snokvist:
Both bot findings addressed in `5a009ee`, one by fixing it and one by disagreeing with the framing but fixing the real problem underneath.

**1. CLAUDE.md duplicating the knob docs — valid, fixed.** The root CLAUDE.md rule is explicit ("never duplicate what a header already doc-comments") and I had restated the room-temperature validation, the unmeasured hot regime and the not-a-neutral-toggle framing in the subtree file. The subtree file now keeps only what is genuinely subtree knowledge — the table is chosen once, the vendor does the same, and what the transition failure mode is and is not — and points at `src/DeviceConfig.h` for when the knob is worth paying for.

**2. `tssi_rate_table` causes send I/O — by design, but it should not have been silent.** Enabling the knob re-introducing register I/O on the send path is its documented purpose, so I do not think the invariant is violated: "nothing reads a register per frame on the send path" is a statement about the default configuration, the same way `DEVOURER_DIS_CCA` does not violate "carrier-sense is enabled by default". Deleting an opt-in because a rule about defaults reads unconditionally would be the wrong fix.

What the finding did surface is a real defect I had missed: the knob was **silent**. Someone who inherits `DEVOURER_TSSI_RATE_TABLE=1` from an environment would meet the ~84 ms per rate crossing as an unexplained frame-rate ceiling, with nothing in the log connecting the two. `InitWrite` now warns once when the knob is active and TSSI is running, naming the cost and the issue — the same reasoning as the `tuning.disable_cca` warning one function above. Verified: warns once with the knob on, silent with it off.

@josephnef the standing offer from the PR description is unchanged and this does not pre-empt it — if you would rather not carry the knob at all, say so and I will drop it, and the capability should then come back (if #390 shows the hot case diverges) implemented as the validated in-place swap at 13.8 ms rather than the 84 ms teardown/rebuild it uses today. Deleting is a one-commit change; I have kept it only because #390 is the experiment that would justify or bury it.

53/53, RTL8733B-only 49/49, ASan+UBSan 49/49 on both commits.

snokvist:
## Ran the heat soak. The knob is deleted — it could never have done anything.

`ba15a5d`. You both landed on "probably delete it"; the soak turns that from a judgement call into a measurement.

**The two curves are bit-identical until thermal delta +18.** Reconstructing the plan against the vendor 64-entry layout: the swing ramp starts at index 18, and every word either side of the baseline is zero in *both* tables. Below +18 the CCK table and the OFDM table are the same bits.

**The part plateaus at +8.** Five-minute max-duty MCS7 soak (`DEVOURER_TX_GAP_US=0`, ~5000 fps submission), thermal polled every 2 s:

```
raw 33 -> 40   delta +1 -> +8
trajectory: 33/1, 37/5, 38/6, 39/7, 39/7, 39/7, 40/8, 40/8, 40/8, 40/8, 40/8
```

It stopped climbing after about two minutes and held +8 for the remaining three — asymptotic, not still-rising-when-I-stopped. That is **less than half** the delta needed for a single table entry to differ.

So a runtime switch would spend 84 ms and 136 USB register round trips, inside `send_packet`, to install a table bit-identical to the one already loaded. There is no temperature this unit reaches at which the knob does anything. It is deleted, along with `select_tssi_rate_table`, the `DeviceConfig` field, the env translation, and `_tx_fatal` — whose only producer was that function's failure path, so keeping it would have left a permanently-null member guarding a dead branch. `send_packet` now has no conditional register-I/O path at all.

Re-verified on air after the deletion: MCS7 279/300, 1 Mbps CCK 300/300 — unchanged from 277/300 and 300/300 before it.

### What I deliberately did not claim

**+8 is a property of this unit, not of the silicon.** Bare module, open air, room ambient, and submission-rate duty rather than measured radiated duty. A sealed BL-M8733BU2 at high ambient could plausibly reach +18 where this one cannot. The **+18 threshold** is the durable half — it comes from the vendor table contents, not the unit.

So the code and docs now state the reasoning rather than just the conclusion, and both point at #389 for the validated in-place implementation (13.8 ms, tracking left enabled) as the thing to revisit if a board ever does get hot enough. I have added the same to #390 so the heat-soak work there has a specific number to test against.

53/53, RTL8733B-only 49/49, ASan+UBSan 49/49. The build warnings in the log are pre-existing (`hal/basic_types.h`, svctx, hopset, chanmig), not from this branch.

josephnef (commented):
Reviewed at ba15a5d3. The code change is good — deleting the knob on the table-identity evidence is the right call, and I verified the +18 claim against `Phy8733b::tssi_thermal_plan` (CCK swing 0 for 0..17, then 3,3,5…; OFDM plan all-zero — both sides of the baseline). Built clean, ctest 53/53 here too. Findings are all doc/consistency, two of them inline below, three here:

1. **The PR description is stale — it argues for keeping the knob the head commit deletes.** The body still says "I kept it rather than deleting it… say so and I will drop it in this PR", but ba15a5d3 already dropped it. Please update the description before merge so the merge record matches the change — and the bit-identical-through-+17 evidence deserves to be in the body, it's the strongest part of the PR.

2. **Two CLAUDE.md sites still say the `disable_cca` warning fires "at `InitWrite`".** The PR moves it to `bring_up_to_phy`, but root `CLAUDE.md` ("setting the config knob warns at `InitWrite`", ~line 337) and `src/rtl8733b/CLAUDE.md` ("`InitWrite` warns rather than dropping it", ~line 141) weren't updated. Given this PR exists partly to fix #395's doc drift, it shouldn't introduce the same shape of drift.

3. **`docs/rtl8733b.md` still narrates measurements of the removed mechanism as current behaviour**: "submitted 250 frames through a CCK/OFDM TSSI table switch on every frame" (~line 95) and "1,000/1,000 frames through 118 CCK/OFDM TSSI table changes" (~line 125) describe a code path this PR deletes — those runs cannot be reproduced on this tree. The delivery/thermal evidence is worth keeping; the table-switch framing should go (docs are current-state only in this repo).

Nothing here touches the hardware-validated behaviour. Happy to approve once the doc drift and the PR body are fixed.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -143,46 +151,40 @@ void Rtl8733bDevice::InitWrite(SelectedChannel channel) {
...
+   * It is never re-selected per frame, so send_packet does no register I/O,
+   * like every other generation.
+   *
+   * The curve choice is inert on this silicon at any temperature the part
+   * reaches: the CCK and OFDM/HT tables are bit-identical for thermal deltas
+   * 0..+17 and first differ at +18, while a five-minute max-duty MCS7 soak
+   * plateaued at +8 after two minutes and stopped climbing. It is set from the
+   * rate class anyway because that costs nothing and is what the vendor does —
+   * but do not expect it to be measurable. */
+  const bool cck_table = _tx_mode_default.has_value() &&
```
josephnef, line 183: The comment says the curve "is set from the rate class anyway … which is what the vendor does", but in the canonical flow no mode is configured yet: txdemo calls `InitWrite` (`examples/tx/main.cpp:998`) *before* `SetTxMode` (`:1058`), so a `DEVOURER_TX_RATE=1M` session loads the OFDM table here, and the CCK branch is only reachable via a later `SetMonitorChannel`. Harmless by this PR's own evidence (inert below +18), but the vendor keys on the *actual* rate at setup while this usually reads an unset optional — one sentence acknowledging the ordering would keep the comment honest. (Alternatively move txdemo's `SetTxMode` above `InitWrite`; it's just a stored optional and valid pre-init.)

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -143,46 +151,40 @@ void Rtl8733bDevice::InitWrite(SelectedChannel channel) {
...
-  _tssi_tracking = false;
-  if (!_phy.prepare_tssi_thermal(_efuse, cck) ||
-      !_phy.enable_tssi_tracking(
-          _channel, _efuse, rtl8733b::kSafeTssiTargetQdbm8733b))
-    return false;
-  _tssi_tracking = true;
-  _tssi_cck = cck;
-  _logger->info("RTL8733B TSSI rate table selected: {}",
-                cck ? "CCK" : "OFDM/HT");
+  _tssi_cck = cck_table;
```
josephnef: With `select_tssi_rate_table` gone, `_tssi_cck` is now write-only — set here and cleared in `Stop()`, never read anywhere. Delete the member, or say what future reader it's being kept for.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Fixed in `70e95fc` — took the comment route rather than the txdemo reorder.

The ordering is now stated where the table is picked: `_tx_mode_default` is usually unset here because txdemo calls `InitWrite` before `SetTxMode`, so a `DEVOURER_TX_RATE=1M` session loads the OFDM/HT table, the CCK branch is reachable only by a caller that configures the mode first or by a later `SetMonitorChannel`, and the vendor keys on the rate in flight instead. The comment says the gap is deliberately not chased, and why — closing it buys a table identical to the one already loaded at every delta this part reaches.

I left txdemo alone because that reorder touches a demo shared by all five backends to close a gap that is provably inert on this one, and `SetTxMode` pre-`InitWrite` is only obviously a stored optional *here* — I have not checked what the Jaguar/Kestrel implementations do with a mode set before bring-up. Happy to make the move if you would rather have the flow correct by construction; it just wants that check first.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Right — deleted in `70e95fc`. It was set in `configure_tx_power`, cleared in `Stop()`, and read nowhere once `select_tssi_rate_table` went; there is no future reader worth carrying it for, and a stale flag describing a table nothing consults is exactly the thing that misleads the next person who greps it. Same reasoning as `_tx_fatal` in `ba15a5d`.

snokvist:
All five findings addressed in `70e95fc`. No behavioural change in the commit — a dead member, two comments and three doc sites — so the hardware numbers from `ba15a5d` stand unchanged.

**1. PR description.** Rewritten. It no longer argues for keeping the knob it deletes, and the two pieces of evidence you called the strongest are now in the body rather than buried in a thread: the curves are bit-identical for deltas 0..+17 and first differ at +18 (CCK swing 0 through 17, then 3,3,5…; OFDM plan all-zero), and the soak trajectory that plateaus at +8. The "+8 is this unit, +18 is the silicon" split is stated there too.

**2. `disable_cca` at `InitWrite`.** Both sites fixed — root `CLAUDE.md` now says the knob warns once at bring-up including RX-only sessions, and `src/rtl8733b/CLAUDE.md` names `bring_up_to_phy` and why it sits there. You are right that this PR of all PRs should not have shipped that.

**3. `docs/rtl8733b.md` narrating the removed mechanism.** Both passages rewritten. The 250-frame run now reads as crossing the CCK/OFDM rate-class boundary per frame (which is what the run actually established — eight `rate_hw` values on air, valid FCS, 20 MHz, STBC/LDPC clear), and the 1,000-frame run as a mixed-rate run holding 32–33 raw against baseline 32. Both now carry an explicit line that they are rate-coverage and thermal evidence only and say nothing about the TX-power path, since at +0..+1 the two curves are the same bits either way.

**4 and 5** answered inline: `_tssi_cck` deleted, and the ordering caveat written into the comment rather than reordering txdemo — with the reason, and the offer to do the reorder instead if you want the flow correct by construction.

53/53, RTL8733B-only ASan+UBSan 49/49.

josephnef (approved):
All five findings addressed in `70e95fc` — verified against the diff, not just the replies. The comment at the table-selection site now states the InitWrite-before-SetTxMode ordering and why the gap is deliberately not chased; both CLAUDE.md sites track the warning's real location; docs/rtl8733b.md re-narrates the two runs without the deleted mechanism and pairs the evidence with its scope; `_tssi_cck` is gone. The commit is no-behavioural-change as claimed (only the two dead-member writes leave the .cpp), and the witness-ladder qualification closes the last open question about the delivery numbers. Approving.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -143,23 +152,44 @@ void Rtl8733bDevice::InitWrite(SelectedChannel channel) {
...
+  /* Pick the thermal-compensation curve once, from the TX mode configured at
+   * this point, and leave it alone — the vendor's own setup keys the table on
+   * `phydm_get_tx_rate` at TSSI-setup time and never re-selects it at runtime.
+   * Re-selecting per frame is the opt-in DEVOURER_TSSI_RATE_TABLE path; the
+   * default keeps send_packet free of register I/O like every other
+   * generation. */
+  const bool cck_table = _tx_mode_default.has_value() &&
+                         _tx_mode_default->mode ==
+                             devourer::TxMode::Mode::Legacy &&
+                         rtl8733b::is_cck_rate_500kbps(
```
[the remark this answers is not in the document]
josephnef: Addressed in 70e95fc, deliberately by documentation rather than by reprogramming on SetTxMode: the comment at the selection site now states the ordering gap and why it is not chased — the two thermal-compensation curves are bit-identical for deltas 0..+17 and this part plateaus at +8 under a five-minute max-duty soak, so reselecting on SetTxMode would install a table identical to the one already loaded at every delta the part reaches. Wiring the reprogram into SetTxMode would reintroduce register I/O on a runtime setter for zero measurable effect. Resolving.

PR #399, opened 2026-08-16
rtl8733b: runtime TX-power offset on the TSSI target

snokvist:
Closes the runtime TX-power gap on the RTL8733B. Follow-up to the FastRetune
port (#398); the issue is snokvist/devourer#1, which deferred these knobs out
of #388's scope deliberately.

## Why now

`Rtl8733bDevice` overrode none of the `IRtlDevice` power family, so every
`SetTxPowerOffsetQdb` landed on `IRtlDevice.h:134` `(void)qdb; return 0;`.
Measured on a CV610 bench 2026-08-14: **18 dB of commanded offset moved the air
by nothing**, EVM bit-identical across every rung, while a matched 8822EU
control on the same receiver moved 5.6 dB for a 6 dB command and then railed.

That the knob was unported is documented. The defect is that `return 0` is
indistinguishable from a successful zero-offset apply — the consumer's state
read `{"applied_qdb":0,"saturated_low":false}`, which is exactly what a healthy
actuator with travel remaining looks like. Reproduced at register level here:
master writes the same all-zero TSSI target dwords whether asked for 0 or −48.

## What it does

On a TSSI-offset PG unit the closed loop **is** the TX-power control, so the
actuator is its per-rate target table: the five packed dwords at
`0x3a00..0x3a10`, rewritten **in place with tracking left enabled**. That is the
shape #389 validated and `fast_retune` already uses for its per-channel rewrite
— not the ~165 ms disable/re-enable pair. Everything that can refuse is computed
before the first chip write; a failed readback rolls back to the offsets the
chip was carrying.

The offset caps first and shifts second, so the calibrated per-rate spread
survives the move; a lowered ceiling would move only the rates above it. It is
sticky by construction — `configure_tx_power` re-folds it on every channel set
and `FastRetune` passes it into the hop rewrite, without which a hop would
recompute from the bare ceiling and silently walk the operator's backoff back
up.

Caps report the dBm-target model Kestrel already uses (`index_max = 0`, one qdB
per step). Only the relative knob is ported: `SetTxPowerIndexOverride` is
overridden **solely to log a refusal** (the base returns `void`, so silence
would otherwise be the caller's only answer), and `SetTxPowerRateDiffs` /
`ReApplyTxPower` keep their `false` defaults. `kSafeTxAgcIndex8733b` was
witnessed unable to carry HT at all — MCS7, 300/300 submitted, 0 captured,
twice.

## Range: the int8 delta field at both ends, `[-128, +127]`

The `0x3a00` bytes are a signed int8 offset from the 64 qdBm anchor, so the
field spans targets from -16 dBm to +47.75 dBm. Neither end is re-clamped at
something softer, the same clamped-only-at-the-hardware-rail answer Jaguar1
(±126) and Jaguar3 (±127) give. `src/TxPower.h` is deliberate that headroom
above the generated table belongs to the operator, and a per-unit EFUSE trimmed
too cold is precisely what a bench calibration exists to correct.

Both knees are **measured and documented rather than enforced**, because on this
part both sit well inside the field and at points only a bench can find:

- **~-96 qdB** — where the loop stops reducing power. Below it three successive
  rungs read 52.96 / 52.97 / 52.99.
- **~+32 qdB** — where the PA compresses: 8.7 dB louder with EVM collapsed to
  -18. See the table below.

An earlier cut of this PR clamped the floor at -64 (a 0 dBm target) on the
assumption that a negative absolute target was meaningless. That was an
assumption rather than a measurement, in a change whose subject is knobs that
assume instead of checking, and it cost **~9 dB of working backoff** — power
keeps falling 7.3 dB past the 0 dBm target with no sign wrap. Corrected in
`03daab5`.

## Measured on air

RTL8731BU DUT → RTL8812AU witness (chip-RSSI ground station — the SDR saturates
at this range, per `tests/txpwr_offset_onair.sh`). Four independent 6-point
passes, one of them on a cold re-enumeration:

| pass | slope (dB/qdB) | span | rms |
|---|---|---|---|
| warm 1 | 0.222 | 14.2 dB | 1.02 |
| warm 2 | 0.231 | 14.8 dB | 0.64 |
| warm 3 (harness) | 0.227 | 14.5 dB | 1.02 |
| **cold replug** | **0.253** | **16.2 dB** | **0.49** |

**Do-nothing control, same session and geometry:** the pre-change binary
measured **flat — 0.3 dB across the same 64 qdB** — because it had no actuator.
Without that control the 14 dB would not be readable.

`step_measured` stays **false**, now for a measured reason rather than an
unexamined one: the step is not constant. The bottom 12 qdB delivered 0.125 and
0.126 dB/qdB — the one structure that reproduced exactly — while everything
above −52 ran 0.233..0.242. Same call the 8822E gets, for the same reason.

### Overdrive buys ~3 dB and then the PA compresses — EVM is the only tell

Sweeping UP from the clip (MCS0, ch36, witness reporting EVM beside RSSI):

| offset | RSSI | EVM | SNR |
|---|---|---|---|
| 0 | 75.4 | −62.0 | 62.0 |
| +16 (top of the PG table) | 78.2 | −50.2 | 63.7 |
| +32 | 84.1 | **−18.0** | 57.9 |
| +48 | 83.7 | −18.0 | 58.1 |
| +64 | 83.8 | −18.0 | 58.2 |

+16 is real gain that already cost 12 dB of EVM. +32 is **not** gain: 8.7 dB
more energy with the constellation collapsed, and +48/+64 move nothing at all.
**SNR held 58..64 throughout and never saw it** — the failure
`docs/bench-testing-near-field.md` exists to warn about. So the vendor's PG
table lands about where this part stops being linear: +16 qdB is the edge of
*usable* overdrive even though the API allows +127. Below the clip EVM is flat
at −58..−61 across all 16 dB of backoff.

## Verification

- `ctest` **54/54**. Per-chip configs: 8733B-only **50/50**, 8733B-off
  **45/45** (the new tests correctly absent). ASan+UBSan selftest passes.
  Clean full build adds **zero** warnings — the 8 present are pre-existing in
  vendored headers, `chanmig`, `hopset` and `Radiotap.c`.
- Cross-builds for armv7 musl (the CV610 craft that runs this adapter).
- `tests/rtl8733b_txpwr_regcheck.sh` — **7/7 on the device, warm and cold**:
  caps, offset-0 parity vs a master build, the null control (master ignores a
  −48 request), the −24 shift, both rails, stickiness across
  `SetMonitorChannel` + `FastRetune`, and `hw_readback` after the hop.
- `tests/rtl8733b_txpwr_selftest.cpp` in `ctest` — offset math, shape
  preservation, rails, and `kMaxPgTargetQdbm8733b` pinned against the generated
  table. Both mutations (drop the shift; implement it as a lowered ceiling)
  fail it.
- CI note: GitHub Actions is billing-blocked account-wide on the fork, so the
  build matrix above was run locally instead. A red X there is not a content
  failure.

## Notes for review

- **`GetTxPowerCaps()` is static and EFUSE-free** even though the actuator
  needs TSSI-offset PG mode. `IRtlDevice.h:194` contracts `GetAdapterCaps` as
  resolved at construction and callable before `Init`, EFUSE is not read until
  bring-up, and consumers cache it — so mutating caps mid-session would be
  worse. The refusal surfaces on the setter (returns 0, logs) and in
  `GetTxPowerState` instead. An offset latched before bring-up on such a unit
  is dropped loudly and zeroed by `configure_tx_power`, so reported state never
  claims an offset no register carries. That path has no hardware coverage —
  every unit seen is TSSI-offset PG — but it writes no registers.
- **A positive session offset lifts bring-up targets above the safe clip** by
  design. That is a deliberate widening of what bring-up may program, not a
  hole in `enable_tssi_tracking`'s `max_target_qdbm > 64` guard, which still
  refuses a caller that tries to raise the ceiling argument itself.
- **`tests/txpwr_offset_onair.sh` gained a `knob` column.** Every existing row
  passes `DEVOURER_TX_PWR`, exactly what the loop hardcoded before, so the
  Jaguar/Kestrel rows are unchanged. The 8733B row sweeps
  `DEVOURER_TX_PWR_OFFSET_QDB` because `DEVOURER_TX_PWR` drives the unported
  flat-index knob — verified a no-op on the device (0x10 vs 0x3f produce
  identical registers), so sweeping it would draw a flat line inside the test
  meant to catch flat lines. Its cells stop at 0: the overdrive half is not
  monotone in received power and cannot be scored by an RSSI slope.
- **A separate `rtl8733b_txpwr_regcheck.sh`** rather than a row in
  `txpwr_offset_regcheck.sh`, because every cell there reads TXAGC indices this
  dBm-model family reports as −1. Kestrel is absent from that script for the
  same reason and has its own.
- One physical unit, one witness, near-field geometry, integer-quantised RSSI,
  no SDR — as with every other on-air claim in this backend.

### DIFF
--- a/docs/rtl8733b.md
+++ b/docs/rtl8733b.md
@@ -137,6 +137,87 @@ not survive the link. Measured with the witness, 300 frames submitted at MCS7
 with the flat index: zero captured, twice. A unit whose EFUSE carries no TSSI
 calibration has nothing to drive the loop and takes the flat path.
 
+### The runtime lever
+
+Because the loop is the power control, the runtime knob is the loop's target.
+`IRtlDevice::SetTxPowerOffsetQdb` shifts every per-rate target that survives
+the safe ceiling — cap first, then shift, so the calibrated per-rate spread
+comes through the move intact — and writes it as the five packed dwords at
+`0x3a00..0x3a10`, **in place, with tracking left enabled**. That is the same
+shape `fast_retune` uses for its per-channel rewrite (#389); the alternative,
+a TSSI disable/re-enable pair, costs ~165 ms and buys nothing here.
+
+The capability report is the dBm model — `index_max = 0`, one qdB per step —
+over `[-64, +127]`, the int8 per-rate delta field's range and the same
+clamped-only-at-the-hardware-rail answer Jaguar1 (±126) and Jaguar3 (±127)
+give. Offset 0 is `kSafeTssiTargetQdbm8733b` (16 dBm), a first-light clip
+sitting at or below this part's factory targets (18–20 dBm at 2.4 GHz,
+16–19 at 5 GHz); −64 qdB puts the target at 0 qdBm, where the delta field
+bottoms out at the same moment.
+
+The positive half is deliberately **not** re-clamped at the factory table.
+`src/TxPower.h` states for every family that headroom above the generated
+table is the operator's call, and a per-unit EFUSE trimmed too cold is exactly
+what a bench calibration exists to correct — clamping here would make this the
+one backend where a measured operating point cannot be commanded. What that
+costs is measured below.
+
+The offset is sticky: a full `SetMonitorChannel` re-folds it against the new
+channel's targets, and `FastRetune` carries it into the hop's in-place rewrite.
+
+The flat-index and per-rate-diff knobs are **not** ported. The flat index has
+no measured dB-per-step slope on this part and, as above, cannot carry HT.
+
+**Measured on air.** Two independent 6-point passes at ch36 against an
+RTL8812AU witness (chip-RSSI ground station — the SDR saturates at this range,
+see `tests/txpwr_offset_onair.sh`): the lever is monotone and worth
+**14.2 / 14.8 dB** of received power for the full 16 dB of command, overall
+slope **0.222 / 0.231 dB per qdB** against the 0.25 nominal. Its adversarial
+counterpart, in the same breath: the step is not constant. The bottom 12 qdB
+delivered **0.125 and 0.126** dB/qdB — the only structure that reproduced
+exactly across both passes — while everything above −52 qdB ran 0.233..0.242
+and the mid-range scattered 0.219..0.252 between passes. Near the floor the
+loop returns about half the dB commanded, and `saturated_low` does not warn
+about it (that clamp fires only at −64). `TxPowerCaps::step_measured`
+therefore stays **false**, the same call the 8822E gets for the same reason: a
+controller should calibrate its own dB-per-qdB, or close the loop on the
+ground's RSSI.
+
+The control that makes those numbers readable: the **pre-change binary, in the
+same session and geometry, measured flat** — 0.3 dB across the same 64 qdB of
+command, because it had no actuator to move. One unit, one witness, near-field
+geometry, integer-quantised RSSI, no SDR. Harnesses:
+`tests/rtl8733b_txpwr_regcheck.sh` (registers, including that null control),
+`tests/txpwr_offset_onair.sh` (slope), `tests/rtl8733b_txpwr_selftest.cpp`
+(the offset math, in `ctest`).
+
+### Overdrive: about 3 dB, and then the PA compresses
+
+Sweeping the other way — up from the clip, MCS0 at ch36, with the witness
+reporting EVM beside RSSI:
+
+| offset | witness RSSI | witness EVM | witness SNR |
+|---|---|---|---|
+| 0 | 75.4 | −62.0 | 62.0 |
+| +16 (top of the PG table) | 78.2 | −50.2 | 63.7 |
+| +32 | 84.1 | **−18.0** | 57.9 |
+| +48 | 83.7 | −18.0 | 58.1 |
+| +64 | 83.8 | −18.0 | 58.2 |
+
+The +16 rung is real gain, +2.8 dB, though the constellation has already given
+up 12 dB of EVM to get it. The +32 rung is **not** gain: the witness hears
+8.7 dB more energy while EVM collapses to −18, and +48 and +64 change nothing
+at all — RSSI and EVM both pinned. That is the PA in hard compression, and
+`SNR never saw it` (58..64 throughout), which is the failure mode
+`docs/bench-testing-near-field.md` exists to warn about: strong RSSI plus poor
+EVM means back power off, the opposite of the weak-link response.
+
+Two things follow. The vendor's PG table lands about where this part stops
+being linear, so **+16 qdB is the edge of usable overdrive** even though the
+API allows +127 — a caller sweeping for its own operating point should watch
+EVM, not RSSI, and stop where EVM turns. And the negative half is clean by
+comparison: EVM sits flat at −58..−61 across all 16 dB of backoff.
+
 The chip keeps two thermal-compensation curves, one for CCK and one for
 OFDM/HT. The table is chosen once per channel set from the configured TX mode
 and then left alone, which is what the vendor driver does —
@@ -189,7 +270,14 @@ a request a caller may be making only through an inherited environment.
 These results have **not** been claimed:
 
 - No SDR was available, so occupied bandwidth, spectral mask, EVM and absolute
-  output power were not measured.
+  output power were not measured. The runtime power lever's slope is a
+  *relative* witness-RSSI measurement for the same reason: it says the lever
+  moves ~14 dB, not what any rung radiates in dBm.
+- The TX-power offset's dB-per-qdB is not constant across its advertised range
+  (0.125 in the bottom 12 qdB vs 0.233..0.242 above −52), so
+  `step_measured` stays false and a controller must calibrate its own slope.
+  Raising `kSafeTssiTargetQdbm8733b` above 16 dBm is deferred — that wants a
+  conducted measurement, not a witness receiver.
 - The experimental 5/10 MHz sequence has register-readback and normal-path RX
   evidence only. Narrowband TX and decode by an independent narrowband peer are
   deferred; `AdapterCaps::narrowband_ok` remains false.
--- a/docs/rtl8733b.md
+++ b/docs/rtl8733b.md
@@ -148,7 +148,7 @@ shape `fast_retune` uses for its per-channel rewrite (#389); the alternative,
 a TSSI disable/re-enable pair, costs ~165 ms and buys nothing here.
 
 The capability report is the dBm model — `index_max = 0`, one qdB per step —
-over `[-64, +127]`. Offset 0 is `kSafeTssiTargetQdbm8733b` (16 dBm), a
+over `[-128, +127]`. Offset 0 is `kSafeTssiTargetQdbm8733b` (16 dBm), a
 first-light clip sitting at or below this part's factory targets (18–20 dBm at
 2.4 GHz, 16–19 at 5 GHz).
 
--- a/docs/rtl8733b.md
+++ b/docs/rtl8733b.md
@@ -156,14 +156,11 @@ That range is asymmetric, and only one end is a hardware limit. The
 `0x3a00..0x3a10` bytes hold a **signed int8 offset from the 64 qdBm anchor**, so
 the field itself spans [−128, +127] — targets from −16 dBm to +47.75 dBm.
 
-- **+127 is the field**, the same clamped-only-at-the-hardware-rail answer
-  Jaguar1 (±126) and Jaguar3 (±127) give.
-- **−64 is a judgement, not the field's floor** (−128 is). It is where the
-  absolute target reaches 0 dBm. Stopping there is backed by the sweep: the
-  bottom 12 qdB of travel delivers only 0.125 dB/qdB against 0.25 nominal —
-  reproduced exactly across two passes — so the loop is already running out of
-  authority as its target approaches zero, and nothing below 0 dBm has been
-  characterised. `saturated_low` marks that boundary.
+Both ends are that field, and neither is re-clamped at something softer — the
+same clamped-only-at-the-hardware-rail answer Jaguar1 (±126) and Jaguar3 (±127)
+give. Where the chip stops responding is measured and written down below rather
+than enforced in the setter, because on this part both ends stop responding well
+inside the field and at points only a bench can find.
 
 The positive half is deliberately **not** re-clamped at the factory table.
 `src/TxPower.h` states for every family that headroom above the generated
@@ -200,6 +197,28 @@ geometry, integer-quantised RSSI, no SDR. Harnesses:
 `tests/rtl8733b_txpwr_regcheck.sh` (registers, including that null control),
 `tests/txpwr_offset_onair.sh` (slope), `tests/rtl8733b_txpwr_selftest.cpp`
 (the offset math, in `ctest`).
+
+### The backoff floor is about −96 qdB, not the 0 dBm target
+
+An earlier cut of this work clamped the offset at −64 qdB, reasoning that a
+target below 0 dBm was meaningless. Sweeping past it says otherwise (MCS0,
+ch36):
+
+| offset | target | witness RSSI | witness EVM |
+|---|---|---|---|
+| −32 | 8.0 dBm | 69.40 | −60.8 |
+| −64 | 0.0 dBm | 62.07 | −59.3 |
+| −80 | −4.0 dBm | **54.80** | −54.0 |
+| −96 | −8.0 dBm | 52.96 | −53.1 |
+| −112 | −12.0 dBm | 52.97 | −52.8 |
+| −128 | −16.0 dBm | 52.99 | −52.8 |
+
+Power keeps falling past the 0 dBm target — another **7.3 dB** between it and
+−4 dBm — with no sign wrap, and pins from about **−96 qdB**, where three
+successive rungs agree within 0.03 dB. EVM softens from −59 to −53 across the
+whole extension and the link stays decodable throughout. So usable travel is
+~23 dB below the clip rather than the ~16 dB the first cut allowed, which
+matters for near-field bench work and for a link that wants to sit quiet.
 
 ### Overdrive: about 3 dB, and then the PA compresses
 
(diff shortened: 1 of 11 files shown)


snokvist:
Pushed `03daab5`: **the offset floor moves from −64 to −128 qdB**, and the PR description's range statement is now out of date in one direction — the measurement below supersedes it.

Review question that prompted it: if *the field is the limit and the operator judges* holds at +127, why did the bottom stop at −64? It didn't have a good answer. −64 was a 0 dBm absolute target, clamped there on the assumption that a negative target is meaningless — an assumption, not a measurement, in a change whose whole subject is knobs that assume instead of checking. So I swept past it (MCS0, ch36, witness RSSI + EVM):

| offset | target | RSSI | EVM |
|---|---|---|---|
| −32 | 8.0 dBm | 69.40 | −60.8 |
| −64 | 0.0 dBm | 62.07 | −59.3 |
| −80 | −4.0 dBm | **54.80** | −54.0 |
| −96 | −8.0 dBm | 52.96 | −53.1 |
| −112 | −12.0 dBm | 52.97 | −52.8 |
| −128 | −16.0 dBm | 52.99 | −52.8 |

Power keeps falling straight past the 0 dBm target — **another 7.3 dB**, no sign wrap — and pins from about **−96 qdB**, where three successive rungs agree within 0.03 dB. EVM softens −59 → −53 across the extension and the link stays decodable throughout.

The old clamp was discarding **~9 dB of working backoff**. Usable travel is ~23 dB below the 16 dBm clip, not ~16 dB.

So both ends of the range are now the int8 delta field, and both knees are documented rather than enforced: **~−96 qdB** where the loop stops reducing, **~+32 qdB** where the PA compresses (EVM −18). `saturated_low` now means the field rail; the selftest pins that −64 rails nothing and the ladder keeps its calibrated spread through it, −100 still preserves that spread, and −128 rails the colder rate while the hot one sits exactly on the floor.

Re-verified on the device: regcheck **7/7**, on-air 0.223 dB/qdB over the shipped −64..0 range, ctest **54/54**.

src/rtl8733b/Rtl8733bDevice.h
```diff
@@ -75,6 +91,13 @@ class Rtl8733bDevice : public IRtlDevice {
   bool _phy_ready = false;
   bool _tx_ready = false;
   bool _tssi_tracking = false;
+  /* Session TX-power offset (qdB, <= 0) and the rails the last apply hit.
+   * Sticky by construction: configure_tx_power folds it back in on every
+   * channel set, and FastRetune passes it to the in-place hop rewrite. */
+  int16_t _tx_offset_qdb = 0;
```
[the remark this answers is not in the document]
snokvist: Confirmed and fixed in `2956621` — thanks, this was a real contradiction and both sites were stale.

`_tx_offset_qdb`'s "qdB, <= 0" comment and root `CLAUDE.md`'s `[-64, 0]` were written when the range genuinely was one-sided. Neither was updated when the positive half landed, and by then they were doubly stale: the floor had also moved from −64 to −128 after a bench sweep showed power keeps falling ~7.3 dB past the 0 dBm target and only pins around −96 qdB.

Both now state `[-128, +127]` — the int8 delta field at each end — with the measured knees recorded as documented rather than enforced (~−96 qdB where the loop stops reducing, ~+32 qdB where the PA compresses and only EVM shows it). I also swept the rest of the tree for the same claim; nothing else carried it. ctest 54/54.

src/rtl8733b/Phy8733b.h
```diff
@@ -78,6 +78,28 @@ inline constexpr uint8_t kSafeTxAgcIndex8733b = 0x10;
...
+ * it is the reference point that says where the vendor's calibration ends and
+ * the operator's own begins.  Pinned against the generated table by
+ * tests/rtl8733b_txpwr_selftest.cpp so the figure quoted in the docs cannot
+ * drift away from the data. */
+inline constexpr uint8_t kMaxPgTargetQdbm8733b = 80;
+
+/* Which rail the runtime TX-power offset clamped at, if any — the signal a
+ * closed-loop controller uses to know the knob has run out of travel
+ * (IRtlDevice::GetTxPowerState).  `low` is set when a rate's shifted target hit
+ * the 0 qdBm floor (the low bound is that absolute floor, not the int8 field's
```
josephnef: **Stale — this documents the inverse of the shipped clamp.** The comment says `low` fires at "the 0 qdBm floor (the low bound is that absolute floor, not the int8 field's -128)", which describes the pre-`03daab5` cut. The code now clamps at `shifted < -64` — i.e. exactly the int8 delta field's −128 (a −16 dBm target) — and deliberately does **not** stop at 0 qdBm; that's the whole point of the ~9 dB correction. This is the load-bearing struct a closed-loop controller reads to know the knob ran out of travel, so it shouldn't be documented backwards. Suggested: `low` is set when a rate's shifted target hit the int8 delta field's −128 floor (a shifted target below −64 qdBm); `high` when it hit the field's +127 ceiling.

docs/rtl8733b.md
```diff
@@ -137,6 +137,116 @@ not survive the link. Measured with the witness, 300 frames submitted at MCS7
...
+Because the loop is the power control, the runtime knob is the loop's target.
+`IRtlDevice::SetTxPowerOffsetQdb` shifts every per-rate target that survives
+the safe ceiling — cap first, then shift, so the calibrated per-rate spread
+comes through the move intact — and writes it as the five packed dwords at
+`0x3a00..0x3a10`, **in place, with tracking left enabled**. That is the same
+shape `fast_retune` uses for its per-channel rewrite (#389); the alternative,
+a TSSI disable/re-enable pair, costs ~165 ms and buys nothing here.
+
+The capability report is the dBm model — `index_max = 0`, one qdB per step —
+over `[-64, +127]`. Offset 0 is `kSafeTssiTargetQdbm8733b` (16 dBm), a
```
josephnef: **Stale range.** `GetTxPowerCaps` reports `offset_min_qdb = -128`, and this same section says three paragraphs later that both ends are the int8 field `[-128, +127]` — as do both CLAUDE.md updates and the caps comment. The `[-64, +127]` here is a leftover from the cut this PR's own description corrects, and it's internally contradictory within one section.

tests/rtl8733b_txpwr_selftest.cpp
```diff
@@ -0,0 +1,217 @@
...
+   *     That last part is the point — a per-unit EFUSE trimmed too cold is the
+   *     case an operator calibrates their own operating point for, and this is
+   *     the only lever this backend gives them to command it. */
+  rtl8733b::TssiOffsetSat8733b sat_up;
+  const auto up16 = rtl8733b::Phy8733b::tssi_rate_offsets(t, 0, 0, kCeiling,
+                                                          16, &sat_up);
+  expect("+16 qdB shifts the ladder up, shape intact",
+         up16 && (*up16)[0] == 16 && (*up16)[7] == 12 &&
+             ((*up16)[0] - (*up16)[7]) == ((*base)[0] - (*base)[7]));
+  expect("a positive offset within the field rails nothing",
```
josephnef, line 120: Minor: missing `up16 &&` in the condition. Rails publish only on success, so a nullopt return from `tssi_rate_offsets` leaves `sat_up` default-false and this cell passes vacuously. The adjacent `+16 qdB shifts the ladder up` cell would still catch the nullopt, so low severity — but the guard is free. Take it or leave it.

josephnef (dismissed):
**Approve, with the two stale-doc fixes below requested before merge.** Both are leftovers from the pre-`03daab5` cut (the −64 floor that cost ~9 dB) that survived the `2956621` sweep of stale one-sided range claims.

What I verified against the head, by hand:

- **Cap-then-shift math**: traced every selftest cell against the implementation — `-64` passes through the 0 dBm target unrailed (−64/−68 on the synthetic ladder), `-128` puts the anchor exactly on the field floor while the colder rate rails with `sat.low`. The `shifted < -64` clamp = int8 delta −128 = −16 dBm target, all arithmetic in `int`, no wrap.
- **Rails accumulate locally, publish only on success** — a nullopt caller can't see partial flags.
- **Refuse-before-first-write and rollback-on-failed-readback** in `set_tssi_offset` are both real, and `_fr_tssi_offsets` is correctly left describing the state being restored.
- **Locking**: `SetTxPowerOffsetQdb` / `GetTxPowerState` / `FastRetune` / `SetMonitorChannel` all take `_reg_mu`, including the FastRetune → full-path fallback.
- **Stickiness plumbing**: `configure_tx_power` re-folds on every channel set — and the post-enable `set_tssi_offset` call is a nice no-write way to derive chip-truth saturation flags — while `FastRetune` carries the offset into the hop rewrite.
- **`tssi_offsets_confirmed` recomputes from `(channel, ceiling, offset)` rather than checking the shadow against itself** — the right design in a PR that exists to abolish self-agreeing shadows, and it makes `hw_readback` genuinely chip-derived on a write-mostly family.

The validation record is unusually complete for a one-unit backend: master-build null control (what makes the 14 dB readable), four on-air passes including a cold replug, both-rail sweeps, a mutation-killing selftest, and `kMaxPgTargetQdbm8733b` pinned against the generated table. Every favourable number carries its adversarial counterpart in the same breath, per the house rule.

One non-blocking observation, no change requested: on the flat-PG path `GetTxPowerState` reports `flat_index = agc.ofdm_ref_a`. Per `src/TxPower.h`, `flat_index >= 0` conventionally reads as "a flat override is active", which no caller can have set here (the override setter refuses). It *is* chip truth — the chip really does run a flat index — but a one-line comment at the site acknowledging the reading would spare a generic consumer trying `SetTxPowerIndexOverride(-1)` to clear it.

src/rtl8733b/Phy8733b.h
snokvist: Confirmed and fixed in `2b5d3a8`. You're right that it documented the inverse: the clamp is `shifted < -64` → `delta = shifted - 64` = **-128**, so `low` has always fired at the field floor, and the comment described the rule that measurement had just removed.

Now:

> `low` is set when a rate's shifted target hit the int8 delta field's -128 floor — a shifted target below -64 qdBm, i.e. -16 dBm — and deliberately NOT at the 0 qdBm target, which the loop keeps responding past by ~7 dB; `high` when a rate hit the field's +127 ceiling. Both rails are that field and nothing softer.

Agreed on why this one mattered most of the three — it's the field a controller reads to decide it's out of travel, and backwards there means a controller stops backing off ~9 dB early, which is the exact quantity the `03daab5` sweep recovered.

docs/rtl8733b.md
snokvist: Fixed in `2b5d3a8` — now `[-128, +127]`, matching `GetTxPowerCaps`, the caps comment, both CLAUDE.md files and the derivation three paragraphs below it in this same section.

I also swept the tree again for the same claim rather than just patching the cited line, and it caught one more you didn't flag: `tests/txpwr_offset_onair.sh:65` called 0 qdBm "its 0 qdBm floor". The slope compression measured there is real and stays documented, but it isn't the floor — reworded to say the sweep *stops* at -64 because that's where the shipped monotone assertion holds, not because the lever ends there.

tests/rtl8733b_txpwr_selftest.cpp
snokvist: Taken — `up16 &&` added in `2b5d3a8`. Your reading is right: rails publish on success only, so a nullopt leaves `sat_up` default-false and the cell asserts nothing. Cheap guard, and a vacuous pass in a selftest whose whole job is to kill mutations is worth closing even at low severity.

ctest 54/54 after all three fixes.

snokvist:
Thanks for the review — all three requested fixes are in `2b5d3a8`, threads replied to and resolved. ctest 54/54, build clean.

Summary of the three, since they share one root cause: each was a leftover from the pre-`03daab5` cut, where the floor was −64 qdB (a 0 dBm target). The `2956621` sweep caught the *one-sided range* claims but not the *floor location* claims, which are a different sentence shape. The `TssiOffsetSat8733b` one was the worst of them — a controller reading `low` per that comment stops backing off about 9 dB early, which is exactly the travel the sweep recovered. Sweeping again for the floor phrasing turned up a fourth site you didn't cite, `tests/txpwr_offset_onair.sh:65`, fixed in the same commit.

**On the non-blocking `flat_index` observation** — I took it, as a comment rather than a value change. You've read it exactly as intended: it is chip truth, the unit really does run a flat index, and there is no override to clear. But `src/TxPower.h`'s convention does say `flat_index >= 0` means an override is active, so a generic consumer can reasonably try `SetTxPowerIndexOverride(-1)` to release something nothing set. That call now meets a logged refusal, which is the honest answer, and the site says so:

> `flat_index >= 0` reads as "a flat override is active" per `src/TxPower.h`, which no caller can have set here — `SetTxPowerIndexOverride` refuses, and there is nothing to clear. The index is simply what this unit runs at, because a no-TSSI-calibration EFUSE leaves bring-up's flat index as the level.

I'd rather leave the value as chip truth than report −1 to satisfy the convention, since −1 would claim "the EFUSE per-rate baseline is active" on a unit where it isn't. Happy to reverse that if you'd rather the convention win over the reading.

One scope note for the record: **no hardware coverage exists for that flat-PG branch** — every unit seen so far is TSSI-offset PG, so the path is reasoned and reviewed but never executed on silicon. It writes no registers, only reads. Called out here rather than buried because it's the one part of this PR the bench record doesn't cover.

The device suites were not re-run for this commit: the diff is four comments, one doc range, and one test guard — no register sequence, no arithmetic, no plan change. Last full device pass on `2956621` stands (regcheck 7/7, on-air 0.223 dB/qdB, cold-replug included). Say the word if you'd like a re-run on the head anyway.

josephnef (approved):
Re-approving on `2b5d3a8` — verified all three fixes on the head: the `TssiOffsetSat8733b` comment now states the shipped clamp correctly (int8 field's −128, deliberately not 0 qdBm), `docs/rtl8733b.md` reads `[-128, +127]` consistently, and the `up16 &&` guard is in. Credit for catching the fourth instance of the same stale claim in `txpwr_offset_onair.sh` that my pass missed, and the `flat_index` convention note at the flat-PG site is exactly what I had in mind. LGTM.

PR #406, opened 2026-08-30
rtl8733b: port the hardware-ARQ surface — ACK responder, retry limit, honoured ACK window

snokvist:
Closes the four items in snokvist/devourer#2, filed by a consumer scoping a ground→craft return path on the 8733BU. Three are now measured working on hardware; the fourth is root-caused and refused out loud rather than faked.

## What changed

**1. `tx.ack_timeout_us` was silently ignored.** `init_wmac()` hardcoded REG_ACKTO to `0x21` (33 µs) and the config field was read nowhere in `src/rtl8733b/`. So the one library default every generation programs identically was, on this die alone, pinned to the bottom of the 33..128 µs spread that default exists to abolish. Now applied at bring-up with the same register and 1..255 clamp the Jaguars use, and **read back before it is claimed**:

| `DEVOURER_ACK_TIMEOUT_US` | `<unset>` | 128 | 33 | 200 |
|---|---|---|---|---|
| REG_ACKTO reads | 128 | 128 | 33 | 200 |

The vendor write stays in `init_wmac()` — that MAC plane is shared verbatim with `rtl8733bprobe`, which wants the vendor recipe and carries no `DeviceConfig`.

**2. `SetAckResponder` ported and measured** → `ack_responder_ok = true`. The `src/AckResponder.h` recipe applies unchanged, and that is not an assumption: the vendor tree's own port-0 descriptor names these three registers (net_type `REG_CR_8733B + 2` = 0x0102, MACID 0x0610, BSSID 0x0618). MAC bring-up leaves net_type at No Link because `init_mac` writes only REG_CR's low half — exactly why a monitor radio here never ACKed.

| phase (8733B as responder, 8812AU soliciting) | ack_rate | mean retries | frames |
|---|---|---|---|
| armed | **1.00** | 0.00 | 1736/1736 |
| re-armed on a *different* MAC | **1.00** | 0.00 | 1733/1733 |
| disarmed (control) | **0.00** | 12.0 (pinned at limit) | 1714 |

The arm refuses a group MAC, is read back before being reported, and **teardown disarms it** — `_mac.stop()` clears only REG_CR's low half, so a session ending with `teardown_power_down` off would otherwise leave an unowned radio auto-ACKing. Verified: after such a session the peer reads ack_rate 0.00.

**3. `tx.retry_limit` drives real autonomous retransmission** → `tx_retry_limit_ok = true`. This could not be measured the way the Jaguars were — that A/B reads the TX side's own CCX reports, and this die has none. New `tests/rtl8733b_retry_limit_onair.sh` judges from the air: unicast to an unowned RA so no ACK ever returns, a passive monitor counting airings per submitted frame. It takes a **dose-response** rather than an on/off pair, because one pair could be ambient and a straight line through three levels cannot:

| retry_limit | 0 | 3 | 12 | 0 | 12 |
|---|---|---|---|---|---|
| airings/frame (expected 1+N) | 0.93 | 3.93 | 12.27 | 0.93 | 12.27 |

**4. CCX / `tx.report` is NOT ported — the cause is the firmware.** Everything under this backend's control was verified correct on air: descriptor SPE_RPT and SW_DEFINE at the vendor's own bit positions, C2H decoded at the vendor's `dword2[28]`, bulk-IN as the vendor's delivery path (its USB interrupt handler is an empty stub), and the fw-offload C2H format this firmware speaks per its own dispatch. With an RX loop live and 3160 frames received, the chip returned **zero** C2H packets in any format — with and without net_type armed, peer both ACKing and silent.

The plumbing is therefore **not shipped**: descriptors stay byte-identical and the knob warns at bring-up. A knob that looks granted while producing nothing is the exact failure the issue was filed about. The outstanding lead (halmac H2C queue + `MEDIA_STATUS_RPT` to register the descriptor MACID — this backend has no H2C transport at all) is recorded so the next attempt starts where this one stopped.

## Counterparts

Stated because the numbers above are uniformly favourable: **one physical unit and one peer generation** (`f72b` RTL8731BU vs RTL8812AU), no second responder die, no vendor-driver A/B, and the second `b733` sample has not run these cells. The retry figure is witnessed by a monitor that can only *lose* airings, never invent them, so it is a floor — its ~0.07/frame shortfall is the `rx.txhit` sampling quantization the harness documents.

## Verification

- Builds clean and `ctest` green on **default / 8733B-only / 8733B-off / ASan+UBSan** (54/50/49/54 tests).
- Lifecycle soak 6/6 HEALTHY; ASan teardown clean on hardware.
- Monitor RX unaffected by arming the responder (201/197/181/210 frames across alternating arms, no trend).
- Added the 8733B to `tests/teardown_gen_sanity.sh`, whose cell list predated the chip.
- `docs/scheduled-mac.md`'s cross-generation ARQ matrix updated — the 8733B is a new combination there: closed-loop ARQ **without** a CCX report.

https://claude.ai/code/session_01VjcZRPjNtsy2mxfDkAWNpd

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -59,9 +59,10 @@ construction from the `SYS_CFG2` chip-id (Kestrel: PID-first):
   only what an independent witness decoded — legacy OFDM + HT MCS0-7, BCC,
   20/40 MHz on 2.4/5 GHz, plus long-preamble CCK on 2.4 GHz at 20 MHz, plus
   10 MHz narrowband (5 MHz refused — `src/rtl8733b/CLAUDE.md`).
-  Everything the backend has not ported (TSF/beacons, hardware ACK, A-MPDU,
+  Everything the backend has not ported (TSF/beacons, A-MPDU, CCX/`tx.report`,
   the flat-index and per-rate TX-power knobs) falls through to `IRtlDevice`'s
-  not-ported defaults rather than being faked, so read the base class before
+  not-ported defaults rather than being faked (hardware ACK/BlockAck IS ported
+  and measured, as is `tx.retry_limit` — `src/rtl8733b/CLAUDE.md`), so read the base class before
   assuming a cross-generation feature below applies here. `FastRetune` IS
   ported (intra-band, TSSI kept live — `src/rtl8733b/CLAUDE.md`). SGI, LDPC, STBC, VHT
   and HE are refused outright. Scope, one-unit validation record and the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -61,8 +61,7 @@ construction from the `SYS_CFG2` chip-id (Kestrel: PID-first):
   10 MHz narrowband (5 MHz refused — `src/rtl8733b/CLAUDE.md`).
   Everything the backend has not ported (TSF/beacons, A-MPDU, CCX/`tx.report`,
   the flat-index and per-rate TX-power knobs) falls through to `IRtlDevice`'s
-  not-ported defaults rather than being faked (hardware ACK/BlockAck IS ported
-  and measured, as is `tx.retry_limit` — `src/rtl8733b/CLAUDE.md`), so read the base class before
+  not-ported defaults rather than being faked, so read the base class before
   assuming a cross-generation feature below applies here. `FastRetune` IS
   ported (intra-band, TSSI kept live — `src/rtl8733b/CLAUDE.md`). SGI, LDPC, STBC, VHT
   and HE are refused outright. Scope, one-unit validation record and the
--- a/docs/rtl8733b.md
+++ b/docs/rtl8733b.md
@@ -372,13 +372,13 @@ a request a caller may be making only through an inherited environment.
 
 ## Hardware ARQ
 
-Opened by issue #2; measured against an RTL8812AU peer on the validation unit.
+Measured against an RTL8812AU peer on the validation unit.
 Both directions of a reliable-unicast link now work on this die.
 
 | knob | state | evidence |
 | --- | --- | --- |
-| `tx.ack_timeout_us` | honoured (was silently pinned at 33 us) | REG_ACKTO read back at bring-up: `<unset>/128/33/200 -> 128/128/33/200` |
-| `SetAckResponder` | ported, `ack_responder_ok = true` | 8733B as responder: armed 1736/1736 ACKed, retries_mean 0.00; re-armed on a different MAC 1736/1736; disarmed 0.00 with retries pinned at 12 |
+| `tx.ack_timeout_us` | honoured | REG_ACKTO read back at bring-up: `<unset>/128/33/200 -> 128/128/33/200` |
+| `SetAckResponder` | ported, `ack_responder_ok = true` (normal-ACK singles; BlockAck untested) | 8733B as responder: armed 1736/1736 ACKed, retries_mean 0.00; re-armed on a different MAC 1736/1736; disarmed 0.00 with retries pinned at 12 |
 | `tx.retry_limit` | live, `tx_retry_limit_ok = true` | airtime dose-response 0/3/12 -> 0.93/3.93/12.27 airings per frame (expected 1 + N), repeatable over a 0/3/12/0/12 ladder |
 | `tx.report` (CCX) | **not** ported — firmware emits nothing | see Known gaps below |
 
@@ -394,8 +394,13 @@ single delta that ambient conditions could fake — carries the claim. Its
 counterparts: one physical unit and one peer generation, like every other
 on-air claim in this section — note the document as a whole rests on two units,
 but these ARQ cells ran on the `f72b` sample only; the monitor can only lose
-airings, never invent them, so the measured ratio is a floor; and the ~0.07/frame shortfall is the `rx.txhit`
-sampling quantization the harness documents rather than loss.
+airings, never invent them, so the measured ratio is a floor. Two different things account for the shortfalls, and only the smaller one is an
+artefact: the `rx.txhit` readout quantizes to at most 99 airings (the event
+fires on the first 10 hits then every 100th), which is <=0.066/frame at 1500
+frames and covers the 0 and 3 arms entirely. It does NOT cover the 12 arm —
+12.27 against 13 is 0.73/frame, about 1095 airings, an order of magnitude past
+that bound. That residue is monitor loss (or genuinely fewer airings), which is
+why the ratio is reported as a floor and not a point estimate.
 
 ## Known gaps and deferred validation
 
@@ -432,8 +437,9 @@ These results have **not** been claimed:
   successfully, but an independent RTL8812AU decoded both broadcast and
   unicast probes as long GI. STBC and LDPC remain rejected. A-MPDU and
   throughput were not independently validated and remain unadvertised or
-  unsupported by this backend. (ACK/BlockAck response is no longer in this
-  list — it is ported and measured, see Hardware ARQ below.)
+  unsupported by this backend. Normal-ACK response IS ported and measured
+  (Hardware ARQ above); BlockAck response to an aggregate remains untested, as
+  does A-MPDU itself.
 - No CCX path, so no `tx.report` events — and unlike every other gap here the
   cause is the firmware, not a missing port. Everything under the backend's
   control was verified correct on air (descriptor SPE_RPT/SW_DEFINE set at the
--- a/docs/rtl8733b.md
+++ b/docs/rtl8733b.md
@@ -370,6 +370,33 @@ silently or failing to initialise. Silence is the only outcome that would let a
 refused knob look like a granted one; a hard failure would be a harsh answer to
 a request a caller may be making only through an inherited environment.
 
+## Hardware ARQ
+
+Opened by issue #2; measured against an RTL8812AU peer on the validation unit.
+Both directions of a reliable-unicast link now work on this die.
+
+| knob | state | evidence |
+| --- | --- | --- |
+| `tx.ack_timeout_us` | honoured (was silently pinned at 33 us) | REG_ACKTO read back at bring-up: `<unset>/128/33/200 -> 128/128/33/200` |
+| `SetAckResponder` | ported, `ack_responder_ok = true` | 8733B as responder: armed 1736/1736 ACKed, retries_mean 0.00; re-armed on a different MAC 1736/1736; disarmed 0.00 with retries pinned at 12 |
+| `tx.retry_limit` | live, `tx_retry_limit_ok = true` | airtime dose-response 0/3/12 -> 0.93/3.93/12.27 airings per frame (expected 1 + N), repeatable over a 0/3/12/0/12 ladder |
+| `tx.report` (CCX) | **not** ported — firmware emits nothing | see Known gaps below |
+
+Implementation detail, the register recipe and the full CCX bench narrative:
+`src/rtl8733b/CLAUDE.md` "Hardware ARQ". The cross-generation ARQ matrix this
+die now joins is `docs/scheduled-mac.md`.
+
+The retry measurement could not use the 12/0/12 CCX A/B the Jaguars are judged
+by, because this die has no `tx.report` to judge itself with; it counts airings
+at a passive monitor instead (`tests/rtl8733b_retry_limit_onair.sh`), and takes
+three dose levels rather than an on/off pair so that a straight line — not a
+single delta that ambient conditions could fake — carries the claim. Its
+counterparts: one physical unit and one peer generation, like every other
+on-air claim in this section — note the document as a whole rests on two units,
+but these ARQ cells ran on the `f72b` sample only; the monitor can only lose
+airings, never invent them, so the measured ratio is a floor; and the ~0.07/frame shortfall is the `rx.txhit`
+sampling quantization the harness documents rather than loss.
+
 ## Known gaps and deferred validation
 
 These results have **not** been claimed:
@@ -403,9 +430,24 @@ These results have **not** been claimed:
   the normal NIC image.
 - SGI remains disabled: a descriptor with its short-GI bit set submitted
   successfully, but an independent RTL8812AU decoded both broadcast and
-  unicast probes as long GI. STBC and LDPC remain rejected. ACK/BlockAck
-  response, A-MPDU and throughput were not independently
-  validated and remain unadvertised or unsupported by this backend.
+  unicast probes as long GI. STBC and LDPC remain rejected. A-MPDU and
+  throughput were not independently validated and remain unadvertised or
+  unsupported by this backend. (ACK/BlockAck response is no longer in this
+  list — it is ported and measured, see Hardware ARQ below.)
+- No CCX path, so no `tx.report` events — and unlike every other gap here the
+  cause is the firmware, not a missing port. Everything under the backend's
+  control was verified correct on air (descriptor SPE_RPT/SW_DEFINE set at the
+  vendor's own bit positions, C2H decoded at the vendor's dword2[28], bulk-IN
+  delivery as the vendor uses, and the fw-offload C2H format this firmware
+  speaks per its own dispatch), yet with an RX loop live and 3160 frames
+  received the firmware returned **zero** C2H packets in any format — with and
+  without net_type armed, peer ACKing and silent. The outstanding lead is the
+  halmac H2C queue + a MEDIA_STATUS_RPT registering the descriptor MACID; this
+  backend has no H2C transport at all. The knob warns at bring-up instead of
+  stamping descriptors that buy nothing. Consequence: an 8733BU can be either
+  end of a hardware-ARQ link but cannot see per-frame delivery, so detecting a
+  departed peer needs an application-level timeout
+  (`src/rtl8733b/CLAUDE.md` "Hardware ARQ").
 - Fast retune is now ported and independently witnessed (intra-band,
   same-width; `src/rtl8733b/CLAUDE.md` has the measured contract and its
   counterparts): channel-state readback parity 7/7 hops, a 299/300 post-hop
--- a/docs/scheduled-mac.md
+++ b/docs/scheduled-mac.md
@@ -258,8 +258,8 @@ nonzero limit for absolute numbers; 8821AU row re-measured ch6):
 | 8821AU | 62% | 0% | works (94% closed-loop at retry 8) |
 | 8812EU | 98% | 0% | works |
 | 8812CU | 69% | 0% | works |
-| 8733B | 100% (retry 12) | 0% | works (1736/1736, retarget- and disarm-proof) |
-| 8852CU (Kestrel) | 0% | 0% | not implemented (SetAckResponder is J1/2/3-only) |
+| 8733B | unmeasured | 0% | works (closed-loop 1736/1736 at retry 12; single-shot cell never run) |
+| 8852CU (Kestrel) | 0% | 0% | not implemented on the AX generation |
 
 Unmeasured for lack of plugged hardware: 8821CU / PCIe 8821CE (recipe-shared
 with the 8822B; their `AdapterCaps.ack_responder_ok` stays false-as-unmeasured
--- a/docs/scheduled-mac.md
+++ b/docs/scheduled-mac.md
@@ -173,6 +173,13 @@ scheduled MAC runs TX+RX anyway, so this is the relevant session shape.
 | Jaguar2 8812BU | 0.91 / 2.1 (run-to-run 0.12–0.91) | 0.64 / 5.3 | yes (12) | 0.86 | 0 |
 | Jaguar3 8822CU | 1.00 / 0.24 | 1.00 / 0.13 | yes (12) | 0.96 | 0 |
 
+The RTL8733B is deliberately absent from this table: it is the one die that
+runs closed-loop hardware ARQ with **no CCX report at all**, so none of these
+columns can be filled for it as the soliciting TX. Its retry knob is measured
+from the air instead (airings per submitted frame, 0/3/12 -> 0.93/3.93/12.27)
+and its responder side is measured with this same harness pointed the other
+way — `docs/rtl8733b.md` "Hardware ARQ".
+
 The OFF-phase pin is set by `DEVOURER_TX_RETRY_LIMIT` (the matrix runs 12,
 the value the descriptors used to hardcode) — the knob, not a descriptor
 constant, is now the single source of truth for the retry limit on
@@ -251,6 +258,7 @@ nonzero limit for absolute numbers; 8821AU row re-measured ch6):
 | 8821AU | 62% | 0% | works (94% closed-loop at retry 8) |
 | 8812EU | 98% | 0% | works |
 | 8812CU | 69% | 0% | works |
+| 8733B | 100% (retry 12) | 0% | works (1736/1736, retarget- and disarm-proof) |
 | 8852CU (Kestrel) | 0% | 0% | not implemented (SetAckResponder is J1/2/3-only) |
 
 Unmeasured for lack of plugged hardware: 8821CU / PCIe 8821CE (recipe-shared
@@ -300,8 +308,11 @@ carry this table per die.
    2,048-bit window leaked 2,846 delivered frames out of coverage when a
    stalled spsc-fat pool drained ~3 k frames in one receipt interval; the
    8192 default clears that bench worst case ~2.7×).
-2. **Closed-loop hardware ACK + autonomous retry is GO on Jaguar1 and
-   Jaguar3** (100% delivery, retries ≈ 0.2–0.3) including retargeting an
+2. **Closed-loop hardware ACK + autonomous retry is GO on Jaguar1, Jaguar3 and
+   the RTL8733B** (100% delivery, retries ≈ 0.2–0.3; the 8733B closes the loop
+   as responder at 1736/1736 but has no CCX report of its own, so its TX-side
+   retry evidence is airtime rather than `tx.report` — `docs/rtl8733b.md`)
+   including retargeting an
    arbitrary UE MAC mid-session (re-arm `SetAckResponder`, change the
    descriptor RA — both fully dynamic). Requires a nonzero
    `DEVOURER_TX_RETRY_LIMIT` — the hardware ARQ loop retransmits until ACK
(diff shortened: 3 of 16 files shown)


src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -755,6 +806,81 @@ size_t Rtl8733bDevice::build_tx_block(const uint8_t *packet, size_t length,
...
+  if ((mac.bytes[0] & 0x01u) != 0) {
+    /* A station cannot ACK-target a group address. Refuse rather than arm a
+     * responder that can never fire — the same footgun AckResponder.h records
+     * from the AP-mode work. */
+    _logger->error("RTL8733B: ACK responder needs a UNICAST MAC (I/G set in "
+                   "{:02x}) — not armed",
+                   mac.bytes[0]);
+    return false;
+  }
+  devourer::ack::enable(_device, mac.data());
```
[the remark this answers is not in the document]
snokvist: Fixed in 2f90e9e — and this was a real bug, thank you. `_ack_armed` was latched after `ack::enable()` returned; that call is five register writes ending with the net_type field that actually arms the engine, and any of them can throw on a stalling device, including after the arming write reached hardware. The exception unwinds into `Init()`/`InitWrite()`'s `catch`, which calls `Stop()`, and with the flag still false the teardown skipped the disarm — leaving an unowned radio auto-ACKing when `teardown_power_down` is off. The flag is now latched before the first write, so every failure path (throw mid-enable, or a failed readback) leaves teardown believing it must disarm. A redundant disarm costs one USB round-trip; a missed one costs an unowned transmitter. Re-verified on hardware: responder 1349/1349 ACKed at retries_mean 0.00, and after an armed session ends with power-down disabled the peer reads ack_rate 0.00.

CLAUDE.md
```diff
@@ -59,9 +59,10 @@ construction from the `SYS_CFG2` chip-id (Kestrel: PID-first):
   only what an independent witness decoded — legacy OFDM + HT MCS0-7, BCC,
   20/40 MHz on 2.4/5 GHz, plus long-preamble CCK on 2.4 GHz at 20 MHz, plus
   10 MHz narrowband (5 MHz refused — `src/rtl8733b/CLAUDE.md`).
-  Everything the backend has not ported (TSF/beacons, hardware ACK, A-MPDU,
+  Everything the backend has not ported (TSF/beacons, A-MPDU, CCX/`tx.report`,
   the flat-index and per-rate TX-power knobs) falls through to `IRtlDevice`'s
-  not-ported defaults rather than being faked, so read the base class before
+  not-ported defaults rather than being faked (hardware ACK/BlockAck IS ported
+  and measured, as is `tx.retry_limit` — `src/rtl8733b/CLAUDE.md`), so read the base class before
```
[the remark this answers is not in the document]
snokvist: Fixed in 2f90e9e. Agreed — the root file is cross-cutting only, and the parenthetical I added ('hardware ACK/BlockAck IS ported and measured, as is tx.retry_limit') was subtree detail at the wrong scope. Dropped. The not-ported list itself is still updated there, since that list is the cross-cutting fact the root entry exists to carry; the ARQ narrative stays in `src/rtl8733b/CLAUDE.md`.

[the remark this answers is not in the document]
snokvist: Fixed in 2f90e9e, by the same edit as the scope finding above. You are right that a link to the nested document does not satisfy the counterpart rule — the counterpart has to sit in the same breath as the favourable claim, not one hop away. The claim is removed from the root file entirely. In `src/rtl8733b/CLAUDE.md` the Hardware ARQ section opens with its counterparts before any number is quoted: one physical unit and one peer generation, no second responder die, the `b733` sample has not run these cells, no vendor-driver A/B, and the retry figure is witnessed by a monitor that can only lose airings and never invent them, so it is a floor rather than a point estimate. The disarmed-control result (ack_rate 0.00, retries pinned at 12) is quoted alongside the armed one in every place the armed number appears.

src/rtl8733b/CLAUDE.md
```diff
@@ -276,6 +276,91 @@ request the setter refuses loudly instead vanishes without a word. The warning
...
+on-air claim in this subtree carries. Neither direction has been cross-checked
+against a second responder die, the second `b733` sample has not run these
+cells, and no vendor-driver A/B exists to compare against. The retry number in
+particular is witnessed by a monitor that can only ever LOSE airings, never
+invent them, so the measured ratio is a floor rather than a point estimate.
+
+`DeviceConfig::tx::ack_timeout_us` **is** honoured, and was not until that
+issue: `bring_up_to_phy` overwrites REG_ACKTO (0x0640) from the config with the
+same 1..255 clamp the Jaguars use, right after MAC bring-up wrote the vendor
+0x21 — 33 us, the bottom of the 33..128 per-chip spread that field's one
```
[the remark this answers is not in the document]
snokvist: Fixed in 2f90e9e. The `ack_timeout_us` restatement was the real violation — range, clamp, default, register address and the rationale for the single library-wide default are all doc-commented at the `DeviceConfig` declaration, and I had copied them. The subtree file now points at the header and keeps only what is specific to this backend: that `init_wmac()` still writes the vendor 0x21 and why (that MAC plane is shared verbatim with `rtl8733bprobe`, which carries no `DeviceConfig`), that bring-up overwrites it afterwards, and the readback evidence. A related duplicate you did not flag — the `GetAdapterCaps()` comment restating the same capability measurements as `AdapterCaps.h` — was trimmed to a pointer in the same spirit, matching how jaguar1/2/3 annotate that call site. The measurement figures that remain in `AdapterCaps.h` follow the per-die convention already used there for the 8821A/8814A rows.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -76,6 +77,40 @@ void Rtl8733bDevice::bring_up_to_phy() {
...
+                  ackto_got, ackto_want);
+  else
+    _logger->info("RTL8733B: ACK window {} us (REG_ACKTO 0x640, verified)",
+                  ackto_got);
+  /* DEVOURER_ACK_RESPONDER — opt-in only, never a default: it turns a passive
+   * monitor into an active SIFS-timed transmitter. Sited here with the other
+   * bring-up knobs so an RX-only session (Init) arms it too, which is the
+   * session shape a pure responder actually runs. */
+  if (_cfg.rx.ack_responder)
+    SetAckResponder(*_cfg.rx.ack_responder);
```
josephnef: **The return value of the config-requested arm is discarded**, so a refused arm — group MAC in `DEVOURER_ACK_RESPONDER`, or a readback mismatch — still yields a green init and a session running as a plain monitor.

Concrete trigger: an operator copies the canonical TX SA `57:42:75:05:d6:00` (first byte 0x57 has the I/G bit set); `SetAckResponder` logs an error and returns false, `Init()` succeeds anyway, the TX peer sees 0% ACK with retries pinned at the limit, and the operator debugs the RF link instead of the config. That is the exact silently-refused-knob failure this backend refuses everywhere else (`disable_cca`, `SetTxPowerIndexOverride`) — and the failure shape issue #2 was filed about. A config-driven arm that fails should fail the bring-up (or at minimum escalate past a buried error line): the caller asked for a responder, not a monitor.

docs/rtl8733b.md
```diff
@@ -403,9 +430,24 @@ These results have **not** been claimed:
   the normal NIC image.
 - SGI remains disabled: a descriptor with its short-GI bit set submitted
   successfully, but an independent RTL8812AU decoded both broadcast and
-  unicast probes as long GI. STBC and LDPC remain rejected. ACK/BlockAck
-  response, A-MPDU and throughput were not independently
-  validated and remain unadvertised or unsupported by this backend.
+  unicast probes as long GI. STBC and LDPC remain rejected. A-MPDU and
+  throughput were not independently validated and remain unadvertised or
+  unsupported by this backend. (ACK/BlockAck response is no longer in this
```
josephnef: **Overclaim past the measurement**: this removes "ACK/BlockAck response … not independently validated" and says "it is ported and measured" (with the matching wholesale removal of "hardware ACK/BlockAck" from `src/rtl8733b/CLAUDE.md`'s Not-ported list) — but only normal-ACK singles were measured. BlockAck response to a received A-MPDU never ran on this die, and A-MPDU itself remains unvalidated in this very bullet.

`src/AckResponder.h` promises the same gate is a hardware **BlockAck** responder; a consumer arms an 8733B against an A-MPDU peer on that promise, and if the HALMAC 87xx immediate-response engine differs for aggregates, every A-MPDU re-airs to the retry limit while the docs say this was "measured". Scope the claim to what ran: normal-ACK response measured; BA response untested pending A-MPDU validation.

tests/rtl8733b_retry_limit_onair.sh
```diff
@@ -0,0 +1,123 @@
...
+       DEVOURER_TX_QOS_DATA=1 DEVOURER_TX_RA=$RA \
+       DEVOURER_TX_RATE=$RATE DEVOURER_TX_PAYLOAD_BYTES=200 \
+       DEVOURER_TX_GAP_US=$GAP_US DEVOURER_TX_FRAMES=$FRAMES \
+       DEVOURER_TX_RETRY_LIMIT=$arm DEVOURER_TX_PWR_OFFSET_QDB=$PWR_QDB \
+       DEVOURER_LOG_LEVEL=warn \
+       timeout -s INT 90 "$BUILD/txdemo" >"$OUT/tx_$tag.jsonl" 2>"$OUT/tx_$tag.err" || true
+  sleep 3
+  sent=$(grep '"ev":"tx.stats"' "$OUT/tx_$tag.jsonl" | tail -1 |
+         sed -n 's/.*"submitted":\([0-9]*\).*/\1/p'); sent=${sent:-0}
+  hits=$(grep -o '"ev":"rx.txhit","hits":[0-9]*' "$OUT/wit_$tag.jsonl" | tail -1 |
```
josephnef, line 94: **`FRAMES` is an advertised knob with no floor, and this readout is quantized** — rxdemo emits `rx.txhit` for the first 10 hits then every 100th, so the value read here is the largest multiple of 100 ≤ the true count (or 10, below 100). Small `FRAMES` therefore false-FAILs healthy hardware: `FRAMES=60`, retry=0 → ~55 real airings → last `rx.txhit` reads `hits=10` → 0.17 airings/frame → FAIL against the ≥0.6 band, on a perfectly working retry engine. Even `FRAMES=150` sits one lost airing from a false FAIL.

The header comment bounds the quantization error only at the default `FRAMES=1500`. Since the script already refuses to report a non-run as zero, refuse `FRAMES` below a floor (~1000) the same way — it's the same false-verdict shape.

docs/scheduled-mac.md
```diff
@@ -251,6 +258,7 @@ nonzero limit for absolute numbers; 8821AU row re-measured ch6):
 | 8821AU | 62% | 0% | works (94% closed-loop at retry 8) |
 | 8812EU | 98% | 0% | works |
 | 8812CU | 69% | 0% | works |
+| 8733B | 100% (retry 12) | 0% | works (1736/1736, retarget- and disarm-proof) |
```
josephnef: **Closed-loop number in a single-shot column.** The table's preamble defines this column's 62–98% figures as single-shot ACK rates; "100% (retry 12)" is a closed-loop figure, so this row is not comparable to its neighbours — someone ranking responder dies here concludes the 8733B (100%) outperforms the 8812EU (98%) and picks it as the ground-station responder, when the 100% actually corresponds to the 8821AU row's *parenthetical* 94%-closed-loop figure, not the column's metric. The 8733B's single-shot cell is genuinely unmeasured; mark it that way and move the closed-loop 1736/1736 into a parenthetical like the 8821AU row does.

docs/scheduled-mac.md
```diff
@@ -251,6 +258,7 @@ nonzero limit for absolute numbers; 8821AU row re-measured ch6):
...
 | 8852CU (Kestrel) | 0% | 0% | not implemented (SetAckResponder is J1/2/3-only) |
```
josephnef: Stale one line below the new row: "SetAckResponder is J1/2/3-only" is now contradicted by the 8733B "works" row directly above it (and by `AdapterCaps.ack_responder_ok = true`). One-word fix per the PR's own AdapterCaps.h wording: "not implemented on the AX generation".

src/rtl8733b/CLAUDE.md
```diff
@@ -276,6 +276,94 @@ request the setter refuses loudly instead vanishes without a word. The warning
...
+
+**`tx.retry_limit` drives real autonomous retransmission** — `tx_retry_limit_ok`
+is now true. It could not be measured the way the Jaguars were: that A/B reads
+the TX side's own CCX reports, and this die has none. `tests/rtl8733b_retry_limit_onair.sh`
+judges from the air instead — unicast to an unowned RA so no ACK ever returns,
+a passive monitor counting airings per submitted frame — and takes a
+dose-response rather than an on/off pair, because one pair could be ambient and
+a straight line through three levels cannot. Measured 0 -> 0.93, 3 -> 3.93,
+12 -> 12.27 airings/frame against an expected 1 + N, repeatable across a
+0/3/12/0/12 ladder. The ~0.07 shortfall is the `rx.txhit` sampling
```
josephnef: **The quantization attribution is arithmetically wrong for the retry=12 arm** (same sentence in `docs/rtl8733b.md`, Hardware ARQ section). Quantization understates by at most 99 airings — ≤0.066/frame at FRAMES=1500. That covers the 0.93 and 3.93 arms, but 12.27 vs the expected 13 is a 0.73/frame (~1095-airing) shortfall, 10× past the bound. "…not loss" explains away ~5.6% of that arm's airings which are real monitor loss (or genuinely fewer airings).

The correct statement is already half-written above: quantization accounts for the 0/3 arms, and the 12-arm shortfall is the "monitor can only lose airings, so the ratio is a floor" caveat in action. As written, this lets the favourable measurement over-generalize — the failure the repo's own doc rules name.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -798,6 +927,27 @@ void Rtl8733bDevice::Stop() {
...
+     * auto-ACKing — an active SIFS-timed transmitter with no session owning
+     * it. The knob is opt-in precisely because it makes the radio transmit;
+     * it must not outlive the session that asked for it.
+     *
+     * Guarded like the TSSI rollback above and for the same reason: this is a
+     * read-modify-write, register reads throw on a disconnected device
+     * (UsbTransport rtw_read), and letting that escape here would abandon the
+     * REST of teardown — _mac.stop() and the card-disable sequence — which is
+     * worse than losing the disarm on a device that has already gone away. */
+    if (_ack_armed) {
```
josephnef: **Altitude: the disarm belongs inside `Halmac8733bMac::stop()`, not as a flag-guarded special case here.** The root cause this comment documents is that `stop()` clears only REG_CR's low half; one net_type RMW of 0x0102[1:0] inside `stop()` makes the disarm unconditional and deletes the `_ack_armed` flag, the latch-before-first-write subtlety in `SetAckResponder`, and this special case. Verified harmless to `rtl8733bprobe` — it never arms net_type, so the clear is a no-op there.

The hazard is real as written: any future path that calls `_mac.stop()` without this flag dance leaves an unowned SIFS-timed auto-ACK transmitter on the air — the exact scenario the comment itself documents. And the exception rationale is hollow: the guarded `ack::disable` is immediately followed by *unguarded* `_mac.stop()`, whose own `rtw_read8` throws on the same disconnected device and abandons the rest of teardown anyway; the flag's cost argument ("a USB round-trip the common path should not pay") is likewise undercut by `stop()`'s existing unguarded RMW.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -755,6 +806,84 @@ size_t Rtl8733bDevice::build_tx_block(const uint8_t *packet, size_t length,
...
+   * Stop(); if the flag were set only on the success path, that teardown would
+   * believe there was nothing to disarm while the chip went on auto-ACKing.
+   * Same reasoning for a failed readback below. A redundant disarm costs one
+   * USB round-trip; a missed one leaves an unowned transmitter on the air. */
+  _ack_armed = true;
+  devourer::ack::enable(_device, mac.data());
+  /* Read back before claiming it. This is the first port of the responder onto
+   * HALMAC 87xx, and this backend does not report a write it cannot verify
+   * (the same standard SetCcaMode and SetTxPowerOffsetQdb are held to). */
+  const uint8_t net_type = static_cast<uint8_t>(_device.rtw_read8(0x0102) & 0x03u);
```
josephnef: **The register recipe now lives in two files.** This readback re-hardcodes `AckResponder.h`'s map a second time — raw 0x0102/0x0610/0x0614 reads plus manual LE `want_lo`/`want_hi` packing mirroring `ack::enable` byte-for-byte. A future `AckResponder.h` change (port-1 variant, different net_type value) silently diverges from this copy: a stale check passes, or a healthy arm fails.

Same altitude problem with the I/G refusal above: it enforces per-backend a precondition `AckResponder.h` states for *all* backends but enforces nowhere. The right home for both is shared — an `ack::verify()` + is-unicast check beside `enable()`/`disable()`. That also fixes Jaguar1/2/3, which stay fire-and-forget and return `true` for a group MAC or an unlatched write — the exact "silently dead responder read as on=0/off=0" harness artifact `AdapterCaps.h` documents from the 8821AU episode.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -76,6 +77,40 @@ void Rtl8733bDevice::bring_up_to_phy() {
...
+  /* ACK window (DEVOURER_ACK_TIMEOUT_US): the one library default every
+   * generation programs identically, replacing the vendor value init_wmac just
+   * wrote (0x21 = 33 us — the bottom of the 33..128 per-chip spread the single
+   * default exists to abolish). Same clamp and same register as jaguar1/2/3;
+   * see the DeviceConfig field doc for the range budget it buys.
+   *
+   * This was silently ignored until issue #2, which is the failure this
+   * backend refuses everywhere else (disable_cca, SetTxPowerIndexOverride): a
+   * config value that reads as applied while the radio runs something else.
+   * The window is load-bearing here now — both ends of a hardware-ARQ link are
```
josephnef: **Convention: no issue-# tags or changelog framing in code comments** — "until issue #2" here, "silently ignored here once (issue #2)" in `Halmac8733bMac.cpp` `set_ack_timeout_us`, "(issue #2)" in the test header; and the same historical framing in the docs ("was silently pinned at 33 us", "is honoured, and was not until that issue", the "(ACK/BlockAck response is no longer in this list…)" tombstone) against the docs-are-current-state-only rule. Git tracks provenance; issue numbers mean nothing without the tracker; and the "until"/"no longer" narratives describe code states that no longer exist. Each comment's rationale half — e.g. "a value reported without a readback is the same shape of claim" — already stands alone; keep that, drop the history.

josephnef (changes requested):
The code core traced clean across the review: the AckResponder register map was verified against the vendor tree (net_type `REG_CR_8733B + 2` = 0x0102, MACID 0x0610, BSSID 0x0618), the arm/teardown exception paths are coherent, the harness's event parsing matches rxdemo's actual emission, and nothing new touches the per-frame send path. The measurements hold up. Requesting changes for one runtime gap (a config-requested arm whose failure is silently discarded — the exact failure shape this PR was filed about), one harness false-verdict hole, and a cluster of doc claims that overrun what was measured. Ten findings total, nine confirmed by an adversarial verify pass; details inline.

One finding has no diff line to anchor to because the file isn't touched by the PR:

**`src/DeviceConfig.h` — the authoritative field docs were not extended.** `tx.retry_limit` still enumerates "the 11ac generations and Kestrel" (omitting the 8733B this PR measures live), and `tx.ack_timeout_us`'s register footnote still lists REG_ACKTO 0x640 only for the 11ac generations and Kestrel, though the 8733B now programs the same 0x640. Root CLAUDE.md names these doc-comments the per-var reference ("do not look for a second copy here") — a caller auditing coverage from them concludes both knobs are inert on the 8733B and budgets the link around the 33 µs vendor value, re-deriving exactly the pre-PR behavior this PR fixed.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Fixed in `466bf5e`. `Init()` now checks the config-requested `SetAckResponder()` result and throws when the arm is refused. Hardware rerun on this head: `57:42:75:05:d6:00` emitted the `UNICAST MAC` error plus `configured ACK responder could not be armed` and exited 134; `02:12:34:56:78:9a` armed normally.

docs/rtl8733b.md
snokvist: Fixed in `466bf5e`. The claim is now explicitly scoped to normal-ACK single-frame response. The BlockAck attempt is recorded as indeterminate: the A-MPDU control did not separate (armed and disarmed both reported 0% delivered / retries 0), because per-frame `tx.report` is not a valid instrument with `AGG_EN`. No measured BA claim remains.

tests/rtl8733b_retry_limit_onair.sh
snokvist: Fixed in `466bf5e`: the harness now refuses `FRAMES < 1000` instead of producing a quantized verdict. The default 1500-frame hardware rerun passed the full `0 3 12 0 12` ladder at `0.93, 3.93, 12.27, 0.93, 12.27` airings/frame with `tx_retry_limit_ok: true`.

docs/scheduled-mac.md
snokvist: Fixed in `466bf5e`. The RTL8733B single-shot table cell is now `unmeasured`; the closed-loop 1736/1736 figure is retained only as a labeled parenthetical, matching the table semantics.

docs/scheduled-mac.md
snokvist: Fixed in `466bf5e`: the stale J1/2/3-only sentence now says the responder is not implemented on the AX generation.

src/rtl8733b/CLAUDE.md
snokvist: Fixed in `466bf5e`. The text now attributes quantization only to the retry 0/3 residues and identifies the retry-12 shortfall as witness-side monitor loss, with the measured ratio explicitly treated as a floor.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Fixed in `466bf5e`. The unconditional net-type clear now lives in `Halmac8733bMac::stop()`; `_ack_armed` and the device-level special case were removed. Hardware teardown rerun with `DEVOURER_TEARDOWN_POWER_DOWN=0` produced zero `ok:true` reports from the subsequent retry-12 probe, confirming no responder state leaked past the session.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Addressed in `466bf5e`: `ack::is_unicast()` and `ack::verify()` now live beside the shared recipe, and the RTL8733B backend no longer duplicates register packing. Jaguar1/2/3 now share the unicast guard. I did not make their existing fire-and-forget arms depend on readback yet: J1 and J3 were hardware-checked, but there is no J2 bench cell, so doing so could turn a healthy J2 arm into a refused arm on an unverified die. I am happy to extend `verify()` per die once that cell is available.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: The changelog framing and the local issue tags were removed in `466bf5e`; the remaining rationale is current-state only. One clarification on the stated convention: the current tree still has 18 `issue #N` references under `src/` (including `CfoTracker.h`, `UsbTransport.cpp`, `UsbOpen.cpp`, `jaguar1/FirmwareManager.cpp`, and `jaguar1/RtlJaguarDevice.cpp`), so a repository-wide no-tag convention is not presently reflected by the tree. I removed these tags because they were unnecessary here; normalizing the existing references would be a separate tree-wide cleanup.

snokvist:
Addressed the unanchored `src/DeviceConfig.h` finding in `466bf5e`: the authoritative docs for both `tx.retry_limit` and `tx.ack_timeout_us` / REG_ACKTO 0x640 now explicitly include RTL8733B.

josephnef (approved):
Re-reviewed at `466bf5e`. All ten findings verified fixed against the actual delta, not the replies:

- **Config-arm discard** — a refused `DEVOURER_ACK_RESPONDER` now throws out of `bring_up_to_phy`, failing the bring-up, and the unwind path is safe because the disarm is now unconditional in `stop()`.
- **Disarm altitude** — `ack::disable` moved into `Halmac8733bMac::stop()`, unconditional; `_ack_armed`, its latch-ordering subtlety and the `Stop()` special case are all deleted. Net −27 lines at the device layer.
- **Duplicated register recipe** — `ack::is_unicast()` / `ack::verify()` now live beside `enable()`/`disable()` in `AckResponder.h`; the 8733B consumes both, and the deliberate scoping of `verify()` (MACID + net_type checked, BSSID written-not-checked, documented why) is the right call. The J1/J2/J3 adoption of the unicast guard — with verify() explicitly deferred pending a per-die bench cell — is exactly the split I'd have asked for, and the 8812AU/8812CU regression check covers it.
- **FRAMES floor** — refuses below 1000 with the quantization rationale in the abort message; the header's error bound is restated at the floor (≤0.099/frame), which is correct.
- **Doc scope** — BlockAck claim retracted to normal-ACK singles in both docs, and the indeterminate `ampdu_ba_check.sh` run is reported as indeterminate with the CCX/AGG_EN explanation rather than either suppressed or spun. The quantization attribution now splits the 0/3 arms (covered by the ≤0.066 bound) from the 12 arm's 0.73/frame residue, attributed to monitor loss and folded into the ratio-is-a-floor caveat. Arithmetic checks out.
- **scheduled-mac table** — single-shot cell now `unmeasured` with the closed-loop figure parenthetical, matching the 8821AU row's convention; Kestrel line corrected.
- **DeviceConfig.h** — both field docs now name the 8733B.
- **Conventions** — issue tags and changelog framing gone from the new comments and docs (the one surviving `issue #` in `DeviceConfig.h` is a pre-existing line this PR never touched); the rationale halves stand alone as promised.

Independently verified on this end: default build + full `ctest` green (54/54) at `466bf5e`.

One non-blocking observation, no action needed: the failed-verify path in `SetAckResponder` no longer attempts an immediate rollback — a mid-session runtime call that verify-fails with net_type latched would leave a half-armed engine until teardown. The comment documents the choice and the scenario needs a silent readback mismatch without a USB exception, so deferring to the unconditional `stop()` disarm is defensible; just noting the window exists.

PR #407, opened 2026-09-01
rtl8733b: harden ARQ and qualify BlockAck response

snokvist:
Follow-up to #406 after its merge.

## Summary

- Make ACK-responder arm/disarm fail closed: check register writes, close the gate before retargeting, verify RTL8733B rollback/disarm, and keep teardown progressing after a failed safety clear.
- Propagate construction-time responder refusal on Jaguar1/2/3 instead of green-initializing a passive monitor.
- Apply `tx.ack_timeout_us` to both RTL8733B response-window registers (`REG_ACKTO` and `REG_ACKTO_CCK`) and verify both readbacks.
- Replace sampled/ambient-prone retry evidence with run-unique payload-counter witness ledgers; add an independent RTL8733B soliciting-TX ACK-closure test.
- Add a CCX-independent three-radio BlockAck test. The witness proves A-MPDU structure and retry-copy closure, and directly decodes addressed `0x94` compressed BlockAck frames, including the bitmap.
- Keep CCX wording scoped: no H2C/MACID registration transport is added here, and the absence of reports is not assigned to firmware.

## Hardware evidence

Current rig:

- TX: `0bda:b812`, RTL8822B/88x2BU, Jaguar2
- RTL8733B DUT: `0bda:f72b`
- independent witness: `0bda:8812`, RTL8812AU/Jaguar1

BlockAck, ch36/MCS3/retry 12 (`tests/rtl8733b_blockack_onair.sh`):

- active but unarmed RTL8733B: 1,605 unique aggregated payloads, 12.720 witnessed copies/payload, `paggr=0.665`, max burst 9, **0** matching BlockAcks;
- armed RTL8733B: 128,702 unique aggregated payloads, 1.001 copies/payload, `paggr=1.0`, max burst 9, **14,402** addressed compressed BlockAcks; all 14,402 carried nonzero bitmaps;
- the harness checks responder and witness liveness before and after TX and aborts unless the preparatory disarm and shutdown are verified.

Other reruns:

- RTL8733B soliciting normal-ACK TX, MCS3: responder on/off = 1.016/12.879 copies per observed payload, with 99.8%/100% witness coverage.
- Retry dose response `0/3/12/0/12`: 1.00/4.00/12.34/1.00/12.34 copies per frame; verdict true.
- J2 construction-time group responder: nonzero initialization failure with the unicast diagnostic; valid unicast still arms. After valid-process exit, a J1 probe measured 0/1,311 ACK successes and retries pinned at 12, so no teardown leak was inferred.
- 11M CCK current J2-witness rerun separated on/off at 1.004/12.113 copies, but its one-shot on arm captured 55.7%, below the harness's fixed 60% coverage floor; this rerun is supportive, not counted as a new formal pass. The documented qualifying CCK cell used the prior J3 witness.

## Checks

- Default build succeeds.
- `ctest --test-dir build --output-on-failure`: 54/54 pass.
- `bash -n` and `shellcheck` pass for all three RTL8733B on-air ARQ harnesses.
- `git diff --check` passes.

## Scope

This qualifies RTL8733B as a **BlockAck responder** only on the measured Jaguar2-to-RTL8733B combination. RTL8733B A-MPDU TX remains unported and unmeasured. CCX/`tx.report` remains unimplemented pending a separately tested H2C queue plus `MEDIA_STATUS_RPT` registration path.


src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -98,14 +99,18 @@ void Rtl8733bDevice::bring_up_to_phy() {
...
+      ackto.cck != ackto_want) {
+    _logger->error(
+        "RTL8733B: ACK window did not latch — REG_ACKTO={} us "
+        "REG_ACKTO_CCK={} us, wanted {} us",
+        ackto.non_cck, ackto.cck, ackto_want);
+    throw std::runtime_error("RTL8733B: configured ACK window did not latch");
+  }
+  _logger->info(
+      "RTL8733B: ACK window {} us (REG_ACKTO 0x640 + CCK 0x639, verified)",
+      ackto_want);
```
josephnef, line 117: `bring_up_to_phy` throws on `!ackto.writes_ok` even when both ACK-window readbacks equal the wanted value — contradicting this PR's own readback-is-the-authority rule (`AckResponder.h`'s `disable_verified` comment calls a failed-status-but-landed write realistic). A transient control-transfer status error becomes a hard bring-up abort: `set_ack_timeout_us` returns `writes_ok=false` with `non_cck==cck==wanted`, and a session with verified-correct register state refuses to start.

The same warn→throw escalation also applies to pure monitor-RX bring-up, where the ACK window doesn't matter — and the claim that `REG_ACKTO_CCK` 0x639 always latches rests on one f72b unit. Pre-PR behavior was warn-and-continue. Suggest: throw only when the *readback* disagrees; treat write status as diagnostic.

src/jaguar1/RtlJaguarDevice.cpp
```diff
@@ -777,7 +782,10 @@ bool RtlJaguarDevice::SetAckResponder(const devourer::MacAddr &mac) {
 }
 
 void RtlJaguarDevice::ClearAckResponder() {
-  devourer::ack::disable(_device);
+  if (!devourer::ack::disable(_device)) {
```
josephnef: `ClearAckResponder` now throws `std::runtime_error` from a void, teardown-shaped API on a raw write status — on all four backends (also `RtlJaguar2Device.cpp:394`, `RtlJaguar3Device.cpp:2185`, `Rtl8733bDevice.cpp:912`) — while this same PR treats the identical failure as swallow-and-warn in `Halmac8733bMac::stop()` because "teardown is best-effort after disconnect".

A consumer disarming the responder during shutdown after a USB error/unplug (the common reason the 0x0102 write fails) now gets a throw; in a destructor or unwind path that's `std::terminate`. And per `AckResponder.h`'s own comment, a failed write status doesn't prove the write missed — so the throw can fire on a successfully closed gate.

src/AckResponder.h
```diff
@@ -40,23 +40,50 @@
...
-                                      ((uint32_t)mac[2] << 16) |
-                                      ((uint32_t)mac[3] << 24));
-  dev.rtw_write16(0x061c, (uint16_t)(mac[4] | (mac[5] << 8)));
+inline bool enable(RtlAdapter &dev, const uint8_t mac[6]) {
   const uint8_t nt = dev.rtw_read8(0x0102);
-  dev.rtw_write8(0x0102, static_cast<uint8_t>((nt & ~0x03u) | 0x03u));
+  /* Close the gate before changing identity. Besides avoiding a transient
+   * responder for a half-written MAC during retargeting, this makes every
+   * failed identity write leave the radio passive. */
+  if (!dev.rtw_write8(0x0102, static_cast<uint8_t>(nt & ~0x03u)))
```
josephnef: Fail-passive has a hole on re-arm: if `enable()`'s first write (the gate close, `rtw_write8(0x0102, nt & ~0x03)`) fails during a retarget, it returns false having changed *nothing* — gate still open, MACID still the old MAC. The radio keeps SIFS-ACKing MAC A while `SetAckResponder(B)` reports failure, and the comment claims every failure leaves the radio passive. On Jaguar2 (no teardown power-down) the stale responder answers until re-enumeration.

Only `Rtl8733bDevice` attempts `disable_verified` rollback on `enable()==false`; the three Jaguar callers just log "not armed". Either roll back in `enable()` itself or in all four callers.

src/AckResponder.h
```diff
@@ -40,23 +40,50 @@
...
-  dev.rtw_write<uint32_t>(0x0610, (uint32_t)mac[0] | ((uint32_t)mac[1] << 8) |
-                                      ((uint32_t)mac[2] << 16) |
-                                      ((uint32_t)mac[3] << 24));
-  dev.rtw_write16(0x0614, (uint16_t)(mac[4] | (mac[5] << 8)));
-  dev.rtw_write<uint32_t>(0x0618, (uint32_t)mac[0] | ((uint32_t)mac[1] << 8) |
-                                      ((uint32_t)mac[2] << 16) |
-                                      ((uint32_t)mac[3] << 24));
-  dev.rtw_write16(0x061c, (uint16_t)(mac[4] | (mac[5] << 8)));
+inline bool enable(RtlAdapter &dev, const uint8_t mac[6]) {
   const uint8_t nt = dev.rtw_read8(0x0102);
```
josephnef: The bool error contract only covers writes. `enable()`/`disable()`/`is_disabled()` all begin with `rtw_read8(0x0102)`, which *throws* `std::ios_base::failure` on a failed transfer (`UsbTransport.h:167`) — read failures bypass the return-false/rollback/log paths entirely. A timed-out 0x0102 read inside `ack::enable` propagates uncaught out of `SetAckResponder` (and out of `Init`/`InitWrite` as a different exception type than the intended `runtime_error`).

Likewise `(void)disable(dev)` inside `disable_verified` is not actually best-effort — its internal read can throw before `is_disabled` runs — and `is_disabled`'s "a control transfer may report failure" comment is unreachable for reads: they never report, they throw.

tests/rtl8733b_arq_tx_onair.sh
```diff
@@ -0,0 +1,149 @@
...
+
+python3 - "$RESULTS" "$RETRY_LIMIT" "$RATE" <<'PY'
+import json, sys
+rows = {r["phase"]: r for r in map(json.loads, open(sys.argv[1]))}
+limit = int(sys.argv[2])
+rate = sys.argv[3]
+on, off = rows["on"], rows["off"]
+on_c = on["copies_per_observed_frame"]
+off_c = off["copies_per_observed_frame"]
+ok = (on["coverage"] >= 0.60 and off["coverage"] >= 0.60 and
```
josephnef, line 168: The verdict band is arithmetically unsatisfiable at the guard's own floor: the guard admits `RETRY_LIMIT` 3..63, but the verdict requires both `off_c <= 1.15*(limit+1)` and `off_c >= 5*on_c`. At `RETRY_LIMIT=3`, perfect hardware gives `on_c >= 1.0` (copies per observed frame is ≥1 by construction), so the verdict demands `off_c >= 5.0` while the band caps it at 4.6 — contradiction, exit 1 on every healthy run (and it's near-unsatisfiable at 4). The sibling `rtl8733b_blockack_onair.sh` floors `RETRY_LIMIT` at 5 for exactly this reason — floor this guard the same way (or derive the floor from the band arithmetic).

tests/rtl8733b_arq_tx_onair.sh
```diff
@@ -0,0 +1,149 @@
...
+print(json.dumps({"ev": "rtl8733b.arq_tx", "phase": phase,
+                  "sent": sent, "observed_frames": len(copies),
+                  "coverage": round(len(copies) / sent, 3),
+                  "copies": total,
+                  "copies_per_observed_frame": round(total / len(copies), 3)}))
+PY
+  tail -1 "$RESULTS"
+}
+
+run_phase on 1
```
josephnef, line 157: The responder-ON phase runs before the OFF control arm and teardown is `pkill -9`, which skips `Stop()` — Jaguar1's responder disarm lives only in `Stop()`/the destructor, and its own comment says the chip "stays in ACT with its RF front end live" after the owner dies. So the 8812AU's hardware responder (net_type=AP + MACID in silicon) can stay armed into the OFF arm: if the orphaned MAC still ACKs `RESP_MAC`, `off_c` collapses toward 1 and the run FAILs on healthy hardware. The sibling blockack script added `force_responder_passive()` and runs off-before-on for exactly this SIGKILL hazard; this script needs both.

tests/rtl8733b_retry_limit_onair.sh
```diff
@@ -81,18 +114,19 @@ for arm in $ARMS; do
...
-  sudo env DEVOURER_VID=$DUT_VID DEVOURER_PID=$DUT_PID DEVOURER_CHANNEL=$CH \
-       DEVOURER_TX_QOS_DATA=1 DEVOURER_TX_RA=$RA \
-       DEVOURER_TX_RATE=$RATE DEVOURER_TX_PAYLOAD_BYTES=200 \
-       DEVOURER_TX_GAP_US=$GAP_US DEVOURER_TX_FRAMES=$FRAMES \
-       DEVOURER_TX_RETRY_LIMIT=$arm DEVOURER_TX_PWR_OFFSET_QDB=$PWR_QDB \
+  # shellcheck disable=SC2024
+  sudo env DEVOURER_VID="$DUT_VID" DEVOURER_PID="$DUT_PID" \
+       DEVOURER_CHANNEL="$CH" DEVOURER_TX_QOS_DATA=1 \
+       DEVOURER_TX_RA="$RA" DEVOURER_TX_SA="$TX_SA" \
+       DEVOURER_TX_RATE="$RATE" DEVOURER_TX_PAYLOAD_BYTES=200 \
```
josephnef, line 123: Airings are counted with `grep -c` on `rx.seq` with no CRC filter, while both sibling harnesses in this PR discard crc-flagged copies — and the documented witness for these cells (RTL8812CU, Jaguar3) keeps FCS-error frames unconditionally (ACRC32|AICV set in monitor bring-up). A marginal-RF arm delivering crc-error duplicates inflates `airings_per_frame` past the `1.15*expected` ceiling and FAILs the dose-response on healthy hardware (or masks a real shortfall at the floor). The three harnesses silently measure different quantities while `src/rtl8733b/CLAUDE.md` describes them as one method — filter crc here like the siblings do.

tests/rtl8733b_blockack_onair.sh
```diff
@@ -0,0 +1,306 @@
...
+  exit 2
+fi
+if [ "$TX_VID:$TX_PID" = "$RESP_VID:$RESP_PID" ] ||
+   [ "$TX_VID:$TX_PID" = "$WIT_VID:$WIT_PID" ] ||
+   [ "$RESP_VID:$RESP_PID" = "$WIT_VID:$WIT_PID" ]; then
+  echo "ABORT: TX, responder, and witness must be distinct adapters" >&2
+  exit 2
+fi
+
+ESC_BUILD=$(printf '%s' "$BUILD" | sed 's/[][\.*^$/]/\\&/g')
```
josephnef: `ESC_BUILD`'s sed class `[][\.*^$/]` escapes sed/BRE metachars but not the ERE metachars `+ ? ( ) { } |` that `pkill -f` matches with. A BUILD path containing any of them (e.g. `/home/user/devourer-g++/build`: the `+` survives unescaped, anchored pattern never matches — verified with `grep -E`) makes `cleanup()`/`stop_receivers()` silently kill nothing, stderr discarded — the previous arm's receiver keeps the adapter claimed into the next arm, which is the cross-arm contamination this harness rewrite exists to prevent. Same construct in `tests/rtl8733b_arq_tx_onair.sh:33` and `tests/rtl8733b_retry_limit_onair.sh`; either escape the ERE set too or match on `pkill -f -x`-safe fixed strings.

src/rtl8733b/CLAUDE.md
```diff
@@ -322,42 +321,53 @@ session ends with `teardown_power_down` off, the peer reads ack_rate 0.00 with
...
-delivered at retries 0, so the control arm did not separate — which is the
-documented consequence of per-frame CCX accounting not surviving AGG_EN
-(`docs/aggregation.md`), not a verdict on this chip. A-MPDU is unported here
-anyway. So: normal-ACK response measured; BlockAck response untested, pending
-an A-MPDU-capable instrument that does not judge by `tx.report`.
+8733B as RESPONDER): armed 1725/1725 reports ACKed at retries_mean 0.00;
+re-armed on a **different** MAC, 1728/1728 again (the address is arbitrary, not
+baked in); disarmed, 0/1723 successful reports with retries pinned at 12.
+
+The other direction is now measured independently rather than inferred.
```
josephnef: Historical/changelog framing in current-state docs: "is now measured independently rather than inferred" (here), "BlockAck response is now separately measured" / "The replacement tests/rtl8733b_blockack_onair.sh" (:337, :341), "The corrected harness uses..." (:359), and `docs/rtl8733b.md:399` "avoiding the old sampled/quantized rx.txhit readout" — each references a deleted past state a future reader must reconstruct from git. Repo convention is current-state only (git is the changelog): "is measured independently by tests/..." carries the same facts without the narration.

josephnef (changes requested):
Reviewed the ARQ hardening + BlockAck qualification. The goal is right — no silently dead responder, a verified ACK window — but the hardening contradicts its own stated contracts in three places (throw-on-write-status vs readback-is-authority, fail-passive with a stale-responder hole, a bool error contract the read path bypasses), and the three on-air harnesses have already drifted apart from each other inside this one PR. Details inline; one finding is outside the diff:

**`examples/chanmig/main.cpp:576` — the chanmig drone discards `SetAckResponder`'s bool.** With this PR's new deliberate fail-passive paths (`enable()` now closes the gate and returns false on a failed write), a single transient control-transfer failure at drone bring-up leaves the radio passive by design while chanmig proceeds anyway. The ground side's hardware-ARQ frames all retry to the limit against a silent responder and the operator debugs the RF link — the exact silent-dead-responder scenario the 8733B comments say this hardening exists to prevent, unhandled at the one in-tree runtime call site.

snokvist:
Addressed the requested reliability findings in commit 4c2b0f7:

- ARQ A/B receiver teardown now uses graceful SIGINT and rejects a phase if responder disarm was not verified; SIGKILL remains fallback cleanup.
- Jaguar1/2/3 arm failures now perform verified rollback and throw if passive state cannot be proven; disarm uses readback authority.
- BlockAck phase teardown checks stop-time disarm warnings.
- BlockAck verdict now requires compressed-and-nonzero bitmap on the same addressed BA event.
- chanmig now aborts drone bring-up when SetAckResponder refuses the arm.

Local build and all 54 ctest tests pass; shellcheck/bash -n and git diff --check also pass.

snokvist:
@josephnef All requested findings are addressed in 4c2b0f7 (and Qodo now reports 0 unresolved issues). Please re-review the updated head when convenient.

src/rtl8733b/Rtl8733bDevice.cpp
josephnef: Still open in 4c2b0f7 — `bring_up_to_phy` is unchanged: `!ackto.writes_ok ||` still fails the bring-up even when both readbacks equal `ackto_want`. The condition should be readback-only (`ackto.non_cck != ackto_want || ackto.cck != ackto_want`), with `writes_ok` at most a diagnostic.

src/jaguar1/RtlJaguarDevice.cpp
josephnef: Partially addressed in 4c2b0f7: switching to `disable_verified` fixes the readback-authority half — the throw can no longer fire on a successfully closed gate. But the other half stands: `ClearAckResponder` still throws from a void teardown-shaped API while `Halmac8733bMac::stop()` swallow-and-warns the identical failure, and after an unplug it's now the *read* inside `disable_verified` that throws — `std::ios_base::failure`, not even the intended `runtime_error` — in a destructor/unwind path that's `std::terminate`.

src/AckResponder.h
josephnef: Still open in 4c2b0f7 — the `try/catch` added around the rollback covers only that one call. `ack::enable`'s first `rtw_read8(0x0102)` still throws `ios_base::failure` out of `SetAckResponder` uncaught on all four backends, and `ClearAckResponder`'s `disable_verified` read does the same (see the teardown thread). The bool contract still doesn't cover reads.

tests/rtl8733b_arq_tx_onair.sh
josephnef: Still open in 4c2b0f7 — the guard still admits `RETRY_LIMIT` 3..63 and the verdict still requires `off_c <= 1.15*(limit+1)` AND `off_c >= 5*on_c`, which is unsatisfiable at 3 (band caps off_c at 4.6, `5*on_c >= 5.0`). Floor at 5 like the blockack sibling.

tests/rtl8733b_retry_limit_onair.sh
josephnef: Still open — `tests/rtl8733b_retry_limit_onair.sh` is untouched in 4c2b0f7: airings are still counted with an unfiltered `grep -c` on `rx.seq` while both sibling harnesses discard crc-flagged copies and the 8812CU witness keeps FCS-error frames.

tests/rtl8733b_blockack_onair.sh
josephnef: Still open in 4c2b0f7 — `ESC_BUILD` still escapes only the BRE set, and the new `stop_receivers()` uses the same `"^$ESC_BUILD/rxdemo"` pattern, so the graceful-SIGINT path inherits the bug: on a path containing `+ ? ( ) { } |` the SIGINT pkill silently matches nothing, the 5s wait expires, and every phase aborts (or worse, the pgrep also misses and the phase 'passes' with the receiver still running). Now load-bearing for the disarm-verification fix, so worth fixing in all three scripts.

src/rtl8733b/CLAUDE.md
josephnef: Still open — neither `src/rtl8733b/CLAUDE.md` nor `docs/rtl8733b.md` changed in 4c2b0f7; the historical framing ("is now measured", "The replacement", "The corrected harness", "avoiding the old ... readout") is all still there.

src/AckResponder.h
josephnef: Resolved in 4c2b0f7 — callers now roll back with `disable_verified` and refuse to report unknown state as passive. (The best-effort retry write in `enable()` is harmless; the caller-side verification is what closes the hole.)

tests/rtl8733b_arq_tx_onair.sh
josephnef: Resolved in 4c2b0f7 — graceful SIGINT teardown plus the verified-disarm gate before the next phase closes the orphaned-responder hazard even with ON-before-OFF ordering. (Note the new `stop_receivers()` inherits the ESC_BUILD escaping bug — see that thread.)

josephnef:
Re-reviewed at 4c2b0f7 — thanks, the reliability core is in good shape now: the stale-active-responder hole (verified rollback), the harness SIGKILL/orphan-responder hazard (graceful SIGINT + verified-disarm gate), and the chanmig arm check are all resolved, along with the Qodo items.

But "all requested findings" isn't accurate — **6 of my findings are untouched and 1 is half-fixed** (details in the threads):

- `Rtl8733bDevice.cpp` `bring_up_to_phy` still throws on `!writes_ok` even when both readbacks match — the condition should be readback-only.
- `ClearAckResponder` (half-fixed): readback authority ✓, but it still throws from a void teardown-shaped API — and post-unplug the internal read now throws `ios_base::failure` in a destructor path (`std::terminate`), while `Halmac8733bMac::stop()` swallow-and-warns the same failure.
- The bool error contract still doesn't cover reads: `ack::enable`'s first `rtw_read8(0x0102)` throws uncaught out of `SetAckResponder` on all four backends.
- `rtl8733b_arq_tx_onair.sh`: verdict still arithmetically unsatisfiable at the guard's own `RETRY_LIMIT=3` floor — floor at 5 like the blockack sibling.
- `rtl8733b_retry_limit_onair.sh`: untouched — still counts airings with no CRC filter against a witness that keeps FCS-error frames.
- `ESC_BUILD` BRE-vs-ERE escaping: untouched and now load-bearing — the new `stop_receivers()` uses the same broken pattern, so on an ERE-metachar path the graceful path silently matches nothing.
- Docs: the historical framing in `src/rtl8733b/CLAUDE.md` / `docs/rtl8733b.md` is all still there (repo convention is current-state only).

Keeping changes-requested until these are in.

tests/rtl8733b_arq_tx_onair.sh
snokvist: Addressed in b039144. rtl8733b_arq_tx_onair.sh now requires RETRY_LIMIT=5..63, matching the BlockAck harness and making the 5x separation constraint satisfiable.

src/jaguar1/RtlJaguarDevice.cpp
snokvist: Addressed in b039144. disable_verified is now noexcept and returns false on read/write exceptions. All four ClearAckResponder implementations log and return rather than throwing from the void teardown-shaped API.

src/AckResponder.h
snokvist: Addressed in b039144. The shared enable, verify, and disable_verified helpers now contain transport read exceptions within their bool contracts. Failed arms attempt verified rollback; callers return false and explicitly log UNKNOWN hardware state when rollback cannot be proven.

tests/rtl8733b_retry_limit_onair.sh
snokvist: Addressed in b039144. rtl8733b_retry_limit_onair.sh now JSON-parses rx.seq and counts only crc=false events, matching both sibling harnesses. Recomputing the retained qualifying logs produced the same ladder because those logs contained zero CRC-bad rx.seq events.

src/rtl8733b/CLAUDE.md
snokvist: Addressed in b039144. The ARQ sections in docs/rtl8733b.md and src/rtl8733b/CLAUDE.md now describe current measurement contracts directly; the original/replacement/corrected/old-readout historical framing has been removed.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Addressed in b039144. Bring-up now fails only when either ACK-timeout readback differs from the requested value. A failed USB write status with matching readbacks emits a warning but does not override hardware state.

tests/rtl8733b_blockack_onair.sh
snokvist: Addressed in b039144 across all three RTL8733B ARQ scripts. ESC_BUILD now escapes the complete ERE metacharacter set, including + ? ( ) { } and |; shellcheck/bash -n and a metacharacter-path match/non-match check pass.

snokvist:
@josephnef The seven still-open findings from your latest pass are addressed in b039144, with per-thread replies and local verification. Please re-review the updated head when convenient.

josephnef (approved):
Re-reviewed at b039144 — all 7 outstanding findings are verified fixed, independently confirmed against the diff plus a local build (54/54 ctest, `bash -n` on all three harnesses):

- `bring_up_to_phy` now throws on readback mismatch only; a failed transport status with matching readbacks warns. ✓
- `ClearAckResponder` is non-throwing on all four backends, `disable_verified`/`enable`/`verify` are `noexcept` with the read path contained in the bool contract, and the `IRtlDevice` doc states the false-is-not-proof-of-passive semantics. ✓
- `RETRY_LIMIT` floor is 5 (band 3.6–6.9 vs the ≥5.0 separation — satisfiable, matching the sibling). ✓
- `rtl8733b_retry_limit_onair.sh` now counts only crc-clean `rx.seq` events, matching the siblings — and thanks for recomputing the retained qualifying logs. ✓
- `ESC_BUILD` escapes the full ERE set in all three scripts (checked the produced pattern against a `+ ( ) |` path with `grep -E` — matches). ✓
- The historical framing is gone from the ARQ sections of both docs. ✓

Approving.

PR #409, opened 2026-09-03
rx: gate PHY-status parsing on the RX-descriptor PHYST bit (Jaguar1/2/3)

gilankpam:
## Problem

`REG_RX_DRVINFO_SZ` is a **global** register on every generation that carries per-frame drvinfo — Jaguar1 `_InitDriverInfoSize_8812A(4)` (`HalModule.cpp:268`), Jaguar2 `0x060F=4` (`HalJaguar2.cpp:2578`), Jaguar3 `0x060F=4` (`HalJaguar3.cpp:485`). The drvinfo space is therefore reserved on **every** frame, while the PHY writes a report only into frames whose RX-descriptor **PHYST** bit (DW0 bit 26) is set.

All three RX loops gated the phy-status parse on the drvinfo *size* alone. On all-but-one subframe of an A-MPDU the area holds bytes left by an earlier frame, and the page nibble can alias a valid page number, so the parse succeeded on garbage and folded it into the RF running averages behind `GetRxQuality()`, `GetActiveRxPaths()` and the `rx.path` event — contaminated RSSI/SNR/EVM tails that track A-MPDU density rather than the link.

On **Jaguar2** it went further than telemetry: the garbage `cfo_tail` fed `_cfo.add()`, which steers the closed-loop XtalCap crystal trim.

## Change

**The gate, on all three generations.** Decode the descriptor PHYST bit and require it before parsing.

| Generation | Site | Was |
|---|---|---|
| Jaguar1 | `FrameParser.cpp:237` | bit decoded at line 88, memcpy gated on `drvinfo_sz` only — the comment there already named `pattrib->physt` as the vendor's gate it was approximating |
| Jaguar2 | `RtlJaguar2Device.cpp:622` | `GET_RX_DESC_PHYST_8822B` present but **unused**; bit never decoded into `Rx8822bFrame` |
| Jaguar3 | `RtlJaguar3Device.cpp:330` | bit never decoded |

**`RxAtrib.physt` gets one cross-generation meaning**, documented at its declaration (`src/RxPacket.h`): the raw descriptor bit — "the PHY wrote a report into this frame's drvinfo" — explicitly *not* "the report parsed" and *not* "the signal fields are valid". Kestrel never sets it, and the comment says so and why (its PHY status arrives as its own PPDU-status frame). This resolves the open question the first revision raised: the parse outcome lives in a local instead.

**`parse_phy_sts_jgr2` / `parse_phy_sts_jgr3` return `PhyStsFill` (`None` / `Power` / `Full`)** — *which fields they filled*, not whether they liked the buffer. A bool could not express the two real tiers, and was wrong in both directions:

- it returned **false for OFDM pages 2..6** *after* filling valid per-path RSSI, `ldpc`, `stbc` and `bw` from the common header those pages share with type1 — dropping a real measurement, and leaving `GetRxQuality`/`GetActiveRxPaths` to freeze with no diagnostic if the BB page selector ever left type1 (devourer's BB table pins `0x8C0[25:22]=1`, but the vendor auto-switch and the debug page helper do not restore it)
- it returned **true for the CCK page**, whose layout carries path-A power and no SNR/EVM/CFO at all

The CFO trackers now take only a `Full` fill on Jaguar2 and Jaguar3.

**`RxQualityAccumulator` folded SNR unconditionally** (`src/RxQuality.h`) while **its own doc comment three lines above** claimed it folded only when present. EVM and the noise-floor term did filter; SNR did not. Not a Jaguar3 bug — Jaguar1, Jaguar2 and the RTL8733B all feed CCK frames through the same accumulator, so `snr_min_db` has been pinned to 0 on any 2.4 GHz channel with CCK traffic on all of them, and `snr_mean_db` dragged toward zero in proportion to CCK density. SNR now carries its own sample count like EVM, surfaced as `RxQuality::snr_valid` so a CCK-only window reports "no SNR" rather than a fabricated 0. `RxPathActivityAccumulator` already filtered per-metric and is unchanged.

**Stale comment dropped** (`RtlJaguar3Device.cpp`): it claimed CCK frames have short drvinfo and carry no report. Both false — the register write is global, and `parse_phy_sts_jgr3` decodes the CCK page-0 report explicitly.

## Testing

- **`tests/rx_physt_selftest.cpp`** (ctest `rx_physt_bit`) exercises both the bit decode and the return contract, on both generations: descriptor bit set/clear through `parse_rx_8822c` and `parse_rx_8822b`; null buffer and a 27-byte report → `None` with the attrib verified untouched; CCK page → `Power` with `snr`/`evm`/`cfo_tail` asserted still 0; OFDM page 1 → `Full` with rssi/snr/evm/cfo/ldpc/bw checked; OFDM page 5 → `Power` with common-header RSSI present and the type1-only fields still 0.
- `rx_quality_selftest` gained the mixed-window and CCK-only-window SNR cases.
- `ctest`: **55 passed, 1 skipped** of 56 (`la_csi_math` skips, needs numpy), 0 failures — rebased onto `master` at 1413b14.
- Per-chip configs built clean: `JAGUAR1`-only, `JAGUAR2_8822B`-only, `JAGUAR3_8822C`-only. The selftest is registered under `DEVOURER_JAGUAR3 OR DEVOURER_JAGUAR2_8822B OR DEVOURER_JAGUAR2_8821C` — verified absent in the Jaguar1-only config and passing in the other two.
- No new compiler warnings.
- **Hardware**: the contamination was found on an 8822EU RX session; the bit-level decode and the parse tiers are what the selftests pin. Not re-validated on-air since the split from the working branch, hence **draft**.

## Deliberately not included

- **Demo-side consumers** (`examples/rx/main.cpp`): the hopset per-slot accumulators, the `g_rxagg` energy/LinkHealth aggregate and the `DEVOURER_RX_ALLPATHS` `rx.path` rows read `rssi[0]/snr[0]/evm[0]` unconditionally. One correction to how this was framed in review: gating them on `RxAtrib.physt` would drop **every** sample on Kestrel, which never sets it. The generation-neutral gate is `rssi[0] > 0` — the same test `RxQualityAccumulator` already applies. Follow-up.
- **`rx.txhit` emitting `physt`** so an encoding-truth-table run can tell a real BCC fallback from a subframe that carried no report. Follow-up.
- An unrelated `usb.rx_zerocopy` default flip that rode the same working branch.

https://claude.ai/code/session_01JiAS32956Z3cmTbnmcXYkT

### DIFF
--- a/src/jaguar3/FrameParserJaguar3.h
+++ b/src/jaguar3/FrameParserJaguar3.h
@@ -217,6 +217,11 @@ struct Rx8822cFrame {
   uint8_t shift;         /* SHIFT_SZ */
   uint32_t tsfl;         /* hardware TSF-low at receive */
   bool paggr;            /* MPDU arrived inside an A-MPDU */
+  bool physt;            /* a PHY-status report was written for THIS frame;
+                          * drvinfo space is reserved on every frame
+                          * (RX_DRVINFO_SZ is global), so without this bit the
+                          * area holds stale bytes — notably on all-but-one
+                          * subframe of an A-MPDU */
   uint8_t ppdu_cnt;      /* 2-bit received-PPDU counter */
   uint32_t next_offset;  /* 8-byte-aligned offset of the next frame in an agg */
 };
@@ -239,6 +244,7 @@ inline bool parse_rx_8822c(const uint8_t *buf, size_t buflen,
   out.rx_rate = static_cast<uint8_t>(GET_RX_DESC_RX_RATE_8822C(buf));
   out.tsfl = static_cast<uint32_t>(GET_RX_DESC_TSFL_8822C(buf));
   out.paggr = GET_RX_DESC_PAGGR_8822C(buf) != 0;
+  out.physt = GET_RX_DESC_PHYST_8822C(buf) != 0;
   out.ppdu_cnt = static_cast<uint8_t>(GET_RX_DESC_PPDU_CNT_8822C(buf));
 
   uint32_t frame_off =
@@ -269,16 +275,19 @@ inline bool parse_rx_8822c(const uint8_t *buf, size_t buflen,
  * vendor's s(8,1) fields). The page type is taken from byte0 low nibble
  * (page_num) rather than guessed from the rate: 0 = CCK type0, else an OFDM
  * page; per-stream EVM/SNR are only present on the type1 OFDM page.
- * Requires physts_len >= 28. */
-inline void parse_phy_sts_jgr3(const uint8_t *physts, uint16_t physts_len,
+ * Requires physts_len >= 28. Returns true iff physts pointed at a page
+ * layout this function actually understands (page 0 CCK or page 1 OFDM
+ * type1) and `a` was filled from it; false on a null/short buffer or any
+ * other page number (caller should not trust `a`'s signal fields then). */
+inline bool parse_phy_sts_jgr3(const uint8_t *physts, uint16_t physts_len,
                                rx_pkt_attrib &a) {
   if (physts == nullptr || physts_len < 28)
-    return;
+    return false;
   const uint8_t page_num = physts[0] & 0x0f;
   if (page_num == 0) {
     /* type0 (CCK): DW0 = page_num(0), pwdb_a(1). Single path-A power. */
     a.rssi[0] = physts[1];
-    return;
+    return true;
   }
   /* OFDM header (valid for every jgr3 OFDM page): per-path pwdb[4] at bytes
    * 1..4, DW1 byte5 l_rxsc[3:0]/ht_rxsc[7:4], DW1 byte7 flags. */
@@ -303,7 +312,9 @@ inline void parse_phy_sts_jgr3(const uint8_t *physts, uint16_t physts_len,
       a.evm[i] = static_cast<int8_t>(physts[16 + i]);
       a.snr[i] = static_cast<int8_t>(physts[24 + i]);
     }
+    return true;
   }
+  return false;
 }
 
 } /* namespace jaguar3 */
--- a/src/jaguar3/FrameParserJaguar3.h
+++ b/src/jaguar3/FrameParserJaguar3.h
@@ -275,19 +275,21 @@ inline bool parse_rx_8822c(const uint8_t *buf, size_t buflen,
  * vendor's s(8,1) fields). The page type is taken from byte0 low nibble
  * (page_num) rather than guessed from the rate: 0 = CCK type0, else an OFDM
  * page; per-stream EVM/SNR are only present on the type1 OFDM page.
- * Requires physts_len >= 28. Returns true iff physts pointed at a page
- * layout this function actually understands (page 0 CCK or page 1 OFDM
- * type1) and `a` was filled from it; false on a null/short buffer or any
- * other page number (caller should not trust `a`'s signal fields then). */
-inline bool parse_phy_sts_jgr3(const uint8_t *physts, uint16_t physts_len,
-                               rx_pkt_attrib &a) {
+ * Requires physts_len >= 28. Returns which fields of `a` were filled
+ * (PhyStsFill): None on a null/short buffer, Full only on the type1 OFDM page
+ * that carries EVM/SNR/CFO, Power on the CCK page and on every other OFDM page
+ * — those share the common header, so their per-path power (and ldpc/stbc/bw)
+ * IS a measurement even though the type1-only fields are left at 0. Callers
+ * must not fold a field the return value does not claim. */
+inline PhyStsFill parse_phy_sts_jgr3(const uint8_t *physts, uint16_t physts_len,
+                                     rx_pkt_attrib &a) {
   if (physts == nullptr || physts_len < 28)
-    return false;
+    return PhyStsFill::None;
   const uint8_t page_num = physts[0] & 0x0f;
   if (page_num == 0) {
     /* type0 (CCK): DW0 = page_num(0), pwdb_a(1). Single path-A power. */
     a.rssi[0] = physts[1];
-    return true;
+    return PhyStsFill::Power;
   }
   /* OFDM header (valid for every jgr3 OFDM page): per-path pwdb[4] at bytes
    * 1..4, DW1 byte5 l_rxsc[3:0]/ht_rxsc[7:4], DW1 byte7 flags. */
@@ -312,9 +314,15 @@ inline bool parse_phy_sts_jgr3(const uint8_t *physts, uint16_t physts_len,
       a.evm[i] = static_cast<int8_t>(physts[16 + i]);
       a.snr[i] = static_cast<int8_t>(physts[24 + i]);
     }
-    return true;
+    return PhyStsFill::Full;
   }
-  return false;
+  /* Pages 2..6: the common header above was parsed and is valid, so this is a
+   * Power fill rather than a failure — reporting None here would throw away
+   * real per-path RSSI, and would silently freeze GetRxQuality/GetActiveRxPaths
+   * if the BB page selector ever left type1 (devourer's BB table pins
+   * 0x8C0[25:22]=1, but the vendor auto-switch and debug page helpers do not
+   * restore it). */
+  return PhyStsFill::Power;
 }
 
 } /* namespace jaguar3 */
--- a/src/jaguar3/RtlJaguar3Device.cpp
+++ b/src/jaguar3/RtlJaguar3Device.cpp
@@ -322,12 +322,17 @@ void RtlJaguar3Device::StartRxLoop(Action_ParsedRadioPacket packetProcessor) {
          * present (monitor_rx_cfg enables APP_PHYSTS + RX_DRVINFO_SZ=4, so the
          * 32-byte report is counted in drvinfo). Skips C2H reports and any
          * frame whose drvinfo is too short (e.g. CCK, which carries no OFDM
-         * report). The report sits immediately after the 24-byte descriptor. */
-        if (!is_c2h && f.drvinfo_size >= 28)
-          jaguar3::parse_phy_sts_jgr3(data + off + jaguar3::RXDESC_SIZE_8822C,
-                                      f.drvinfo_size, p.RxAtrib);
+         * report). The report sits immediately after the 24-byte descriptor.
+         * f.physt (RX desc DW0 bit 26) says the PHY actually WROTE a report
+         * for this frame — the drvinfo space itself is reserved on every
+         * frame, so on A-MPDU subframes without the bit it holds stale bytes
+         * whose page nibble can alias 0/1 (contaminated RSSI/SNR tails). */
+        if (!is_c2h && f.physt && f.drvinfo_size >= 28)
+          p.RxAtrib.physt = jaguar3::parse_phy_sts_jgr3(
+              data + off + jaguar3::RXDESC_SIZE_8822C, f.drvinfo_size,
+              p.RxAtrib);
         p.Data = std::span<uint8_t>(const_cast<uint8_t *>(f.frame), f.frame_len);
-        if (!p.RxAtrib.crc_err) {
+        if (!p.RxAtrib.crc_err && p.RxAtrib.physt) {
           _rxq.add(p.RxAtrib.rssi[0], p.RxAtrib.snr[0], p.RxAtrib.evm[0]);
           _rxpaths.add(p.RxAtrib.rssi, p.RxAtrib.snr, p.RxAtrib.evm,
                        2); /* 8822C/8822E are 2T2R */
--- a/src/jaguar3/RtlJaguar3Device.cpp
+++ b/src/jaguar3/RtlJaguar3Device.cpp
@@ -318,25 +318,37 @@ void RtlJaguar3Device::StartRxLoop(Action_ParsedRadioPacket packetProcessor) {
           devourer::emit_tx_report(
               _logger->events(),
               devourer::parse_ccx_halmac(f.frame, f.frame_len), "halmac");
-        /* Decode the jgr3 PHY-status report (per-frame RSSI/SNR/EVM) when it is
-         * present (monitor_rx_cfg enables APP_PHYSTS + RX_DRVINFO_SZ=4, so the
-         * 32-byte report is counted in drvinfo). Skips C2H reports and any
-         * frame whose drvinfo is too short (e.g. CCK, which carries no OFDM
-         * report). The report sits immediately after the 24-byte descriptor.
-         * f.physt (RX desc DW0 bit 26) says the PHY actually WROTE a report
-         * for this frame — the drvinfo space itself is reserved on every
-         * frame, so on A-MPDU subframes without the bit it holds stale bytes
-         * whose page nibble can alias 0/1 (contaminated RSSI/SNR tails). */
+        /* Decode the jgr3 PHY-status report (per-frame RSSI/SNR/EVM), which
+         * sits immediately after the 24-byte descriptor inside the drvinfo
+         * area (monitor_rx_cfg enables APP_PHYSTS + RX_DRVINFO_SZ=4).
+         * RX_DRVINFO_SZ is a GLOBAL register, so those 32 bytes are reserved
+         * on EVERY frame — CCK included, and the parser decodes the CCK page
+         * 0 report — while the PHY writes a report only where the descriptor's
+         * PHYST bit (DW0 bit 26, f.physt) is set. Parsing without that bit
+         * reads bytes left over from an earlier frame, notably on all-but-one
+         * subframe of an A-MPDU, and a stale page nibble can alias 0/1 so the
+         * parse "succeeds" on garbage (contaminated RSSI/SNR tails). C2H
+         * reports carry no phy-status. */
+        PhyStsFill phy = PhyStsFill::None;
         if (!is_c2h && f.physt && f.drvinfo_size >= 28)
-          p.RxAtrib.physt = jaguar3::parse_phy_sts_jgr3(
+          phy = jaguar3::parse_phy_sts_jgr3(
               data + off + jaguar3::RXDESC_SIZE_8822C, f.drvinfo_size,
               p.RxAtrib);
+        /* The RAW descriptor bit, not the parse outcome — that is the meaning
+         * the shared field carries on Jaguar1 and the RTL8733B too, and what
+         * a caller needs to tell which A-MPDU subframe the report belonged to.
+         * `phy` is the local that says which fields are safe to fold. */
+        p.RxAtrib.physt = f.physt;
         p.Data = std::span<uint8_t>(const_cast<uint8_t *>(f.frame), f.frame_len);
-        if (!p.RxAtrib.crc_err && p.RxAtrib.physt) {
+        if (!p.RxAtrib.crc_err && phy != PhyStsFill::None) {
           _rxq.add(p.RxAtrib.rssi[0], p.RxAtrib.snr[0], p.RxAtrib.evm[0]);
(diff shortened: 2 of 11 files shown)


src/jaguar3/FrameParserJaguar3.h
```diff
@@ -303,7 +312,9 @@ inline void parse_phy_sts_jgr3(const uint8_t *physts, uint16_t physts_len,
       a.evm[i] = static_cast<int8_t>(physts[16 + i]);
       a.snr[i] = static_cast<int8_t>(physts[24 + i]);
     }
+    return true;
   }
+  return false;
```
josephnef: **Mutate-then-return-false breaks the new contract for OFDM pages 2-6.** By this point the function has already written `a.rssi[0..3]`, `ldpc`, `stbc`, `bw` from the OFDM common header — which is valid on *every* jgr3 OFDM page (per the comment above and the vendor `phy_sts_rpt_jgr3_ofdm_cmn` struct) — and then returns false, telling the caller "do not trust `a`'s signal fields". Two consequences: the half-mutated attrib flows to the packetProcessor tagged `physt=false`, and genuinely valid per-path RSSI is dropped from the EMAs. If the BB page selector ever leaves type1 (vendor auto-switch via 0x8c4, or the vendor debug page-switch helper, which never restores the page), every OFDM frame reads `physt=false` and `GetRxQuality`/`GetActiveRxPaths` freeze with no diagnostic. Latent under devourer's own config (the BB table pins 0x8C0[25:22]=1), but the contract is wrong today: either fill nothing on the false path, or return true for pages whose common header was parsed.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -322,12 +322,17 @@ void RtlJaguar3Device::StartRxLoop(Action_ParsedRadioPacket packetProcessor) {
          * present (monitor_rx_cfg enables APP_PHYSTS + RX_DRVINFO_SZ=4, so the
          * 32-byte report is counted in drvinfo). Skips C2H reports and any
          * frame whose drvinfo is too short (e.g. CCK, which carries no OFDM
```
josephnef: **This retained sentence is wrong on both counts and now contradicts the new text three lines below.** `RX_DRVINFO_SZ` is a global write (`HalJaguar3.cpp:485`: 0x060F=0x04 → 32 bytes on every frame, CCK included), and `parse_phy_sts_jgr3` explicitly decodes the CCK page-0 report — so CCK frames are neither short-drvinfo nor report-less. A maintainer trusting this sentence could skip the parse for CCK rates or "optimize" the `>= 28` guard, reintroducing exactly the misparse class this PR guards against. Please drop or correct it.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -322,12 +322,17 @@ void RtlJaguar3Device::StartRxLoop(Action_ParsedRadioPacket packetProcessor) {
...
+         * frame, so on A-MPDU subframes without the bit it holds stale bytes
+         * whose page nibble can alias 0/1 (contaminated RSSI/SNR tails). */
+        if (!is_c2h && f.physt && f.drvinfo_size >= 28)
+          p.RxAtrib.physt = jaguar3::parse_phy_sts_jgr3(
+              data + off + jaguar3::RXDESC_SIZE_8822C, f.drvinfo_size,
+              p.RxAtrib);
         p.Data = std::span<uint8_t>(const_cast<uint8_t *>(f.frame), f.frame_len);
-        if (!p.RxAtrib.crc_err) {
+        if (!p.RxAtrib.crc_err && p.RxAtrib.physt) {
           _rxq.add(p.RxAtrib.rssi[0], p.RxAtrib.snr[0], p.RxAtrib.evm[0]);
```
josephnef, line 344: **CCK page-0 passes the gate having filled only `rssi[0]`, so this line feeds `snr=0` and the CFO tracker gets 0.** `parse_phy_sts_jgr3` returns true for page 0 with `snr`/`evm`/`cfo_tail` never populated; `RxQuality.h` sums `snr_raw` unconditionally (EVM has a `!=0` filter, SNR does not), and `_cfo.add(0)` dilutes `avg_khz` below the 11-unit enable threshold, so on a 2.4 GHz channel with CCK beacon/probe traffic a real CFO can go uncorrected. The feed itself is pre-existing (the old gate was bare `!crc_err`), but the new per-page return was the natural place to require page 1 for the snr/cfo consumers — worth doing while the contract is being defined here.

tests/rx_physt_selftest.cpp
```diff
@@ -0,0 +1,58 @@
...
+  std::memset(buf, 0, kBufLen);
+  /* DW0: PKT_LEN[13:0] = 60, DRV_INFO_SIZE[19:16] = 4 (units of 8 bytes),
+   * SHIFT[25:24] = 0, PHYST = bit 26. */
+  buf[0] = kFrameLen;
+  buf[2] = kDrvInfo / 8;
+  if (physt)
+    buf[3] |= 0x04;
+}
+
+static void test_physt_bit_decoded() {
```
josephnef: **The selftest covers only the descriptor-bit decode — `parse_phy_sts_jgr3`'s new bool return, the value that actually gates the RF EMAs, is never exercised**, despite the header promising a plumbing regression "fails ctest instead of poisoning the RF EMAs". A regression in the return plumbing (the page-1 branch losing its `return true`, or the page gate flipping) passes ctest green while all OFDM RSSI/SNR/EVM accounting silently stops. The function is inline and already #included by this TU; four 28-byte synthetic pages cover it: page 0 → true + rssi[0]; page 1 → true + evm/snr/cfo; page 5 → false; 27 bytes → false. `tests/rtl8733b_rx_parse_selftest.cpp:20-37` has the `set_bits`/fixture-builder shape worth copying over the hand-poked masks.

josephnef (changes requested):
Reviewed with verification passes (every finding below was independently re-verified against the working tree; one candidate — the alleged doc duplication across the four new comment blocks — was checked and refuted, the copies are house-convention self-contained headers with location-specific scope).

The core gate itself checks out: DW0 bit 26 matches the vendor halmac layout, the selftest's byte encoding is right, `Packet p{}` is per-frame zero-initialized, and the new selftest builds and passes locally. The inline comments cover the in-diff issues; five more findings land on files this PR doesn't touch but its own premise implicates:

**1. The identical bug is left live on jaguar2 — and it steers a control loop there** (`src/jaguar2/RtlJaguar2Device.cpp:622`). The loop parses phy-status on only `!is_c2h && f.drvinfo_size >= 28`; `Rx8822bFrame` never decodes PHYST even though `GET_RX_DESC_PHYST_8822B` (same DW0 bit 26) sits unused at `FrameParserJaguar2.h:89` and `REG_RX_DRVINFO_SZ=4` is global (`HalJaguar2.cpp:2578`). On 8822BU/8821CU A-MPDU RX, subframes' reserved-but-unwritten drvinfo bytes are decoded as rssi/snr/evm/**cfo_tail**, and with `cfo_track` on, `_cfo.add(p.RxAtrib.cfo_tail)` (lines 631-632) steers the closed-loop XtalCap crystal trim from garbage. The PR's own comments describe the mechanism as generic; landing the fix at per-chip altitude leaves the worst instance of it in the tree.

**2. Jaguar1 has a plausible cousin of the same gap** (`src/jaguar1/FrameParser.cpp:237`). `pattrib.physt` is decoded at line 88 but the phy-status memcpy is gated on `drvinfo_sz` alone. The vendor driver passes `pattrib->physt ? pbuf+RXDESC_OFFSET : NULL` (rtl8812au `usb_ops_linux.c:179`); the same one-line gate is already parsed and sitting unused here. Affects per-path RSSI/SNR/EVM on `rx.path`/`GetActiveRxPaths` — including 8814AU 4-chain spatial-diversity measurements.

**3. Demo-side consumers are not gated on the new flag** (`examples/rx/main.cpp:911-913, 920-925, 1136-1146`). The hopset per-slot accumulators, the `g_rxagg` energy/LinkHealth aggregate, and `DEVOURER_RX_ALLPATHS` `rx.path` rows all read `rssi[0]/snr[0]/evm[0]` unconditionally, so jaguar3 A-MPDU subframes now inject deterministic rssi=0/snr=0/evm=0 samples. Pre-PR these frames carried random stale garbage; the PR converts that into a systematic directional bias (per-dwell means scale down with aggregation depth, `snr_min` pins to 0, FHSS exclusion scores a more-aggregating channel as worse, LinkHealth reads a strong link as weak). Cheap to close now that `RxAtrib.physt` exists to gate on.

**4. `rx.txhit`'s proof-of-what-flew contract weakens** (`examples/rx/main.cpp:1076`). `ldpc`/`stbc`/`bw` come only from `parse_phy_sts_jgr3`, so physt=0 subframes now deterministically report `ldpc=0/stbc=0` — indistinguishable from a real BCC fallback — and the event doesn't emit `physt` for consumers to tell the difference. In an encoding-truth-table run against an aggregating TX, a sampled txhit on a non-first subframe records an LDPC-qualification pass as BCC fallback. Emitting `physt` in the event (or skipping ldpc/stbc when unset) closes it.

**5. `physt` now carries three per-generation semantics in a shared struct** (`src/RxPacket.h:25`, no doc comment pinning one): raw descriptor bit (jaguar1 `FrameParser.cpp:88`, rtl8733b `Rtl8733bDevice.cpp:355`), bit AND drvinfo>=28 AND page-0/1-parsed (jaguar3, this PR), and never-written/always-false (jaguar2, kestrel). A cross-generation consumer keying on it as "hardware wrote a report" sees false on a jaguar3 PHYST=1 frame with an unrecognized page; one keying on it as "signal fields valid" discards every jaguar2/kestrel frame whose fields ARE populated. The rtl8733b shape is the clean one: raw bit into the field, parse-success in a local for the gate — plus one documented contract at the declaration.

Items 1 and 2 don't have to land in this PR, but item 1 at minimum deserves a tracked issue before merge — the CFO-trim exposure is a live correctness problem this PR's analysis proves exists.

src/jaguar3/FrameParserJaguar3.h
gilankpam: Fixed in 40a2439. Neither of your two options quite worked, because the CCK finding below is the same defect from the other side: one bool cannot express the two real tiers, and the old bool was wrong in both directions — false after filling valid common-header RSSI on pages 2..6, true for a CCK page that has no SNR/EVM/CFO in its layout at all.

`parse_phy_sts_jgr3` now returns `PhyStsFill` (`None` / `Power` / `Full`, declared in `src/RxPacket.h` next to `rx_pkt_attrib`): what it filled, not whether it liked the buffer. Pages 2..6 return `Power` — the common header was parsed and its per-path RSSI is a measurement, so it keeps feeding the EMAs, and the freeze scenario you describe (BB page selector leaving type1 via the vendor auto-switch or the debug helper that never restores the page) degrades to RSSI-only rather than a silent stop. `Full` is page 1 only.

`parse_phy_sts_jgr2` got the same return type, since it has the identical CCK/type1 split.

src/jaguar3/RtlJaguar3Device.cpp
gilankpam: Fixed in 40a2439. Verified both counts before rewriting: `HalJaguar3.cpp:485` writes `0x060F=0x04` unconditionally (32 drvinfo bytes on every frame), and `parse_phy_sts_jgr3` decodes the CCK page-0 report explicitly. The replacement comment says the drvinfo reservation is global and CCK included, and names the descriptor bit as the per-frame fact — so the `>= 28` guard now reads as the bounds check it is, not a rate filter.

src/jaguar3/RtlJaguar3Device.cpp
gilankpam: Fixed in 40a2439, and the downstream half turned out to be a live bug of its own. `RxQualityAccumulator::add` (`src/RxQuality.h`) did `snr_sum_ += snr_raw` and updated `snr_min_` unconditionally, while **its own doc comment three lines above** claimed "SNR/EVM are folded only when present (raw != 0 — CCK / non-type1 phy-status leaves them 0)". EVM and the noise-floor term did filter; SNR did not. So the dilution was not jaguar3-specific — jaguar1, jaguar2 and the 8733b all feed CCK frames through the same accumulator, and `snr_min_db` has been pinned to 0 on any 2.4 GHz channel with CCK traffic on all of them.

SNR now carries its own sample count like EVM, surfaced as `RxQuality::snr_valid` so a CCK-only window reports "no SNR" rather than a fabricated 0. `RxPathActivityAccumulator` already filtered per-metric and needed no change.

The CFO half is fixed at the caller as you suggested: `_cfo.add` now requires `PhyStsFill::Full` on both jaguar2 and jaguar3.

tests/rx_physt_selftest.cpp
gilankpam: Fixed in 40a2439 — the selftest now exercises the return value, and the `set_bits`/fixture shape from `rtl8733b_rx_parse_selftest.cpp` made the jaguar2 half nearly free, so it covers both generations:

- descriptor bit decoded, set and clear, on `parse_rx_8822c` and `parse_rx_8822b`
- `PhyStsFill` tiers: null buffer and 27-byte report → `None` (and the attrib is verified untouched); CCK page → `Power` with `snr`/`evm`/`cfo_tail` asserted still 0; OFDM page 1 → `Full` with rssi/snr/evm/cfo/ldpc/bw values checked; OFDM page 5 → `Power` with the common-header RSSI present and the type1-only fields still 0

`rx_quality_selftest` picked up the mixed-window and CCK-only-window SNR cases.

CMake registration widened to `DEVOURER_JAGUAR3 OR DEVOURER_JAGUAR2_8822B OR DEVOURER_JAGUAR2_8821C`; verified the JAGUAR1-only, JAGUAR2-only and JAGUAR3-only configs all build and that the test is absent in the first and passes in the other two.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -322,12 +322,17 @@ void RtlJaguar3Device::StartRxLoop(Action_ParsedRadioPacket packetProcessor) {
...
-                                      f.drvinfo_size, p.RxAtrib);
+         * report). The report sits immediately after the 24-byte descriptor.
+         * f.physt (RX desc DW0 bit 26) says the PHY actually WROTE a report
+         * for this frame — the drvinfo space itself is reserved on every
+         * frame, so on A-MPDU subframes without the bit it holds stale bytes
+         * whose page nibble can alias 0/1 (contaminated RSSI/SNR tails). */
+        if (!is_c2h && f.physt && f.drvinfo_size >= 28)
+          p.RxAtrib.physt = jaguar3::parse_phy_sts_jgr3(
+              data + off + jaguar3::RXDESC_SIZE_8822C, f.drvinfo_size,
+              p.RxAtrib);
```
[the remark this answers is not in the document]
gilankpam: Taken — `RxAtrib.physt` is now the raw descriptor bit on jaguar3, matching jaguar1 and the rtl8733b, with the parse outcome in a local (`PhyStsFill`) for the EMA gate. That was the alternative offered in the PR description, and the reviewer independently landed on the same shape, so it is settled.

Two additions beyond the suggestion, both from verifying the premise:

- The field now has a documented contract at its declaration (`src/RxPacket.h`) pinning the meaning as "the PHY wrote a report into this frame's drvinfo" — explicitly *not* "the report parsed" and *not* "the signal fields are valid" — and noting that Kestrel never sets it, since its PHY status arrives as a separate PPDU-status frame.
- Jaguar2 was silently in the third state you list: never decoded, always false. `GET_RX_DESC_PHYST_8822B` existed unused at `FrameParserJaguar2.h:89` while the RX loop gated on drvinfo size alone, so the same stale-drvinfo garbage was reaching `_cfo.add()` and steering the closed-loop XtalCap trim. That is fixed in the same commit, so all four generations that carry per-frame drvinfo now populate the field identically.

gilankpam:
Pushed 40a2439 addressing the review. Every finding was checked against the working tree before acting; here is where each landed.

**In-diff (all four inline threads answered on the threads themselves):** the mutate-then-return-false contract, the stale CCK comment, the CCK page feeding `snr=0`/`cfo=0`, and the selftest that did not exercise the return value.

**Items 1 and 2 — both confirmed, both fixed here rather than tracked.**

Item 1 (jaguar2): verified end to end — `GET_RX_DESC_PHYST_8822B` sitting unused at `FrameParserJaguar2.h:89`, `0x060F=4` written globally at `HalJaguar2.cpp:2578`, and `_cfo.add` at `RtlJaguar2Device.cpp:631` steering the XtalCap trim from it. It was ~4 lines once the enum from the jaguar3 fix existed, which was smaller than writing the issue would have been.

Item 2 (jaguar1): also confirmed — `_InitDriverInfoSize_8812A(DRVINFO_SZ=4)` at `HalModule.cpp:268` is the global write, so the size check was never equivalent to the vendor gate the comment at `FrameParser.cpp:231` already cited. One line, plus gating the EMA feed so the CFO tracker does not take a zero tail from a physt=0 frame.

The PR title and description now cover all three generations.

**Item 5 — taken, plus a fourth semantics you did not list.** `RxAtrib.physt` is the raw descriptor bit everywhere, with the parse outcome in a local, and the field now has a documented contract at its declaration. The fourth case was jaguar2 in the never-written state — so all four generations that carry per-frame drvinfo now populate it identically, and the contract notes Kestrel deliberately does not.

**Item 3 — one correction before it becomes a follow-up.** Gating the demo consumers on `RxAtrib.physt` would drop *every* sample on Kestrel, which never sets it (and never can, given how its physts arrives). The generation-neutral gate is `rssi[0] > 0`, which is what `RxQualityAccumulator::add` already applies internally. Left out of this PR to keep it library-side, along with item 4.

**One thing the CCK finding turned up that was not in the review.** Chasing where the zero SNR actually lands: `RxQualityAccumulator::add` folded SNR into `snr_sum_`/`snr_min_` unconditionally, while its own doc comment three lines above claimed it folded only when present. EVM and the noise-floor term did filter; SNR did not. So the dilution was never jaguar3-specific — jaguar1, jaguar2 and the rtl8733b all feed CCK frames through the same accumulator, and `snr_min_db` has been pinned to 0 on any 2.4 GHz channel with CCK traffic on all of them. Fixed with a per-metric sample count, surfaced as `RxQuality::snr_valid`.

`ctest` 55/55 (`la_csi_math` skipped); JAGUAR1-only, JAGUAR2-only and JAGUAR3-only configs all build clean, with the selftest correctly absent from the first. Still draft pending on-air re-validation.

https://claude.ai/code/session_01JiAS32956Z3cmTbnmcXYkT

josephnef (approved):
All review findings verified fixed at ca61f2d — approving.

Re-verified against the branch, not just the replies: the jaguar2 PHYST gate (including the `PhyStsFill::Full` requirement on the XtalCap `_cfo.add` feed), the jaguar1 memcpy gate, the three-tier `PhyStsFill` return (pages 2..6 as `Power` is a better shape than either option I offered — the common-header RSSI stays a measurement and the page-selector freeze scenario degrades to RSSI-only), the raw-bit `physt` contract documented at the `RxPacket.h` declaration, the corrected CCK/drvinfo comment, and the widened selftest (build + full ctest green locally, 56/56, including the new `rx_physt_bit` tiers and the `rx_quality` mixed/CCK-only windows).

The `RxQualityAccumulator` SNR find deserves its own credit: the unconditional `snr_sum_ += snr_raw` contradicting the accumulator's own doc comment was live on every generation, and `snr_valid` is the honest surface for a CCK-only window.

The two demo-side items from the review body (gating the `examples/rx` accumulators on the now-available `RxAtrib.physt`, and emitting `physt` on `rx.txhit`) plus the same ungated-SNR shape in `chanscout` and `cell::UeRxAttribution` are follow-up-sized and tracked separately rather than holding this fix.

PR #410, opened 2026-09-04
rtl8733b: a live ACK disarm must move the identity, not just close the gate

snokvist:
## Problem

On RTL8733B, clearing port-0 `net_type` does not stop immediate responses. A
never-armed port already answers unicast frames addressed to the MACID written
during bring-up, even while `net_type` reads NoLink. Consequently the previous
gate-only `ClearAckResponder()` could report success while the port continued
answering for the configured responder address.

This also exposed a test gap: the matrix's old `off` row starts a fresh process
without ever arming a responder. It was a useful never-armed control, but could
not prove a live arm -> disarm transition.

## Change

- Keep the shared `ack::disable()` gate-only for generations where that is the
  measured recipe.
- Compose RTL8733B disarm from a verified gate clear plus a verified MACID
  retarget to the adapter's valid EFUSE MAC.
- Treat final MACID readback as authoritative even when a transport write
  reports failure; log the inconsistent transport result without returning the
  contradictory "state unknown" after identity is verified.
- Add an RTL8733B-only rxdemo measurement hook. Its timer starts after locked
  bring-up, arming, channel setup, and monitor-RX configuration, so a zero-ms
  run cannot clear a cold port and then be undone by later initialization.
- Add a same-process `disarmed` matrix row for the RTL8733B USB IDs. The harness
  now waits for exact arm/RX/disarm evidence, checks process liveness, requires
  minimum submitted-frame and 80% report-coverage counts, requires at least
  90% of an OFF cell's reports to be retry-pinned, validates numeric controls
  before touching hardware, propagates analyzer failures, and aborts on early
  TX exit instead of producing a green result from empty, sparse, or
  outlier-driven data.
- Correct the generic documentation that described a never-armed monitor as
  necessarily passive or `net_type` as a universal reverse gate.

Restoring the initialization MAC is intentional. Zero is still a unicast
address by the I/G test and can inhibit Realtek TX scheduling; an arbitrary
parking address has not been qualified. This operation restores the pre-arm
identity, not radio silence: on RTL8733B the port may still answer for that
unadvertised initialization address, exactly as it did before arming. Arming
on that same address is therefore a behaviorally idempotent corner case.

## On-air verification

All verdicts are scored on the soliciting adapter's CCX `tx.report`, with retry
limit 12 and high report coverage.

| responder / solicitor | state | result |
|---|---|---|
| `f72b` / RTL8812AU, current tree | armed | 751/751 reported frames ACKed; retries mean 0.00 |
| `f72b` / RTL8812AU, current tree | retargeted | 754/754 ACKed; retries mean 0.01 |
| `f72b` / RTL8812AU, current tree | never armed | 0/741 ACKed; retries mean 12.0 |
| `f72b` / RTL8812AU, current tree, zero-ms post-bring-up disarm | disarmed | 0/758 ACKed; retries mean 12.0 |
| maintainer `b733` / RTL8822CU | armed | `ack_rate=1.00`, retries mean 0.04 |
| maintainer `b733` / RTL8822CU, 0-ms and 2000-ms disarm | disarmed | `ack_rate=0.00`, retries mean 12.0 |
| vehicle-mounted `f72b` / RTL8812AU | armed | 1344/1344 ACKed |
| vehicle-mounted `f72b` / RTL8812AU | disarmed | 0/1342 ACKed; retries mean 12.0 |

The final local matrix completed with 99.9-100% report coverage. A deliberately
occupied responder adapter also made the hardened harness exit nonzero before
transmission, confirming that the old zero-frame false-pass path is closed.

## Other verification

- `ctest`: 55/55
- default build
- RTL8733B-only build
- build without RTL8733B support
- ShellCheck and `git diff --check`
- invalid delay and non-RTL8733B hook requests reject with exit status 1

## Scope boundary

This proves single-frame ACK behavior for the tested RTL8733B samples and
solicitors. It does not claim that a live disarm suppresses BlockAck on this
die, that restoring MACID is required on other generations, or that
`net_type` is inert elsewhere. The independently observed RTL8812AU/Jaguar1
gate-only disarm defect is deliberately left for a separate hardware-tested
follow-up.


src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -885,7 +887,54 @@ void Rtl8733bDevice::ClearAckResponder() {
...
+   * rather than to silence — on a die that ignores the gate, a port matching
+   * its own MAC may still answer a frame addressed to it, exactly as an
+   * unarmed monitor radio would. This cannot create an exposure monitor mode
+   * does not already have, and devourer never advertises that address.
+   *
+   * enable() also writes BSSID (0x0618/0x061c); this does not restore it,
+   * because program_mac never wrote it and there is nothing to restore to.
+   * The on-air result — 100 % unanswered after this disarm, with the BSSID
+   * residue standing — is the evidence that the engine does not match on it. */
+  if (!devourer::ack::retarget(_device, _efuse.mac.data())) {
```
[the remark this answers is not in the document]
snokvist: Confirmed and fixed in 0e94dc9 — this was a real hole, and it is the same failure mode the PR exists to close, reintroduced through the one address the disarm cannot help with.

`ClearAckResponder` disarms by retargeting the identity back to `_efuse.mac`. If the responder was armed to that same address, both MACID writes and the readback succeed without moving the match, and the success line is logged while the port keeps answering.

Fixed by refusing the arm rather than papering over the disarm: `SetAckResponder` now rejects `mac == _efuse.mac` on this die with an explicit reason. The alternative — deriving some other restore address — trades a corner case for a port left on an address that is neither the responder’s nor the adapter’s, and it seemed better simply not to offer an arm that cannot be taken back. Other generations disarm through the net_type gate alone and are unaffected.

Also correct that the self-test hid this by using distinct responder and restore addresses; its “does not cover” block now names the case, since the refusal is device-layer and not reachable from the register-file model.

src/rtl8733b/CLAUDE.md
```diff
@@ -316,14 +316,21 @@ session that quietly answers nothing. And the disarm is unconditional inside
...
+either way.
 
 Measured against an RTL8812AU soliciting TX (`tests/ack_txreport_matrix.sh`,
 8733B as RESPONDER): armed 1725/1725 reports ACKed at retries_mean 0.00;
 re-armed on a **different** MAC, 1728/1728 again (the address is arbitrary, not
-baked in); disarmed, 0/1723 successful reports with retries pinned at 12.
+baked in); and 0/1723 successful reports with retries pinned at 12 in the OFF
+cell. That cell is **never-armed, not disarmed**: `run_phase()` starts no
+responder process at all when the responder MAC is empty. A real disarm was
+later measured NOT to stop this die — clearing net_type leaves a port whose
```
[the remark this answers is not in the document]
snokvist: Agreed and fixed in 0e94dc9. The paragraph now points at `src/AdapterCaps.h` (capability and its measured limits) and `src/AckResponder.h` / `Rtl8733bDevice::ClearAckResponder` (the recipe and why this die needs a retarget), keeping only the harness name.

Worth noting the rule earned its keep here: the “disarm-proof” mislabel this PR corrects had propagated into exactly two places — the `AdapterCaps.h` row and this CLAUDE.md paragraph — which is the drift the rule is meant to prevent.

src/rtl8733b/Rtl8733bDevice.cpp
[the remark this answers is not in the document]
snokvist: Follow-up in 586c1c1 — this finding is now covered in both directions, which it was not when I first replied.

The refusal was device-layer only, so nothing in the register-file model could reach it. The decision now lives in `ack::disarmable_by_retarget(responder, restore)` — a predicate rather than a check inside `retarget()`, since `retarget()` cannot know what the port was armed to. The self-test pins the predicate and demonstrates the underlying failure on the modelled register file: arm on `own`, disarm, retarget to `own` — every write and the readback succeed while the identity never moves, so `retargeted()` reports success for a port still matching the address it was armed to. Mutation-checked (a predicate hardcoded to `true` fails the suite).

Device-verified with devourer's own harness on an RTL8733B, since `DEVOURER_ACK_RESPONDER` reaches the same `SetAckResponder` path and `GetPermanentMacAddress` returns exactly the `_efuse.mac` the guard compares against:

- `DEVOURER_ACK_RESPONDER=<adapter own EFUSE MAC>` -> refused, then #406 init throw, rxdemo exit 134
- `DEVOURER_ACK_RESPONDER=02:12:34:56:78:9a` (control) -> `hardware ACK responder armed ... (net_type=AP)`, ran its full window

The control is what makes it evidence — the refusal is specific to the own MAC, not a broken build.

One observation for the demos rather than something I changed here: `rxdemo` has no try/catch around bring-up, so a refused configured responder surfaces as `std::terminate` rather than a clean exit. `examples/tx/main.cpp` gained that guard in #402; `rxdemo` did not. Pre-existing — this refusal is just a new way to reach it. Happy to add it here or leave it to a separate change, whichever you prefer.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -838,13 +838,29 @@ bool Rtl8733bDevice::SetAckResponder(const devourer::MacAddr &mac) {
...
+   * be switched off without a re-init. Other generations disarm through the
+   * net_type gate alone and carry no such restriction. */
+  if (!devourer::ack::disarmable_by_retarget(mac.data(), _efuse.mac.data())) {
+    _logger->error("RTL8733B: ACK responder cannot be armed to the adapter's "
+                   "OWN MAC — the disarm restores that address, so it could "
+                   "not be turned off again without a re-init");
+    return false;
+  }
   if (!devourer::ack::enable(_device, mac.data())) {
     if (!devourer::ack::disable_verified(_device)) {
```
josephnef: **Arm-failure rollback is still gate-only.** `enable()` writes MACID 0x0610/0x0614 (and BSSID) before its final net_type write. If that write reports failure, or `verify()` reads the gate back wrong, both rollback branches here run only `disable_verified()`, log "not armed" and return false — MACID stays at the responder address. By this PR's own measurement that is a live responder on this die, and nothing else restores MACID (`stop()` deliberately declines to). Run the same `retarget()` + `retargeted()` in both branches, or factor a `disarm_verified()` shared with `ClearAckResponder`.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -885,7 +901,54 @@ void Rtl8733bDevice::ClearAckResponder() {
     _logger->error("RTL8733B: ACK responder disarm did not latch");
```
josephnef: **Retarget is skipped whenever the gate half fails.** A single timed-out control read of 0x0102 throws in the transport, `disable_verified()` catches it and returns false, and this early return means `retarget()` — pure writes to 0x0610/0x0614 that would likely succeed — is never attempted. Same if the readback merely shows the gate still set. The identity half is the one measured to actually silence this die, yet it is only tried when the half measured not to matter succeeds first. `retarget()` applies a "both halves must be ATTEMPTED" doctrine internally; apply it across the gate/identity pair too: log the gate failure and fall through to `retarget()`/`retargeted()` unconditionally.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -885,7 +901,54 @@ void Rtl8733bDevice::ClearAckResponder() {
...
+   * rather than to silence — on a die that ignores the gate, a port matching
+   * its own MAC may still answer a frame addressed to it, exactly as an
+   * unarmed monitor radio would. This cannot create an exposure monitor mode
+   * does not already have, and devourer never advertises that address.
+   *
+   * enable() also writes BSSID (0x0618/0x061c); this does not restore it,
+   * because program_mac never wrote it and there is nothing to restore to.
+   * The on-air result — 100 % unanswered after this disarm, with the BSSID
+   * residue standing — is the evidence that the engine does not match on it. */
+  if (!devourer::ack::retarget(_device, _efuse.mac.data())) {
```
josephnef: **The four rows never separate "net_type ignored on disarm" from "net_type inert on this die".** Every 100 %-unanswered control had MACID at the EFUSE address while the peer solicited the responder MAC. The simpler model — the 8733B engine matches MACID only and 0x0102[1:0] does nothing — fits all four rows, and the comment above half-concedes it ("a port matching its own MAC may still answer ... exactly as an unarmed monitor radio would") without a cell that solicits a never-armed 8733B on its own EFUSE MAC. Under that model every never-armed 8733B monitor already auto-ACKs unicast to its own MAC, an exposure asserted as "cannot create" but never measured. Two cells decide it, both fitting `ack_txreport_matrix.sh` (`run_phase` already takes the RA with responder ''): (a) MACID = responder with net_type left 0, peer soliciting the responder; (b) never-armed port, peer RA = the adapter's EFUSE MAC. The asymmetric-gate story in `AckResponder.h` / `IRtlDevice.h` should only be written once one of those has run.

src/rtl8733b/Rtl8733bDevice.cpp
```diff
@@ -838,13 +838,29 @@ bool Rtl8733bDevice::SetAckResponder(const devourer::MacAddr &mac) {
...
   }
+  /* Refused on this die, and only on this die: ClearAckResponder disarms by
+   * retargeting the identity BACK to _efuse.mac, so a responder armed to that
+   * very address could not be moved off it. The port would keep answering
+   * while the disarm reported success — the exact silent-live-responder
+   * failure this backend's disarm exists to prevent. Refusing the arm is the
+   * honest end of the trade: the alternative is an ACK responder that cannot
+   * be switched off without a re-init. Other generations disarm through the
+   * net_type gate alone and carry no such restriction. */
+  if (!devourer::ack::disarmable_by_retarget(mac.data(), _efuse.mac.data())) {
```
josephnef: Follows from the comment on the measurement below: if net_type is inert on this die, the state this refusal guards (MACID = own MAC, gate closed) is identical to the state every never-armed session ships in, so the refusal is a per-chip API asymmetry with no safety payoff. Every other generation accepts an arm on the adapter's own MAC. Worth deciding on the strength of cell (b) rather than assuming it.

src/AdapterCaps.h
```diff
@@ -176,8 +176,15 @@ struct AdapterCaps {
    * was never verified, so a silently dead responder read as on=0/off=0),
    * 8822B, 8812C/8822C, 8812E/8822E (the 8811A rides the 8812 die path and
    * inherits its row), 8733B (1725/1725 frames ACKed at retries_mean 0.00,
-   * retarget-proof and disarm-proof — tests/ack_txreport_matrix.sh run with
-   * the 8733B as the responder). False-as-unmeasured (the
+   * and retarget-proof: re-armed on a different MAC, 1728/1728 —
+   * tests/ack_txreport_matrix.sh run with the 8733B as the responder). That
+   * matrix does NOT establish disarm on any die, and an earlier revision of
```
josephnef: Changelog framing ("an earlier revision of this row said", "was later measured NOT to") — master's most recent commit swept exactly this out of the current-state docs. State the timeless fact: *That matrix's off cell is never-armed (run_phase passes an empty responder MAC), so it establishes no disarm on any die; on the 8733B a gate-only disarm was measured not to stop the engine.* And line 174 four lines up still calls the 8821A "disarm-proof-verified" from the same never-armed shape.

src/AckResponder.h
```diff
@@ -85,6 +93,70 @@ inline bool disable(RtlAdapter &dev) {
...
+ * schedule a frame when the MAC ID is zero — the T1 canary bug that programming
+ * REG_MACID was introduced to fix (src/jaguar1/HalModule.cpp,
+ * EepromManager.h) — and a radio being disarmed may still be injecting. Zero
+ * would also not remove the match, only move it: 00:00:00:00:00:00 has the I/G
+ * bit clear, so is_unicast() accepts it and a peer could solicit it.
+ *
+ * Gate untouched on purpose: this is the identity half, so a caller can compose
+ * it with disable() in whichever order its die needs, and no generation gets a
+ * behaviour change it was not measured for. */
+inline bool retarget(RtlAdapter &dev, const uint8_t mac[6]) noexcept {
```
josephnef, line 148: `retarget()` and `retargeted()` add a 4th and 5th hand-written copy of the MAC → le32/le16 packing that `enable()` (twice) and `verify()` already carry, in a header whose `verify()` comment promises to keep the register map in one place so `enable()` cannot silently diverge from a copy. An endianness/width fix applied to `enable()` leaves `retargeted()` reading the old layout while the 8733B disarm logs "MACID restored", and the selftest checks both against the same fake so it would not catch it. Factor `macid_lo(mac)` / `macid_hi(mac)` (or a `macid_is(dev, mac)`), express `verify()` as gate-check && `macid_is()`, and route `retarget()`/`retargeted()` through them.

tests/ack_responder_selftest.cpp
```diff
@@ -0,0 +1,214 @@
...
+    CHECK(!devourer::ack::disarmable_by_retarget(own, own));
+    /* And the reason it matters, demonstrated on the register file: arming on
+     * `own` and then "disarming" to `own` leaves the identity untouched, so
+     * retargeted() reports success while the port still matches the address it
+     * was armed to. */
+    CHECK(devourer::ack::enable(dev, own));
+    CHECK(devourer::ack::disable_verified(dev));
+    CHECK(devourer::ack::retarget(dev, own));    /* writes what is already there */
+    CHECK(devourer::ack::retargeted(dev, own));  /* "succeeds" */
+    CHECK(devourer::ack::retargeted(dev, own));  /* still the armed address */
```
josephnef: Lines 194 and 195 are the byte-identical assertion under two different comments, so the second cannot fail unless the first does and the property the comment names ("still the armed address") is not what is checked. Assert the raw registers (`read32(kMacId) == 0xccbbaa02`, `read16(kMacId + 4) == 0x01dd`) or a no-change on `write_calls`/`mem` instead.

tests/ack_responder_selftest.cpp
```diff
@@ -0,0 +1,214 @@
...
+    return static_cast<int>(len);
+  }
+  void rx_loop(int, int, const std::function<void(const uint8_t *, int)> &,
+               const std::function<bool()> &) override {}
+};
+
+constexpr uint16_t kNetType = 0x0102;
+constexpr uint16_t kMacId = 0x0610;
+
+bool macid_clear(FakeRegs &r) {
```
josephnef: `macid_clear()` is never called. Delete it, or use it for the "NOT zero" check at line 143.

josephnef (changes requested):
The fix direction is right and the selftest builds and passes here, but the same file keeps two paths that leave the 8733B in exactly the state this PR exists to remove, and the composed disarm cannot be reached or re-measured from anything in the repo. Inline comments cover those; three things sit on lines outside the diff:

**Stale claims that survive this PR.** `docs/rtl8733b.md:382` still cites `off 0/1723 with retries pinned at 12` as the SetAckResponder off evidence — the cell `AdapterCaps.h` now declares never-armed. `docs/aggregation.md:194` still says "The identity+net_type pair is the whole gate" for every generation, which contradicts the new `AckResponder.h` header ("NOT reliable in reverse"). `docs/scheduled-mac.md:244` and `AdapterCaps.h:174` still call the 8821A "disarm-proof-verified" from a responder-powered-down cell, i.e. the same never-armed shape this PR names as a fallacy four lines below. Either retract that or drop the "same fallacy as the 8821AU line" assertion.

**The measurement is unreproducible from the tree.** `git grep ClearAckResponder` at 586c1c1 finds only definitions — no demo, env knob, signal path or `tests/*.sh` calls it — and the on-air numbers live in two comments with different values (0.42/0.53 % in `ClearAckResponder`, 0.63/1.15 % in the selftest header). `src/rtl8733b/CLAUDE.md` then points at `tests/ack_txreport_matrix.sh` as "the harness", the script the CMake comment says cannot cover the disarm. Please add a mid-run disarm trigger to rxdemo (or an armed-then-disarmed phase in the matrix script), keep one number set, and point CLAUDE.md at that.

**Pre-existing, not caused here, but worth knowing:** `examples/chanmig/main.cpp:574` arms the responder on `57:42:75:xx:xx:01`. `0x57` has the I/G bit set, so the `is_unicast()` refusal from #406 makes `chanmig --role drone` exit 1 on every generation today — and this PR's selftest uses `0x57` as its group-address example.

src/rtl8733b/Rtl8733bDevice.cpp
josephnef: Ran the two cells from the finding on `Rtl8733bDevice.cpp:941`. Both answered.

**Setup.** RTL8733B (`0bda:b733`, EFUSE MAC `14:0a:02:15:af:32`) as the responder in monitor RX on this PR's build (586c1c1), RTL8822CU (`0bda:c812`) soliciting unicast QoS-Data at MCS3 on channel 36, ~1700 frames per cell. Verdict is the soliciting side's CCX `tx.report`, scored by `tests/ack_txreport_analyze.py`. Cell (a) used a throwaway knob on a scratch build that performs `enable()`'s MACID+BSSID writes and skips its net_type write; the log confirmed `net_type=0x00` read back after the identity write. Fresh process per cell, ordered so no cell inherits state from an armed one.

| cell | responder state | RA | ack_rate | retries mean |
|---|---|---|---|---|
| never_armed_ctrl | never armed | MAC1 | 0.00 | 12 (pinned) |
| **cell_b** | never armed | own EFUSE MAC | **1.00** | 0.05 |
| **cell_a** | MACID=MAC1, net_type read back 0x00 | MAC1 | **1.00** | 0.05 |
| armed_ctrl | armed, net_type=AP | MAC1 | 1.00 | 0.05 |

**What it decides.**

- On this die net_type `0x0102[1:0]` is inert in **both** directions, not "arms but does not disarm". The ACK engine matches `REG_MACID` alone. `retarget()` is the entire disarm here and the `disable()` half is decoration, so the "NOT reliable in reverse" framing in `AckResponder.h` / `IRtlDevice.h` and the asymmetric-gate story in the `ClearAckResponder` comment should be replaced with that simpler statement.
- A never-armed 8733B monitor session already auto-ACKs any unicast frame addressed to its EFUSE MAC. The PR's "cannot create an exposure monitor mode does not already have" is literally true, but the exposure itself is undocumented and contradicts the passive-monitor contract on this die. Worth a sentence in `AdapterCaps.h` and `src/rtl8733b/CLAUDE.md`.
- The own-MAC arm refusal (`:858`) guards a state identical to the default, so it buys no safety and leaves a per-chip API asymmetry. Drop it.
- A truly passive 8733B needs a parking MACID, which is a separate follow-up: zero is out (the MAC refuses to schedule TX with MACID 0), so any parking value needs its own on-air cell.

**Caveats.** One unit, one soliciting adapter, one rate and width (MCS3 / 20 MHz). The two findings above about the rollback path (`:865`) and the early return (`:901`) stand as written; if anything they matter more, since the identity write is the only lever.

snokvist: You were right, and it is the second reading — `net_type` is inert on this die, not merely ignored on disarm. Ran your cell (b), plus a negative control and the armed positive control. Single-shot ACK rate (retry limit 0), MCS3, Jaguar1 soliciting, scoring the soliciting side's CCX `tx.report`:

| responder port | peer RA | ACKed |
|---|---|---|
| never-armed | adapter own EFUSE MAC | **85.2 %**, repeat **82.5 %** |
| never-armed | address nobody holds | **0.0 %** (control) |
| armed to X | X | 83.3 % |

A never-armed monitor answers on its own MAC at the same rate as a deliberately armed responder, with a clean negative control. So the engine matches MACID and `0x0102[1:0]` decides nothing here: `program_mac` is what makes it answer, arming merely repoints it.

Cell (a) was already in the PR without my recognising it as such — the pre-fix "disarmed, gate only → 0.53 %" row is exactly MACID = responder with the gate closed, answering.

The asymmetric-gate story is gone from `AckResponder.h`, `IRtlDevice.h`, `AdapterCaps.h` and the backend comment; they now state the inert gate and, explicitly, that every never-armed RTL8733B monitor already auto-ACKs unicast to its own MAC. I had asserted that exposure "cannot be created" — it is pre-existing, which is the same conclusion but now measured rather than argued. The fix itself is unchanged: the identity is still the only thing a disarm can move.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Real bug, fixed. `enable()` writes MACID before its final gate write, so a failed write or a failed `verify()` left the identity on the responder address with nothing to restore it — a live responder by this PR's own measurement.

Factored `Rtl8733bDevice::disarm_ack_responder()` as you suggested and routed all three callers through it: `ClearAckResponder` and both rollback branches in `SetAckResponder`.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Agreed — and it was the wrong way round, as you say: the identity half is the one that decides anything, and it was gated on the half that decides nothing. Worse than I realised when I wrote it, now that the cells show `net_type` is inert.

`disarm_ack_responder()` attempts both halves unconditionally; a gate failure is logged and falls through to the identity writes. Same "both halves must be ATTEMPTED" doctrine `retarget()` applies internally, now applied across the pair.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Removed, on the strength of cell (b) rather than assumption. It guarded MACID = own MAC with the gate closed, which the measurement shows is exactly the state every never-armed session ships in — a per-chip API asymmetry with no safety payoff. The `disarmable_by_retarget` predicate that backed it is gone too.

This also dissolves the earlier Qodo "own-MAC disarm stays armed" finding: arming to the own MAC and disarming are the same state, so there is nothing left to disarm to.

src/AdapterCaps.h
snokvist: Reframed to the timeless statement, close to your wording: the matrix establishes arming only, its off cell is never-armed (`run_phase` passes an empty responder MAC), so no row says anything about disarm on any die. Then the 8733B facts as current state, not as a correction.

Line 174 fixed too — the 8821A no longer claims "disarm-proof-verified". Same never-armed shape, same non-claim; I had seen it and wrongly left it as out of scope.

src/AckResponder.h
snokvist: Fair, and the header says so itself. Factored `macid_lo()` / `macid_hi()` / `macid_is()`; `enable()`, `verify()` and `retarget()` all route through them, and `retargeted()` is deleted — it was `macid_is()` under another name, which is exactly the duplication you describe.

The self-test gained a cell that pins `verify()` and `macid_is()` against the raw bytes in the fake, so a width or endianness change reaching only one of them fails rather than diverging silently.

tests/ack_responder_selftest.cpp
snokvist: Correct — two comments, one assertion, so the second could not fail independently and did not check the property its comment named. That block is gone with the own-MAC refusal; the surviving retarget cell now asserts the raw registers (`read32(kMacId) == 0xccbbaa02`, `read16(kMacId + 4) == 0x01dd`) plus a non-zero check, all independent of the readback helpers.

tests/ack_responder_selftest.cpp
snokvist: Dead after the rework — deleted. The "NOT zero" check at that cell now asserts the raw registers directly.

src/rtl8733b/Rtl8733bDevice.cpp
snokvist: Thank you for running it — and on a second unit (`0bda:b733`, different EFUSE MAC) with a different soliciting adapter. That makes it a genuine per-unit replicate rather than my one dongle, and your cell (a) via a scratch build that skips the net_type write is the clean version of the thing I could only approximate.

Your `1.00` on both cells matches my `85.2 % / 82.5 %` (the difference is single-shot rate at my link margin, not disagreement — my `0.0 %` control and your `never_armed_ctrl 0.00` line up exactly).

All four of your conclusions are now in the diff: the asymmetric-gate framing is replaced with "net_type is inert here, the engine matches MACID" in `AckResponder.h`, `IRtlDevice.h`, `AdapterCaps.h` and the backend comment; the exposure has its own sentence in `AdapterCaps.h` and `src/rtl8733b/CLAUDE.md`; the own-MAC refusal is dropped; and the parking MACID is left as the follow-up you describe, with the reason zero cannot be it recorded next to `retarget()`.

snokvist:
Pushed `ee0d36e`. The three out-of-diff items from the review body are done, and the reachability complaint was the right one — it was the weakest part of the PR.

**Reachability.** `git grep ClearAckResponder` finding only definitions was exactly the problem: nothing in the tree could arm and then disarm inside one session, which is the only shape in which a disarm can be measured or regress. `rxdemo` gains `DEVOURER_ACK_DISARM_AFTER_MS=<n>` (side thread, since `Init()` runs the loop and does not return; ignored without `DEVOURER_ACK_RESPONDER`), and `tests/ack_txreport_matrix.sh` gains a `disarmed` phase that drives it — arm on MAC1, disarm at 2 s, solicit MAC1.

Scored with it, single-shot ACK rate, MCS3, Jaguar1 soliciting an RTL8733B responder:

| cell | ACKed |
|---|---|
| armed, no disarm | 72.2 % |
| armed, disarmed at 2000 ms | **0.0 %** |

First time the disarm has been measured with this repository’s own tooling rather than a downstream consumer’s counters.

**One number set.** The disarm-cycle percentages that disagreed between two comments are gone. What remains is the three-cell block (85.2/82.5 % never-armed on own MAC, 0.0 % control, 83.3 % armed) quoted identically in `AckResponder.h`, `AdapterCaps.h`, `Rtl8733bDevice.cpp` and the selftest header, plus the two rows above.

**Retracted.** `docs/rtl8733b.md` (the `off 0/1723` cell now says never-armed and points at the `disarmed` phase), `docs/aggregation.md` (which half of identity+net_type is the gate is per-die), `docs/scheduled-mac.md` (8821A "disarm-proof-verified" retracted, as `AdapterCaps.h:174` already was). `src/rtl8733b/CLAUDE.md` now states the exposure plainly rather than implying a monitor session is passive on this die.

**Deliberately not here:** the parking MACID. Zero is out for the reason you give, so any parking value needs its own on-air cell — a separate change rather than a rider on this one.

**On `examples/chanmig/main.cpp:574`** — confirmed, `0x57` has the I/G bit set, so `is_unicast()` refuses and `chanmig --role drone` exits 1 on every generation since #406. Not caused here and I have left it alone; happy to send it as its own PR if you would like, since it is a live breakage rather than a docs issue. (The `0x57` in my selftest is the deliberate group-address example, which is the same fact from the other side.)

I have not resolved your threads — they are yours to close once you are satisfied.

josephnef (dismissed):
Re-verified ee0d36e here: selftest passes, and on the bench (RTL8733B `0bda:b733` responder, RTL8822CU soliciting MCS3 on ch36, ~1700 frames/cell, scored from the solicitor's CCX `tx.report`):

| cell | ack_rate | retries_mean |
|---|---|---|
| armed, no disarm | 1.00 | 0.04 |
| armed, `DEVOURER_ACK_DISARM_AFTER_MS=2000` | 0.00 | 12.0 |
| armed, `DEVOURER_ACK_DISARM_AFTER_MS=0` | 0.00 | 12.0 |

Every finding is addressed. One thing to follow up, not blocking: the rxdemo disarm timer counts from before `Init()`, while the arm happens inside bring-up on every generation. At 0 ms the log shows "disarming" printed before "armed"; the disarm still landed after the arm only because `Rtl8733bDevice::Init` holds `_reg_mu` across the whole bring-up, so `ClearAckResponder` blocked until the RX ring was up. Jaguar1/2/3 have no bring-up-wide lock, so a timer shorter than their bring-up would disarm before the arm and the `disarmed` phase would read a false "disarm broken". Anchoring the countdown on the arm (first RX callback, or a loop-entered flag) would make the phase portable. Two nits alongside: `disarm_ack_responder()` returns the gate result after logging "disarmed", so an inert-gate failure makes the callers log "hardware state UNKNOWN" although the identity read back; and the four-number block is now quoted verbatim in four files, where `AdapterCaps.h` could own it.

snokvist:
Pushed `9a03e75` to address the non-blocking findings from the latest review and the additional adversarial/HW pass.

- The disarm countdown now belongs to `Rtl8733bDevice` and starts only after its locked bring-up/arm/channel/RX configuration. A zero-ms run can no longer race ahead of `SetAckResponder` on another backend.
- `disarm_ack_responder()` now always performs the authoritative MACID readback. A false transfer status with a successful readback is diagnosed, while a verified identity move no longer becomes the contradictory caller result `hardware state UNKNOWN` merely because the inert gate readback failed.
- The duplicated four-number explanation is removed from the selftest/CLAUDE text; `AdapterCaps.h` owns the truth table.
- The matrix is now fail-closed: exact readiness/disarm log waits, responder liveness checks, TX timeout-status validation, minimum submissions, analyzer failure propagation, and a real RTL8733B USB-ID guard on the fourth row. It also exits the responder normally after each phase.

Final rerun on the connected `f72b` responder with RTL8812AU solicitor (MCS3, retry limit 12): armed 751/751 reported ACKed, retargeted 754/754, never-armed 0/741, same-process disarmed at zero ms 0/758. Retry means were 0.00, 0.01, 12.0, 12.0; report coverage was 99.9-100%. The full 55-test suite, default build, variant builds, ShellCheck, and whitespace check pass.

The review also uncovered a real but separate Jaguar1/RTL8812AU defect: its gate-only clear logs disarmed while the adapter continues ACKing the old responder MAC. I have kept that out of this die-scoped change and will send it as a separately tested follow-up. I am not claiming BlockAck disarm from the normal-ACK evidence above.

snokvist:
One more fail-closed issue found in the pre-follow-up adversarial review and fixed in `ff0a45f`: `ack_txreport_analyze.py` displayed `report_coverage` but did not include it in `capability_ok`. One matching report among hundreds of submitted frames could therefore pass either an ON or OFF cell.

The verdict now requires 80% coverage by default (`MIN_REPORT_COVERAGE` / `--min-coverage`), and the analyzer selftest pins both sparse-ON and sparse-OFF rejection. The Jaguar1 hardware runs that motivated the audit had 89.5-95.9% coverage and continue to clear this threshold. The PR description has been corrected accordingly.

snokvist:
The second independent adversarial pass found another verdict hole, fixed in `8265b1b`: OFF used `retries_max >= limit`, so one retry-12 outlier plus 99 retry-0 failures passed as “pinned.” The analyzer now emits `retry_pin_rate` and requires 90% by default; a regression vector pins the one-outlier case. Numeric matrix controls are also validated before any hardware process starts, closing the nonnumeric `READY_TIMEOUT` unbounded-wait case.

The existing hardware rows are genuinely pinned (`retry_pin_rate=1.0` when rescored), so this tightens the oracle without changing the measured result.

snokvist:
One final harness-safety fix is in `974b304`: cleanup and graceful stop now signal only the responder PID started by this matrix. The previous name-wide `pkill` could terminate unrelated `rxdemo`/`txdemo` sessions on a shared bench. ShellCheck and syntax checks pass, and the exact-PID lifecycle completed cleanly in both final four-phase RTL8812AU hardware reruns.

snokvist:
The Windows CI failure was a real portability issue in the new timer: `std::min(left, quantum)` was expanded by the Windows `min` macro. Fixed in `b5f3e00` with a macro-proof comparison, and made the new `numeric_limits::max` call macro-proof at the same time. The Linux/default and reduced builds remain green; CI has restarted on this commit.

josephnef (approved):
Re-approving at b5f3e00 (the previous approval was auto-dismissed by the push of 9a03e75, which also dropped the armed auto-merge).

The five commits since ee0d36e address the timer-ordering follow-up from the last review: the countdown now lives in `Rtl8733bDevice::Init` and starts after the locked bring-up, rxdemo refuses the knob on any other generation, and `disarm_ack_responder()` returns true on a verified identity move instead of the inert gate's status. The matrix waits on log lines instead of fixed sleeps and rejects sparse / unpinned verdicts.

Re-verified on hardware from a build of b5f3e00, `tests/ack_txreport_matrix.sh` with the RTL8733B (0bda:b733) as responder and an RTL8822CU (0bda:c812) transmitter, ch36, MCS3, ~1.7k frames per phase:

| phase | ack_rate | retries_mean | retry_pin_rate | coverage | verdict |
|---|---|---|---|---|---|
| on | 1.00 | 0.02 | – | 0.971 | ok |
| retarget | 1.00 | 0.03 | – | 0.972 | ok |
| off | 0.00 | 12.0 | 1.00 | 0.917 | ok |
| disarmed | 0.00 | 12.0 | 1.00 | 0.931 | ok |

Responder log for the `disarmed` phase shows the intended order: armed → RX ring submitted → "disarming ... 2000 ms after completed bring-up" → disarmed. `AckResponderSelftest` passes.

PR #411, opened 2026-09-06
jaguar1: restore port identity when disarming 8812 ACKs

snokvist:
## Problem

Jaguar1's existing `ClearAckResponder()` only wrote `net_type=NoLink`. On the
reference RTL8812AU (`0bda:8812`) that write read back successfully, but the
port continued sending SIFS ACKs for the old responder MAC: 1946/1946
soliciting CCX reports were still ACKed.

This is not the RTL8733B gate model: a never-armed RTL8812AU did not answer on
its own MAC (0/1074), but arm-to-own-MAC followed by the old gate-only clear
did (1120/1120). The clear therefore needs both NoLink and a move away from the
armed MAC.

## Change

- On `CHIP_8812`, snapshot the live port-0 MACID/BSSID before the first arm and
  restore/readback-verify that exact identity during clear and failed-arm
  rollback.
- Reject zero/group rollback identities and a responder equal to the captured
  MACID; either case would make a later retarget unsafe or ineffective.
- Verify successful RTL8812 arms, and roll back if the arm readback fails or a
  later `Init`/`InitWrite` step throws.
- Bind delayed disarm and `Init`/`InitWrite` rollback to the exact completed
  ACK arm. Clearing or replacing that arm invalidates its token, so stale
  cleanup cannot disarm a newer responder.
- Serialize ACK/beacon port-0 transactions. A beacon claims the port before
  its first register mutation, including a partially failed setup, and ACK
  arming is refused until `StopBeacon()` verifies the beacon-function,
  TX-beacon, and net-type stop controls and releases it. An unverifiable stop
  retains ownership and recovery state for retry.
- Extend the backend-owned live-disarm hook and fail-closed CCX matrix to the
  measured `0bda:8812` responder.

## Hardware evidence

Final tree, RTL8812AU `0bda:8812` responder, `0bda:c812` solicitor, channel 36,
MCS3, retry limit 12, 12 seconds per phase:

| phase | submitted / reports | ACK rate | mean retries | coverage | retry-12 pin rate |
|---|---:|---:|---:|---:|---:|
| armed MAC1 | 1150 / 1100 | 1.00 | 0.03 | 0.957 | n/a |
| armed MAC2 | 1173 / 1123 | 1.00 | 0.08 | 0.957 | n/a |
| never armed | 1164 / 1069 | 0.00 | 12.0 | 0.918 | 1.00 |
| arm MAC1, same-process zero-ms clear | 1204 / 997 | 0.00 | 12.0 | 0.828 | 1.00 |

The zero-ms countdown is created by the Jaguar1 backend only after completed
bring-up and verified arm; it cannot win a race against `Init()`. A separate
2000-ms run also produced 0/1077 ACKed with retries pinned at 12 after clear.
The final own-MAC negative test refuses the arm before the responder becomes
active.

The stale-timer race was also exercised directly on this tree: arm A scheduled
an 8000-ms disarm, A was explicitly cleared and replaced by arm B after four
seconds, and A's old timer reported a stale/no-op at its deadline. Arm B then
ACKed 377/377 reports (ACK rate 1.00, mean retries 0.04, coverage 0.883) before
its own verified clear. The normal matching-timer path is covered by the
0/997 same-process disarm cell above.

The ACK-rate result establishes MACID's behavioral role. BSSID is restored and
read back as defensive port-state cleanup; this PR does not claim BSSID gates
ACK behavior.

The all-die port-ownership guard was separately exercised on the RTL8812AU:
`StartBeacon()` refused while the ACK responder owned port 0, succeeded after
verified ACK clear, and a subsequent `ClearAckResponder()` while the beacon
owned the port left it airing. An independent RTL8822CU witness captured 48
valid hardware beacons after that clear. No RTL8821AU/RTL8814AU was connected
for this rerun; their ownership path is the same backend state and lock, while
their existing gate-only responder clear remains unchanged.

The final verified-stop tree was then exercised in the opposite direction.
The RTL8812AU aired 55 valid beacons before `StopBeacon()` returned true; the
RTL8822CU witness saw zero beacons after a 250-ms drain window. In the same
RTL8812AU process, ACK ownership was acquired immediately afterward and the
RTL8822CU solicitor measured 372/372 successful reports (ACK rate 1.00, mean
retries 0.02, coverage 0.882). The responder then completed verified
MACID/BSSID restoration. This hardware result covers the normal stop path;
the transport-failure branches are register-model failure-injection tests.

## Scope and validation

- Measured on RTL8812AU only. The implementation follows the shared
  `CHIP_8812` path (including the RTL8811AU cut), but RTL8811AU was not
  separately measured.
- RTL8814A and RTL8821A retain their existing gate-only clear; HalMAC and
  RTL8733B behavior is unchanged.
- This proves normal immediate-ACK disarm only. It makes no BlockAck-disarm
  claim.
- Default build and 55/55 CTest pass; Jaguar1-only, RTL8733B-only, and
  Jaguar3-only builds pass; ACK analyzer selftests, `AckResponderSelftest`,
  ShellCheck, bash syntax, and `git diff --check` pass.
- Headless tests cover exact MACID/BSSID snapshot/restore and readback,
  zero/group/equal rollback rejection, false-return and thrown-write failure
  injection, all-four-register restore attempts, independent attempts of all
  three beacon stop operations, fail-closed readback, and false-but-landed
  transfer handling.


src/jaguar1/RtlJaguarDevice.cpp
```diff
@@ -768,14 +817,57 @@ bool RtlJaguarDevice::SetAckResponder(const devourer::MacAddr &mac) {
                    "Jaguar1", mac.bytes[0]);
     return false;
   }
+  if (_eepromManager->version_id.ICType == CHIP_8812) {
+    const bool had_restore_identity = _ack_restore_identity.has_value();
+    if (_port0_beacon_claimed) {
```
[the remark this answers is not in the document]
snokvist: Fixed in 716d1b2. ACK ownership is now an explicit `_port0_ack_claimed` state shared by every Jaguar1 die, independent of the CHIP_8812-only restore snapshot. It is claimed before the first arm mutation, retained after an unverified rollback, and released only after verified clear. `StartBeacon` and `SetAckResponder` now reject the opposite owner on every Jaguar1 IC, and a clear without ACK ownership leaves a beacon-owned net_type untouched.

Hardware verification preceded this update on the connected RTL8812AU: beacon-over-ACK was rejected; after verified ACK clear the beacon claimed port 0; a subsequent ACK clear logged no configured responder, and an independent RTL8822CU witness captured 48 valid hardware beacons after that clear. The final 12-second matrix also passed: armed 1112/1112 ACKed, retarget 1094/1094, never-armed 0/1029, same-process zero-ms disarmed 0/1010, with coverage 0.837–0.957 and both OFF retry-pin rates 1.0. RTL8821AU/RTL8814AU were not connected; the ownership checks are common backend state, while their existing gate-only clear remains unchanged.

src/jaguar1/RtlJaguarDevice.cpp
```diff
@@ -611,11 +654,13 @@ bool RtlJaguarDevice::StopBeacon() {
   _device.rtw_write8(0x0102, static_cast<uint8_t>(nt & ~0x03u));
   _bcn_mpdu.clear();
   _bcn_interval_tu = 0;
+  _port0_beacon_claimed = false;
```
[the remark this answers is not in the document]
snokvist: Fixed in 0941ec0. `StopBeacon()` now holds the port mutex while a Jaguar1-specific stop helper independently attempts EN_BCN_FUNCTION-off, StopTxBeacon, and net_type=NoLink, then independently reads back all three behavior bits. The beacon claim, retained payload, and timing are released only when every control verifies inactive; otherwise the method returns false with the port still beacon-owned so ACK arming remains blocked and stop can be retried. A false transport status is accepted only when readback proves the hardware state. Failure-injection coverage exercises successful stop, false/no-store, throwing writes with all three operations attempted, and false-but-landed writes.\n\nHardware verification preceded this update on the connected RTL8812AU: it aired 55 valid beacons before the verified stop, and an independent RTL8822CU witness saw zero after a 250 ms drain window. The same RTL8812AU process then acquired ACK ownership; the RTL8822CU solicitor measured 372/372 successful reports (ACK rate 1.00, mean retries 0.02, coverage 0.882), followed by verified responder identity restoration. RTL8821AU/RTL8814AU were not connected; the stop helper and ownership release are common to every Jaguar1 die.

src/jaguar1/RtlJaguarDevice.cpp
```diff
@@ -1409,7 +1586,39 @@ void RtlJaguarDevice::Init(Action_ParsedRadioPacket packetProcessor,
...
+                    std::chrono::milliseconds(25));
+            std::this_thread::sleep_for(left < quantum ? left : quantum);
+          }
+          if (stop.stop_requested())
+            return;
+          _logger->info(
+              "DEVOURER_ACK_DISARM_AFTER_MS: disarming Jaguar1/CHIP_8812 "
+              "ACK responder {} ms after completed bring-up",
+              delay_ms);
+          ClearAckResponder();
```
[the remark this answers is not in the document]
snokvist: Fixed in 401dbf9. Each completed ACK arm now receives an active token. The token is invalidated before a replacement starts, assigned only after the replacement fully succeeds, retained when a clear cannot be verified, and checked under the port-0 mutex by delayed disarm. Init/InitWrite exception rollback is bound to the same token, and the timer rechecks cancellation after acquiring the mutex. Hardware exercised the exact race: arm A with an 8 s timer, explicitly clear A and arm B after 4 s, then cross A’s deadline. The old timer logged a stale/no-op and B still ACKed 377/377 reports (mean retries 0.04, coverage 0.883), followed by a verified explicit clear. A separate 12 s matching-timer matrix confirmed normal disarm at 0/997 ACKed with retries pinned at 12.

snokvist:
/review\n\nPlease review the current 401dbf9 head. In particular, please re-check delayed-disarm/rollback token ownership and failed-clear semantics; the exact stale-timer replacement race and normal matching-timer disarm were both rerun on RTL8812AU hardware before this update.

src/IRtlDevice.h
```diff
@@ -426,7 +426,9 @@ class IRtlDevice {
    * process does NOT silence it (bench-bitten: a killed probe's beacon kept
    * airing and contaminated the next test's witness) — so any beaconing
    * session that ends without a device power-cycle must call this. Idempotent;
-   * returns false when no beacon was active. */
+   * returns false when no beacon was active or shutdown could not be verified.
```
josephnef: This base-contract sentence now promises a verified stop, but only Jaguar1 verifies: `RtlJaguar2Device::StopBeacon` and `RtlJaguar3Device::StopBeacon` still return true unconditionally after their writes. Either scope the sentence to Jaguar1 ("Jaguar1 additionally reads back the three stop controls and returns false if any is still active") or leave the base contract as it was.

src/AdapterCaps.h
```diff
@@ -180,9 +180,18 @@ struct AdapterCaps {
...
    *
    * The `on`, `retarget`, and legacy `off` rows establish arming, retargeting,
-   * and a never-armed control. Only the RTL8733B-scoped `disarmed` row arms and
-   * clears within one process; it supports the RTL8733B live-disarm claim only
-   * and establishes nothing about disarm on other dies.
+   * and a never-armed control. A backend-owned `disarmed` row supports only
+   * the live-disarm claim for the responder used in that run. On the reference
+   * RTL8812AU, that cell found that the old gate-only clear left 1946/1946
+   * soliciting reports ACKed. Restoring the captured pre-arm MACID produced
+   * 0/1052 ACKed with retries pinned at 12; the never-armed counterpart was
```
josephnef: Three different count sets now describe the same claim: this header says 0/1052 and 0/1033, the PR body says 0/997 and 0/1077, and `docs/scheduled-mac.md` points here as the owner of "the exact counts". They are different runs and all agree, but a caps header comment is turning into a measurement log. Keep one set here (or point at the doc that carries the table) so the next edit does not have to reconcile them.

src/jaguar1/RtlJaguarDevice.cpp
```diff
@@ -768,30 +836,148 @@ bool RtlJaguarDevice::SetAckResponder(const devourer::MacAddr &mac) {
...
+  if (_eepromManager->version_id.ICType == CHIP_8812) {
+    const bool had_restore_identity = _ack_restore_identity.has_value();
+    if (!_ack_restore_identity) {
+      devourer::ack::PortIdentity identity;
+      if (!devourer::ack::snapshot_port_identity(_device, identity)) {
+        _logger->error("Jaguar1/CHIP_8812: ACK responder cannot be armed: "
+                       "the current port identity could not be read");
+        return false;
+      }
+      if (!devourer::ack::has_safe_restore_mac(identity)) {
```
josephnef, line 853: Behaviour change to state in the PR body: before this PR an 8812 with a zero/group MACID at arm time still armed; now the arm is refused. The refusal is right (restoring zero stops TX scheduling), but it is a new failure mode for a caller with an unprogrammed EFUSE MAC.

josephnef (dismissed):
Reviewed at 401dbf9. No correctness findings; the three earlier Qodo items (unverified beacon release, ownership only on CHIP_8812, stale timer clearing a new arm) are all closed at this head by the verified stop, the all-die claim flags and the arm-generation token. Two doc nits inline, neither blocking.

What I checked:

- ASan+UBSan build of the branch, ctest 55/55.
- Token/rollback ordering in `Init`: the disarm `jthread` is declared after the scope-exit guard, so it is stopped and joined before the rollback runs, and a fired timer has already invalidated the token the rollback checks. A failed re-arm invalidates the previous token before its first write, so no stale timer can reach the replacement. Claim flags are retained on every unverified path.
- Hardware regression on the non-8812 Jaguar1 path, since the PR touches every die's arm path. No RTL8812AU was on the bench, so the 8812 cell itself is not reproduced here; an RTL8821AU responder (`0bda:a81a`) against an RTL8822CU solicitor, channel 36 / MCS3 / retry limit 12:

| phase | reports | ACK rate | mean retries |
|---|---:|---:|---:|
| on | 1569 | 1.00 | 0.05 |
| retarget | 1549 | 1.00 | 0.04 |
| off (never armed) | 1526 | 0.00 | 12.0 (pin rate 1.00) |

  The `disarmed` phase reports SKIP on the 8821A as designed.

One behaviour change worth a line in the PR body: on CHIP_8812 an adapter whose MACID reads zero or group at arm time is now refused outright (`has_safe_restore_mac`), where previously the arm went ahead. Correct, since restoring zero stops TX scheduling, but it is a new failure mode a caller can hit.

src/AdapterCaps.h
snokvist: Fixed in 19b604a and carried through 534ffad. `AdapterCaps.h` now states the behavioral result without maintaining a competing run log, and `docs/scheduled-mac.md` owns one canonical table containing the gate-only, verified-restore, never-armed, and same-address adversary counts.

src/jaguar1/RtlJaguarDevice.cpp
snokvist: Addressed in the PR body. The Change section now calls this out as caller-visible: on `CHIP_8812`, `SetAckResponder()` returns false for a zero/group live pre-arm MACID (including an unprogrammed-EFUSE case), or when it equals the responder; it also states that the old path attempted the arm and why refusal is intentional.

src/IRtlDevice.h
snokvist: Fixed in 19b604a and carried through the current 534ffad head. The base `IRadio::StopBeacon` contract again says only that false means no beacon was active; the stronger three-control readback guarantee and retry requirement are now explicitly scoped to Jaguar1.

snokvist:
@josephnef The requested fixups are now on current head `534ffad`, and the branch is merged with current `master` (`90e1cad`). The sole content conflict was the #415 interface rename beside `ScheduleAckResponderDisarmForTest`; the resolution keeps the Jaguar1 hook and uses the new `IRadio`/`IRtlRadio` contract. The two documentation notes are fixed in `19b604a`, the caller-visible zero/group/equal-MACID refusal is now explicit in the PR body, and all three threads have replies and are resolved.

Verification on the integrated tree: ASan+UBSan build, CTest 57/57, `git diff --check`, ShellCheck/bash syntax, and an independent review of ACK/beacon ownership, generation-token ordering, failed-clear retention, and the interface merge. No findings. Hardware status for this integration pass is code-level only; the existing PR hardware evidence is unchanged.

GitHub dismissed your previous approval after the head update, so please re-review/approve `534ffad` when convenient.

josephnef (approved):
Re-approving at `ae8ef3f` (my earlier approval was auto-dismissed by the new commits).

All three of my notes are properly closed: `IRadio::StopBeacon`'s base contract is back to "false means no beacon was active" with the three-control readback scoped to Jaguar1, `AdapterCaps.h` now states the behavioural result and hands the counts to `docs/scheduled-mac.md` (which grew one canonical table instead of three competing ones), and the zero/group/equal-MACID refusal is called out as caller-visible in the PR body. The #415 rename resolution is correct — the tightened `StopBeacon` doc landed in the renamed `src/IRadio.h`, not orphaned in the deleted `IRtlDevice.h`.

Integration on current master (which moved under this branch — #420–#423 landed today): ASan+UBSan build clean, **ctest 63/63**.

## Hardware: the all-die half, on the die that was not re-tested

The identity restore is CHIP_8812-only, but two changes here apply to **every** Jaguar1 die, and the PR body says no RTL8821AU/RTL8814AU was connected for the final rerun. Those are the ones with regression risk: a die whose `0x0550`/`0x0422`/`0x0102` readbacks behaved differently would newly return `false` from a previously-true `StopBeacon()` and then **retain port ownership**, permanently blocking ACK arming.

Measured on the one Jaguar1 on this bench — a TP-Link Archer T2U PLUS `2357:0120`, self-identified by the library as `RTL8821A (8821A), generation=jaguar1`. Whole probe built and run under ASan+UBSan; no sanitizer report. Witness is an independent RTL8812CU (`0bda:c812`) running `rxdemo`, ch36.

Ownership state machine, 9/9:

| check | result |
|---|---|
| `SetAckResponder(MAC1)` on a free port | true |
| `StartBeacon` while ACK owns port 0 | **refused** |
| `StartBeacon` after `ClearAckResponder` | true |
| `SetAckResponder` while beacon owns port 0 | **refused** |
| `StopBeacon` (three controls verified) | true |
| `StopBeacon` again (idempotent) | false |
| `SetAckResponder` after verified `StopBeacon` | true — port genuinely released |
| `StartBeacon` again after final ACK clear | true |
| final `StopBeacon` | true |

And the verified stop really does silence the air, two phases with a fresh witness each and the DUT held **open** across the window (holding it open is the point — Jaguar1 teardown powers the chip down and would stop the beacon by itself, proving nothing about `StopBeacon`):

| phase | witness |
|---|---|
| `StartBeacon`, hold open 25 s | 200 beacon hits (~8/s, matches 100 TU) |
| `StartBeacon`, air 600 ms, verified `StopBeacon`, hold open 25 s | 6 hits — **zero after the stop** |

One method note in case it is useful elsewhere: `rx.txhit` is sampled — `examples/rx` emits every hit up to 10 and only every 100th after that, and the event carries no timestamp — so a before/after total cannot separate the windows. My first attempt read `10` for the stop phase, which only means "somewhere in [10, 100)" and proves nothing. Sizing the pre-stop airing window so a *passing* run lands below 10 puts it in the band where every hit is individually reported and the count is exact, while a beacon that kept airing would run past 10 and hit the 100 sample. 600 ms gives ~6, and 6 < 10 is exact.

Also confirmed that `DEVOURER_ACK_DISARM_AFTER_MS` correctly refuses on this non-CHIP_8812 die (`rxdemo` exits 1 naming the two supported paths), so the measurement hook cannot manufacture false evidence on an unmeasured die.

## What this does not cover

The 8812-specific captured-identity snapshot/restore — the actual subject of the PR — is **not** reproduced here. There is no RTL8812AU on this bench (`0bda:c812` is an RTL8812CU, a Jaguar3), and `tests/ack_txreport_matrix.sh` records the 8821AU as a known-bad ACK responder ("an armed 8821AU never closed the loop while the 8814AU ACKs ~100%"), so no ACK-rate cell is possible on this hardware either. That half rests on your bench evidence, which reads consistently and is scoped correctly in the PR body.

PR #412, opened 2026-09-06
mt7612u: a MediaTek backend, measured — is it in scope?

snokvist:
Opening this to ask a scope question, not to merge: **does a MediaTek backend belong in this project at all?** Every generation here is Realtek, and MT7612U is not. I would rather ask against measurements than against a proposal, so the work is on the branch and `docs/mt7612u.md` carries the numbers, the methods and the limits.

If the answer is no, that is a fine answer and the branch can be closed — the measurements are still recorded for anyone who asks later.

## What changed

`src/mt7612u/` — a standalone C library (~4600 lines including the harness) plus a per-gate bringup tool, and `docs/mt7612u.md`.

**`CMakeLists.txt` is untouched.** There is no `IRtlDevice` implementation, no `WiFiDriver` dispatch, no `DeviceConfig` plumbing and no `ctest` cell. Nothing in the shipped library changes; CI sees a docs-and-new-directory diff. The subtree builds on its own with `make -C src/mt7612u`. That integration work is deliberately not done, because doing it before the scope question is answered would be wasted either way.

## Why a MediaTek port is small

On the Realtek generations the host programs the synthesizer and carries the PHY tables. Here the RF plane lives behind the MCU — a channel change is one 8-byte `CMD_SWITCH_CHANNEL_OP` plus a firmware calibration burst, and there are no RF register tables to reimplement. That is also exactly why it can never hop fast.

## What is measured

All on one MT7612U (`0e8d:7612`, `MT_ASIC_VERSION` `0x76120044`, 2T2R, SuperSpeed) against an RTL8812AU witness running this project's own `rxdemo`/`txdemo`.

**TX rate is authoritative, unconditionally.** Three rates, three exact matches; per-frame alternation 300 sent → **229 received, 229 aired the rate their own index called for, 0 mismatches**. The hardware rate LUT cannot override the descriptor:

| `MT_TXWI_FLAGS_TX_RATE_LUT` | frames | aired |
|---|---|---|
| clear | 84 | 83 × MCS7, 1 × MCS6 |
| **set** | 93 | 93 × MCS7 |

LUT loaded with OFDM 6 Mbps and read back; setting the flag mt76 defines and never sets changed nothing.

**TX power** — all nine registers identical to what the kernel driver programs for ch149 (`MT_TX_PWR_CFG_0..4/7/8/9`, `MT_TX_ALC_CFG_0` = `2f2f171a`), checked against a `usbmon` capture.

**A-MPDU works on injected frames** — the opposite of the Jaguar1 result in `docs/aggregation.md`:

| arm | frames | `paggr=1` |
|---|---|---|
| no AMPDU flag (control) | 352 | **0** |
| AMPDU + `QSEL_MGMT` | 326 | **326** |

| frame bytes | AMPDU off | AMPDU on | gain |
|---|---|---|---|
| 200 | 7.01 Mbit/s | 15.50 Mbit/s | **2.21×** |
| 1400 | 34.03 Mbit/s | **44.55 Mbit/s** | 1.31× |

**Hardware ACK responder** — no separate responder-address register exists, so arming retargets the port identity with `MT_AUTO_RSP_EN` gating; clear moves the identity back, following the finding in #410 that closing the gate alone is not enough.

| responder | frames the stimulus radio received |
|---|---|
| off (control) | **0** |
| armed | **3500+, every one `len` 14** |

**USB bulk aggregation** — `MT_TXD_INFO_NEXT_VLD` chains blocks inside one bulk-OUT transfer. mt76 never sets it, so this is not a port and needed on-air proof: 400 frames in 25 transfers, 352 aired.

Also measured: 40 MHz (242/242 frames at `bw=1`), TSF (200231 µs over a 200000 µs sleep), monitor RX decoding CCK/OFDM/HT with per-chain RSSI, radiotap `send_packet`.

**A register-stream diff against the kernel driver's own probe**: 522 kernel EP0 writes vs 521 ours, 376 common addresses, one final-value mismatch (beacon config, which we skip), six kernel-only addresses (all beacon config), and **zero addresses we write that the kernel does not**.

## Two results stated against interest

**The async rings bought no throughput.** At saturation sync and async both sit at 3040 fps / 34 Mbit/s with an identical 0.329 ms mean submit, because that is the airtime of one frame. What they bought is the callback RX path, concurrent TX+RX on one handle, and 3× lower submit latency below saturation (21 µs vs 65 µs). A-MPDU, not USB parallelism, is what lifted 34 → 44.55 Mbit/s.

**The first ACK-responder attempt returned INCONCLUSIVE, not a pass.** It used this project's retry-collapse method, but `txdemo` injects without retries, so there was nothing to collapse. The gate says so rather than reporting success; the passing result above uses a different observable.

## What it cannot do

| channel switch | measured, mean of 8 |
|---|---|
| full, with the firmware calibration burst | 526 ms |
| calibration skipped | 48 ms |
| this project on Realtek, for reference | 0.5–2.5 ms |

Our implementation has headroom, but the floor is four MCU round trips over bulk endpoints plus firmware time — 10–20 ms at best. **FHSS and per-packet hopping are out of reach for this part**; seconds-scale migration is not. **Narrowband 5/10 MHz has no encoding**: `MT_RATE_BW` is two bits with three defined values.

Also worth knowing: **unicast injection is a 40× cliff** (3037 → 75 fps). The MAC arms an ACK timeout for a peer that never answers, and neither clearing `txwi.ack_ctl` REQ nor a QoS No-Ack policy prevents it. A one-way link must use broadcast.

## Counterparts

Stated because the numbers above are uniformly favourable.

- **One physical unit, one sample.** No second MT7612U, no second board revision.
- **One witness generation.** Every on-air number is an RTL8812AU running this project's `rxdemo`. `paggr`, `bw` and `rate` are that implementation's reading, not an independent instrument.
- **TX power was verified against the kernel's registers, not radiated power.** No spectrum analyser, no power meter. The claim is "identical to what mt76 programs", nothing more.
- **The RX gain correction has never done anything.** This EEPROM carries no gain calibration, so every term is zero and that path is unexercised.
- **The ACK identification rests on length, the 1:1 count and a zero control arm.** The RA bytes inside those ACKs were not read — the dump path omits control-frame bodies.
- **No cold boot was ever tested.** No hub on the test host supports per-port power switching, and `reset_wlan` + `power_on` demonstrably does *not* clear the firmware-running bit.
- **Nothing here runs in CI.** No `ctest`, no sanitizer build, no lifecycle soak.
- **80 MHz, VHT on air and NSS=2 are unexercised.**

One more, on the code rather than the measurements: enabling MAC RX without draining the bulk-IN endpoint wedges the chip below USB level, unrecoverable by `libusb_reset_device()`, the `authorized` toggle or the kernel driver — only a physical replug. Fixed by never enabling RX for a caller that will not drain it, plus an endpoint flush (20 consecutive cycles clean afterwards against a death after ~5 before). **Two things changed at once**, so that run does not attribute the wedge to one of them.

## Verification

- `make -C src/mt7612u` clean at `-Wall -Wextra`.
- Every gate listed in `src/mt7612u/README.md` run on hardware; the numbers above are those runs.
- The library target and CI are untouched by construction — the diff adds a directory and a doc, and modifies nothing.

## Open list

If this is in scope, the work I would continue with, in order:

1. `IRtlDevice`, `WiFiDriver` dispatch, `DeviceConfig`, `CMakeLists.txt`, `ctest` cells.
2. `mt76x2_phy_tssi_compensate()` — without it output power drifts with die temperature.
3. Cold-boot verification on a host with switchable USB power.
4. A second sample and a second witness generation.
5. 80 MHz; VHT and NSS=2 on air.
6. Retune tuning (batch registers via `CMD_RANDOM_WRITE`, drop the inter-command sleep) — worth doing only if 10–20 ms is useful to someone.
7. Whether the single MCS6 frame in the rate-LUT control arm is a witness artefact or a real fallback. Unexplained.


src/mt7612u/include/mt7612u/mt7612u.h
```diff
@@ -0,0 +1,171 @@
...
+};
+
+struct mt7612u_dev;
+
+/*
+ * open: claims the interface, detaching mt76x2u if it holds it, then runs the
+ * full power-on, firmware load and MAC/PHY init. fw_dir may be NULL for the
+ * system default. Returns NULL on failure; err (optional) receives a message.
+ */
+struct mt7612u_dev *mt7612u_open(const char *fw_dir, const char **err);
```
josephnef, line 110: `mt7612u_open`, `_close`, `_keep_detached`, `_set_channel`, `_set_chainmask`, `_start`, `_stop`, `_asic_version` and `_mac_addr` are declared here but no `.c` file defines them. The harness works only because it includes `internal.h` and calls `mt_open` / `mt_init_hardware` directly. Anything linking against this header gets undefined references, so the "standalone C library" cannot be used as one yet. (`set_channel` also takes `unsigned` here and `uint8_t` internally.)

src/mt7612u/rx.c
```diff
@@ -0,0 +1,111 @@
...
+		pad = 2;
+
+	len = (int)info->mpdu_len;
+	if (len > n - MT_DMA_HDR_LEN - MT_RXWI_LEN - pad)
+		len = n - MT_DMA_HDR_LEN - MT_RXWI_LEN - pad;
+	if (len < 0) return 0;
+
+	*frame = buf + MT_DMA_HDR_LEN + MT_RXWI_LEN;
+	/* Fold the L2 pad out by moving the header down over it. */
+	if (pad && len > 24) {
```
josephnef: This folds a fixed 24 bytes over the L2 pad. L2PAD is only set when the header is not 4-aligned, which is QoS data (26 bytes) or 4-address (30 bytes). With a 26-byte header the QoS Control field at offsets 24-25 is overwritten by the two pad zeros, so every A-MPDU / QoS frame reaching the callback has TID and ack-policy zeroed. Those are exactly the frames the aggregation and ACK numbers are about. `tx.c` already computes the real header length from FC; reuse it here.

src/mt7612u/tx.c
```diff
@@ -0,0 +1,211 @@
...
+ * and A-MPDU only ever carries QoS data, so getting this wrong breaks
+ * aggregation before the hardware ever sees it. */
+static int hdrlen_from_fc(const uint8_t *f)
+{
+	unsigned fc = (unsigned)f[0] | ((unsigned)f[1] << 8);
+	unsigned type = (fc >> 2) & 3, stype = (fc >> 4) & 0xf;
+	int len = 24;
+
+	if (type == 1)                       /* control */
+		return ((stype == 0xb) || (stype == 0xa)) ? 16 : 10;
```
josephnef: Only CTS and ACK are 10-byte control frames. BlockAckReq (0x8), BlockAck (0x9) and CF-End / CF-End+CF-Ack (0xe/0xf) are 16 bytes like RTS and PS-Poll. With this table the pad is inserted at offset 10 inside a BAR/BA, so the frame airs corrupted.

src/mt7612u/async.c
```diff
@@ -0,0 +1,213 @@
...
+	pthread_mutex_unlock(&a->tx_lock);
+
+	for (int spins = 0; a->rx_inflight && spins < 100; spins++)
+		mt_usleep(10000);
+
+	a->running = 0;
+	pthread_join(a->evt, NULL);
+
+	for (int i = 0; i < MT_TX_RING; i++)
+		if (a->tx[i]) libusb_free_transfer(a->tx[i]);
```
josephnef, line 248: `mt_async_stop` cancels only the RX transfers. TX transfers are never cancelled; after the bounded wait above (200 x 10 ms) `running` is cleared, the event thread joined, and every transfer freed regardless of `tx_inflight` / `rx_inflight`. On a wedged chip, the case `docs/mt7612u.md` itself describes, TX URBs never complete, so this frees transfers libusb still owns. Cancel TX too and only free once inflight counts reach zero. Related: the `fail:` path at line 102 joins a zeroed `pthread_t` if `libusb_alloc_transfer` fails before `pthread_create`, and `rx_inflight` / `running` are plain `volatile int` shared across threads.

src/mt7612u/usb.c
```diff
@@ -0,0 +1,232 @@
...
+}
+
+uint32_t mt_rr(struct mt7612u_dev *d, uint32_t addr)
+{
+	uint8_t req = rd_req(addr), b[4] = { 0 };
+	uint32_t a = addr & ~MT_VEND_TYPE_MASK;
+
+	if (mt_vendor_req(d, req, REQ_IN, (uint16_t)(a >> 16), (uint16_t)a,
+	                  b, sizeof b) != (int)sizeof b)
+		return ~0u;
```
josephnef: `~0u` on failure is consumed unchecked by `mt_rmw`, `set_wlan_state`, the fw handshake and every `mt_set` / `mt_clear` caller, so one exhausted EP0 read during bring-up writes `0xFFFFFFFF | val` into `MT_WLAN_FUN_CTRL` / `MT_MAC_SYS_CTRL` / BBP AGC. Return a status and make the RMW helpers refuse to write after a failed read. Also: the retry loop retries every error except NO_DEVICE 10x with a 1 s timeout each, so a dead read stalls 10 s and `mt_poll`'s microsecond `timeout_us` is meaningless.

src/mt7612u/usb.c
```diff
@@ -0,0 +1,232 @@
...
+		}
+		libusb_close(d->h); d->h = NULL;
+	}
+	if (d->ctx) { libusb_exit(d->ctx); d->ctx = NULL; }
+}
+
+/* Block write, as mt76u_copy(): one MULTI_WRITE per batch, wValue 0.
+ * The kernel uses this for the WCID address table (8 B) and the shared-key
+ * table (32 B) - 192 transfers that would otherwise be ~700 4-byte writes. */
+void mt_wr_copy(struct mt7612u_dev *d, uint32_t offset, const void *data, int len)
```
josephnef, line 562: `len = (len + 3) & ~3` then memcpy from `data` over-reads the caller's buffer whenever `len` is not a multiple of 4. Current callers pass 8 and 32 so it is latent, but a 6-byte MAC written through this reads two bytes past the source.

src/mt7612u/tools/bringup.c
```diff
@@ -0,0 +1,1181 @@
...
+{
+	struct mt7612u_caps c;
+	uint64_t t1, t2;
+	int64_t delta;
+	int bad = 0;
+
+	if (mt_eeprom_init(&dev)) return 1;
+	if (mt_init_hardware(&dev, NULL)) return 1;
+	if (mt_set_channel(&dev, chan, MT7612U_BW_20)) return 1;
+	if (mt_mac_start(&dev, 1)) return 1;
```
josephnef: This enables MAC RX and then never reads EP4: the gate sleeps 200 ms and injects 300 frames before `mt_mac_stop`. Same at line 978 in the ack gate. `docs/mt7612u.md` says the wedge was fixed by "never enabling RX for a caller that will not drain it", and `mt_rx_flush` drains at most 64 x 4 KiB. Either start the async RX ring here or pass RX off.

src/mt7612u/initvals.h
```diff
@@ -0,0 +1,70 @@
+/* SPDX-License-Identifier: BSD-3-Clause-Clear */
+/* GENERATED from openwrt/mt76 mt76x2/init.c mt76_write_mac_initvals()
```
josephnef: The generator this comment points at (`../../PLAN.md` Stage C) is not in the PR, so as shipped this is a hand-transcribed copy of mt76x2's MAC initvals. Project rule: tables are produced by a `tools/extract_*.py` generator with pinned per-source SHA-256 hashes and a `--check` mode that reproduces the checked-in file byte-for-byte (see `tools/extract_8733b_*.py` for the shape). The constants in `phy.c` and `init.c` belong in the same generator.

src/mt7612u/regs.h
```diff
@@ -0,0 +1,385 @@
...
+ */
+#ifndef MT7612U_REGS_H
+#define MT7612U_REGS_H
+
+#include <stdint.h>
+
+#define BIT(n)              (1u << (n))
+#define GENMASK(h, l)       (((~0u) - (1u << (l)) + 1) & (~0u >> (31 - (h))))
+/* Lowest set bit of a contiguous mask, for FIELD_PREP/GET. */
+#define _SHIFT(m)           (__builtin_ctz(m))
```
josephnef: `__builtin_ctz` under every FIELD_PREP/GET is the first MSVC blocker; `_SHIFT` is also a reserved identifier. A constexpr-style ctz shim (or `_BitScanForward` behind `#ifdef _MSC_VER`) is the usual fix, and it should land with the pthreads / `nanosleep` / `clock_gettime` / `<libusb-1.0/libusb.h>` items during integration rather than per-platform gating.

src/mt7612u/radiotap.c
```diff
@@ -0,0 +1,235 @@
...
+
+				r->phy = MT7612U_PHY_VHT;
+				r->mcs = (uint8_t)(mcs_nss >> 4);
+				r->nss = (uint8_t)(mcs_nss & 0x0f);
+				if (!r->nss) r->nss = 1;
+				if (known & 0x0004) r->stbc = flags & 1;
+
+				if (flags & 0x04)   r->sgi = 1;
+				if (coding & 0x01)  r->ldpc = 1;
+				r->bw = bwc == 0 ? MT7612U_BW_20
```
josephnef: Radiotap VHT bandwidth codes 2 and 3 are the 20 MHz sub-channels of a 40 MHz frame, and 5-10 are the 20/40 sub-channels of 80. Mapping 1-3 to 40 and >=4 to 80 airs a requested 20-in-40 frame at 40 MHz. Also, `DBM_TX_POWER` is silently ignored a few lines up; worth a diagnostic until per-packet power is wired.

src/mt7612u/caps.c
```diff
@@ -0,0 +1,107 @@
+/* SPDX-License-Identifier: BSD-3-Clause-Clear */
+/* TSF access and the static capability descriptor. */
+#include <string.h>
+#include "internal.h"
+
+/*
+ * DW0 is the LOW word. mt76's mt76x02u_restart_pre_tbtt_timer() assembles this
```
josephnef: The claim that mt76's `mt76x02u_restart_pre_tbtt_timer()` assembles TSF backwards is asserted without a file:line. The measured behaviour here stands on its own; cite the upstream line (or drop the upstream-bug claim) before it goes into the doc.

josephnef (commented):
**Scope: yes.** A MediaTek backend belongs here. The project will need a refactor first, and that is the first milestone rather than wiring this subtree in as-is.

The seams the refactor has to open, from reading the current tree:

- `IRtlDevice` / `RtlAdapter` / `WiFiDriver::CreateRtlDevice` are Realtek-branded names on an interface that is already vendor-neutral (Init / InitWrite / StartRxLoop / send_packet / SetMonitorChannel / FastRetune / GetAdapterCaps). That is a rename, not a redesign.
- `WiFiDriver` dispatches from a SYS_CFG2 read issued as a Realtek vendor request. It needs a vendor-neutral VID:PID gate in front of that read, the way Kestrel already gates PID-first.
- `IRtlTransport` is shaped around 16-bit registers and Realtek bulk endpoints. MT7612U is 32-bit registers plus in-band MCU commands over bulk EP8/EP5, so the MediaTek backend gets its own transport behind the same factory rather than a shim over the Realtek one.
- A `ChipGeneration` entry, a `DEVOURER_MT7612U` CMake option following the `DEVOURER_HAVE_*` pattern, and a `regress.py` cell against the MediaTek vendor driver under `reference/`.

**Licensing is fine.** I checked the mt76 file headers upstream: they really are BSD-3-Clause-Clear, so the attribution here is accurate and the code is GPL-2 compatible. A subtree LICENSE notice would be good practice in this tree.

**The code has bugs that sit under the measurements**, so it is not mergeable even as an unwired subtree. Inline below, most severe first: the public API is declared but never defined; RX corrupts every QoS-data header; TX header length is wrong for BAR/BA/CF-End; async teardown frees in-flight TX transfers; a failed register read feeds all-ones into every RMW; and the harness's caps/ack gates enable MAC RX without draining it, which is the wedge pattern the doc says is fixed.

Two cross-cutting items:

- `initvals.h` says it is generated by a script in `PLAN.md`; neither the script nor `PLAN.md` / `INVESTIGATION.md` / `BRINGUP-RESULTS.md` is in the PR, and 13 comments point at them. This project's rule is that tables come from `tools/extract_*.py` with pinned source hashes and a `--check` mode that reproduces the checked-in output. The hand-typed constants in `phy.c` and `init.c` should come from the same generator.
- Windows/MSVC is first-class here. `__builtin_ctz` under every FIELD_PREP/GET, pthreads throughout `async.c`, `nanosleep`, `clock_gettime`, the `<libusb-1.0/libusb.h>` include path and the gcc-only Makefile all need to land in portability shims during the integration, not per-platform gating.

src/mt7612u/include/mt7612u/mt7612u.h
snokvist: Fixed in 2142a5a. All nine are defined now — `open`/`close`/`start`/`stop` in `init.c`, `set_channel`/`set_chainmask` in `phy.c`, `keep_detached`/`asic_version`/`mac_addr` in `usb.c`. You spotted the signature disagreement too: `set_channel` took `unsigned` in the header against `uint8_t` internally, and the header is `uint8_t` now.

`tests/api_link.c` takes the address of all twenty public entry points while including only `<mt7612u/mt7612u.h>` and no internal header, so a declaration that loses its definition is a link error rather than something a caller discovers later. Mutation-tested: deleting one definition gives `undefined reference to mt7612u_asic_version`.

While defining them: `mt7612u_start()` enables RX only when an RX ring is already running, and the header says why.

src/mt7612u/rx.c
snokvist: Correct, and it was every QoS frame that reached the callback, not an edge case. Fixed in 2142a5a: the fold moves the real header length now, exactly as `mt76x02_remove_hdr_pad()` does, and TX and RX share one `mt_hdrlen_from_fc()` so the two sides cannot drift apart again.

`tests/frame_shape` covers it on a synthetic 26-byte-header QoS frame, and carries a negative control that redoes the old fixed-24 fold and asserts the QoS Control really is destroyed — without that, a passing test would say nothing about what it caught. Reverting the fix makes it report `QoS Control zeroed by the pad fold: aa aa`, which is the pad bytes.

src/mt7612u/tx.c
snokvist: Right. Rather than extend the list of 16-byte subtypes, I ported `ieee80211_hdrlen()`: control frames default to 16 and only CTS and ACK are 10. That is the shape that cannot be incomplete, and it is the same function mt76 reaches through `ieee80211_get_hdrlen_from_skb()` on both the TX and RX sides — so this now matches upstream rather than approximating it. It also picks up the HT Control cases (+4 on Order) that the old version missed.

Fixed in 2142a5a, covered in `tests/frame_shape` over all eight control subtypes and the five data shapes.

Worth being explicit: this one has no on-air observable in the gates here, because nothing in them injects a control frame. The unit test is the whole of the evidence for it.

src/mt7612u/async.c
snokvist: All of it, and the wedged chip is the case that makes it real rather than theoretical. In 2142a5a both rings are cancelled, both are waited on with the event thread still running, and if either still has transfers outstanding at the deadline the whole `mt_async` is deliberately leaked rather than freeing memory libusb owns.

Each slot now names its own ring instead of `dev->a`, so a leaked ring's late completions cannot land on a replacement ring. `fail:` no longer joins a zeroed `pthread_t` — an `evt_started` flag guards it.

`running`, `rx_active` and both inflight counters moved under the mutex; you are right that `volatile` orders nothing and makes no read-modify-write atomic. The statistics moved with them, behind `mt_async_stats()`, because the harness was reading them live off the struct — same bug, one level down. And a caller blocked waiting for a TX slot now gives up when the ring stops instead of parking forever.

src/mt7612u/usb.c
snokvist: Fixed in 2142a5a. `mt_rr_chk()` reports failure separately from the value, `mt_rmw()` refuses to write after a failed read, and `mt_wait_for_mac()` distinguishes "not ready" from "no transport". `0xffffffff` is a real `MT_MAC_CSR0` value while the core comes up, so it could never have served as a sentinel.

The second half of the problem mattered as much: `mt_poll()` counted sleeps, so with one access costing up to `VEND_RETRIES × timeout` a caller asking for 200 ms could block for seconds. It polls against a real deadline now and aborts on a read failure. The control timeout also drops from 1000 ms to mt76's own `MT_VEND_REQ_TOUT_MS` of 300, which bounds the worst case at 3 s.

Re-verified on hardware afterwards — bring-up is unaffected.

src/mt7612u/usb.c
snokvist: Fixed in 2142a5a: the *transfer* length is rounded up into a zeroed local buffer, the caller's `len` is not. Latent today as you say, and a 6-byte MAC is exactly the next call site someone would add.

src/mt7612u/tools/bringup.c
snokvist: You are right, and the doc claiming the wedge was fixed a few lines away made it worse rather than better. Both gates start the RX ring before enabling the receiver now (2142a5a) — `caps` in particular then sat through two 200 ms sleeps and a 40 MHz channel switch with nothing reading EP 4.

Re-run on hardware: three consecutive `caps` cycles clean, each logging `async: 16 RX transfers in flight`.

The structural half is that `mt7612u_start()` enables RX only when a ring is already running, so the public API cannot reproduce the pattern even if a caller wants to.

src/mt7612u/initvals.h
snokvist: Done in c64975d. `reference/mt76` is a pinned shallow submodule at `be5ce79`, and `tools/extract_mt7612u_tables.py` follows the `extract_8733b_*.py` shape: `UPSTREAM` string, per-source SHA-256, `EXPECTED` count and table hash, and `--check` byte-comparing the checked-in header.

The source is C initialiser syntax rather than a vendor parameter blob, so the script carries a symbol table over `mt76x02_regs.h` and a small constant-expression evaluator for `BIT`/`GENMASK`/`FIELD_PREP`. That is what lets the four `DEFAULT_PROT_CFG_*` macros — defined inside `mt76_write_mac_initvals()` itself, not in the register header — be computed rather than copied. Both the symbol table and the evaluator raise on anything they do not understand rather than guessing.

It reproduces the hand-typed table byte for byte, all sixty rows including the four computed protection-config words, so it confirms the original transcription instead of silently replacing it. `--check` is load-bearing: appending one newline to the header exits 1 with `stale generated output`.

`reference/mt76` is the one entry in that directory that is not a Realtek vendor drop — the MediaTek parts have no out-of-tree vendor driver to mirror, and mt76 being BSD-3-Clause-Clear rather than GPL-2-only is what lets `src/mt7612u/` carry ported sequences at all. `reference/README.md` says so. No workflow checks out submodules, so the twenty builds are unaffected.

The stale `../../PLAN.md` pointer is gone with the regenerated header.

src/mt7612u/regs.h
snokvist: Both fixed in 2ac9b9a. `_SHIFT` is gone — leading underscore followed by a capital is reserved to the implementation in every scope.

I did not use `_BitScanForward`: it is a function taking an out-parameter, so it cannot appear in a constant expression, and these macros have to stay constant expressions because `ext_cca_chan` in `phy.c` is a static table built from `FIELD_PREP`. `MT_CTZ` isolates the low bit and binary-searches its position — a constant expression on every compiler, folded to one instruction under optimisation.

`tests/field_macros` checks it against `__builtin_ctz` over all 32 single-bit masks and all 528 contiguous `GENMASK(h, l)` ranges, with a `FIELD_PREP`/`FIELD_GET` round-trip on each. It also carries a static initialiser built from `FIELD_PREP`, so if the macro ever stops being constant-foldable the test fails to *compile* rather than passing at runtime.

Same commit: `<libusb.h>` — this project's spelling — is tried first via `__has_include`, with the distribution's `libusb-1.0/` path as the fallback for the standalone build.

Deliberately not done, and I would rather say so than half-do it: `async.c` still uses pthreads and `usb.c` still uses `nanosleep`/`clock_gettime`. This project has no C threading or time shim; its shim is the C++ standard library, which every other backend uses directly. A C shim written now would be deleted the moment the subtree joins the build. Those are the only two files involved and the README names them. Same reasoning for the Makefile staying a Makefile while nothing in `CMakeLists.txt` reaches the subtree — a CMake target now would be a target nothing builds.

src/mt7612u/radiotap.c
snokvist: You are right, and I had this backwards on the first read. Fixed in a9cd156 with a table over the eleven codes the part can express: 2 and 3 are 20 MHz, 5 and 6 are 40, 7–10 are 20. Codes 11 and up are 160 MHz and its sub-channels, and the rate word has no 160 encoding, so those log and fall back to 20 rather than narrowing silently.

What settled it was internal to the repository rather than the spec text: the HT branch a few lines above already reads the equivalent codes correctly, and `src/ieee80211_radiotap.h:112` names them `IEEE80211_RADIOTAP_MCS_BW_20L` and `_20U`. The VHT branch was disagreeing with the repository, not just with the standard.

`tests/frame_shape` now pins all eleven codes; restoring the old expression fails eight of them.

`DBM_TX_POWER` emits one diagnostic per process naming the two knobs that do work, rather than accepting the field and dropping it (2142a5a).

One thing to flag rather than fix from here: `src/jaguar1/RtlJaguarDevice.cpp:1097-1100` and `src/jaguar3/RtlJaguar3Device.cpp:1940-1943` carry the same `bw >= 1 && bw <= 3 → CHANNEL_WIDTH_40`, `bw >= 4 && bw <= 10 → CHANNEL_WIDTH_80` mapping. I have not touched them — changing a shipping backend's on-air behaviour from inside a new-backend PR seemed like the wrong place. Happy to send it as its own PR if you want it.

src/mt7612u/caps.c
snokvist: Cited in 2142a5a: `mt76x02_usb_core.c:155-158`, where

```c
dw0 = mt76_rr(dev, MT_TSF_TIMER_DW0);
dw1 = mt76_rr(dev, MT_TSF_TIMER_DW1);
tsf = (u64)dw0 << 32 | dw1;
dev_dbg(dev->mt76.dev, "TSF: %llu us TBTT %u us\n", tsf, tbtt);
```

`tsf` is consumed only by that `dev_dbg()`, which is why the order has survived upstream. The comment quotes those four lines now.

The `caps` gate also prints both orders against a known 200 ms sleep, so the claim is re-checkable on any sample rather than taken on trust — this run reads 200347 µs for `(DW1 << 32) | DW0`.

snokvist:
Thanks — that review was worth more than the measurements were. Every one of the eleven findings was real, including the VHT one I initially thought I could defend. All are fixed and answered in-thread; I have left the threads for you to resolve.

**Five commits** on top of the original:

| | |
|---|---|
| `2142a5a` | the nine confirmed defects |
| `c64975d` | `initvals.h` generated instead of transcribed |
| `2ac9b9a` | the portability fixes that stand on their own |
| `081f952` | offline tests documented, everything re-measured on hardware |
| `a9cd156` | the radiotap VHT bandwidth mapping |

Two of the fixes change what the driver puts on air or hands to a caller, so nothing here is claimed from a rebuild alone. The gates were re-run on the same bench against the same RTL8812AU witness: `rtap` tags A 401 / B 388, `ampdu` 41 734 of 85 580 witnessed frames with `paggr=1` and the same 15.50 and 44.55 Mbit/s the PR reported, `caps` clean on three consecutive cycles with the RX ring now started before the receiver, 40 MHz re-confirmed 300/300 at `bw=1`, and 278 ambient frames on ch1 through the rewritten pad fold with `rx_err=0`.

`make -C src/mt7612u check` is new and runs three offline binaries — no hardware, no privileges. Each was mutation-tested rather than merely written: reverting any one of the frame-shape fixes, the VHT mapping or a public definition makes the suite fail. The header-length fix has no on-air observable in these gates, because nothing here injects a control frame, and the docs say that rather than implying otherwise.

`CMakeLists.txt` is still untouched.

---

**On sequencing.** Your review reads two ways and it changes what I build next, so let me state how I read it and you can correct me.

> the code has bugs that sit under the measurements, so it is not mergeable *even as* an unwired subtree

I took that to mean an unwired subtree *is* the mergeable shape once the bugs are gone, and that

> the project will need a refactor first, and that is the first milestone rather than **wiring this subtree in** as-is

puts the refactor before *wiring*, not before *landing*. So my plan is: this PR lands unwired, and the refactor plus integration arrives as its own PR. Bundling a rename that touches all three shipping backends together with a brand-new backend makes one diff nobody can review. Say the word if you meant the other order and I will hold this one.

**On the refactor itself**, one question, because I would rather ask than guess at the shape of your codebase. Everything substantive in your list I agree with: a vendor-neutral VID:PID gate ahead of the Realtek `SYS_CFG2` read the way Kestrel already gates PID-first, MediaTek getting its **own** transport behind the same factory rather than a shim over `IRtlTransport` (32-bit registers plus in-band MCU over EP8/EP5 do not fit it), a `ChipGeneration` entry, a `DEVOURER_MT7612U` option following `DEVOURER_HAVE_*`, and a `regress.py` cell.

The part I would push back on gently is the rename. `IRtlDevice` is already vendor-neutral in substance — `Init` / `InitWrite` / `StartRxLoop` / `send_packet` / `SetMonitorChannel` / `FastRetune` / `GetAdapterCaps` say nothing about Realtek — so a MediaTek backend does not need the rename to sit behind it, and doing it inside this work would touch every backend for a naming reason. I would rather implement against `IRtlDevice` / `WiFiDriver::CreateRtlDevice` as they stand, and send the rename separately as its own mechanical, reviewable-in-isolation PR if you still want it. Your call — it is your codebase, and if you would rather have the rename first I will do that instead.

That also decides where the remaining portability work lands. `async.c` still uses pthreads and `usb.c` `nanosleep`/`clock_gettime`; this project has no C threading or time shim because its shim is the C++ standard library. Rather than write a C shim that gets deleted at integration, I would like to do that swap as part of the integration commit, where the subtree joins the build and the answer is `<thread>`/`<chrono>` like everywhere else. Two files, and the README already names them.

snokvist:
One more, found while working out what the integration has to look like, and it is a limit rather than a fix.

`Packet::Data` in `src/RxPacket.h` is documented as the full 802.11 frame **including the trailing FCS**, and every Realtek parser here honours that. MT7612U cannot.

Every RX buffer does carry 4–7 bytes past `MPDU_LEN`: over 4263 ambient frames the tail was exactly 4 on 3375 of them and 5–7 on the rest — a fixed 4-byte field plus USB 4-byte alignment, which is exactly what an FCS would look like. Handing a consumer `len + 4` would have looked right in every log.

It is not the FCS. CRC-32 over the MPDU matched those four bytes on **0 of 4263** frames. I checked the probe's own CRC-32 against the standard `"123456789"` → `0xcbf43926` vector before believing the negative. They are the FCE info trailer — `mt76u_get_rx_entry_len()` computes `min_len = MT_DMA_HDR_LEN + MT_RX_RXWI_LEN + MT_FCE_INFO_LEN`, and `dma.h:48` has `MT_FCE_INFO_LEN 4`. mt76 never sets `RX_FLAG_INCLUDE_FCS` for this family either: the MAC strips the checksum and does not hand it up.

So a consumer that trims four bytes at its protocol boundary — which the `Packet::Data` comment invites, and which `tools/bf_report_decode.py` already does — would eat four bytes of payload off every MT7612U frame.

Documented in `docs/mt7612u.md` and on the RX callback in the public header (5e85f7a). No code change; the code was already right, it just did not say what it could not do. Raising it now because it is the kind of thing that should be settled before the integration commit rather than discovered inside it — and because it is a real asymmetry against the Realtek backends that a caps flag or a documented boundary will have to carry.

snokvist:
The integration exists and works, on a branch rather than a second PR — that ordering is yours to call, so I have not opened one: [`feat/mt7612u-integration`](https://github.com/snokvist/devourer/compare/master...snokvist:devourer:feat/mt7612u-integration), stacked on this branch.

Shape, following your list: `RtlMt7612uDevice : IRtlDevice`, dispatched on VID:PID ahead of the `SYS_CFG2` read (the same place Kestrel gates, for a stronger version of the reason — `0x00FC` is a Realtek address issued with a Realtek vendor request, so on MediaTek silicon that read is the wrong protocol rather than merely ambiguous). It does **not** shim over `IRtlTransport`: the C library here is the transport, and `mt7612u_open_handle()` adopts the handle `WiFiDriver` already opened, reset and claimed rather than reopening it and racing the caller's lock. `ChipGeneration::Mediatek`, `DEVOURER_MT7612U` (OFF by default), three `ctest` cells.

The one place I diverged from your list is the rename, and I have argued the case in the comment above rather than quietly skipping it. Everything else is as you described it.

**Two bugs surfaced that only integration could find** — both invisible to the bring-up harness, because it never crosses this boundary:

- **RSSI base.** devourer carries RSSI as an unsigned byte biased by 110 (`LinkHealth.cpp:8`). Casting the HAL's signed dBm straight in reported −63 dBm as **193** — a claimed +83 dBm.
- **The receive filter.** `mt7612u_start()` leaves `MT_RX_FILTR_CFG` at mt76's managed-station value, which drops control frames and anything not addressed here. A/B against the harness on the same channel in the same minute: the integrated path saw **0 OFDM frames of 244**, the harness **103 of 402**. With a monitor filter the same run takes **7096 frames** carrying the full CCK/OFDM/HT mix. `Rtl8733bDevice::Init` calls `configure_monitor_rx` for exactly this reason and I had simply not done the equivalent.

The second is the one worth the paragraph: 244 beacons looks like a working receiver. Only the A/B against a second implementation of the same silicon made it visible, and that is now written down in `docs/mt7612u.md` rather than just fixed.

`ctest` 58/58 with the option ON, 55/55 OFF; the mapping selftest is mutation-tested; and this project's own `rxdemo` drives the adapter end to end through the factory.

One thing to flag about the branch split: it also carries a **subtree** fix that arguably belongs in this PR instead — `mt_rx_parse()` now drops a frame whose rate word names no valid PHY. `MT_RATE_PHY` is three bits, so 5–7 are representable and mean nothing; mt76 returns `-EINVAL` and drops them, while mine decoded one through the default arm and reported the lowest legacy rate, i.e. garbage presented as a real CCK frame. It measured 0 on the bench so it was not the cause of the filter symptom, just a hole found while chasing it. It is entangled with an API addition made for integration, which is why it landed there — say the word and I will split it back into this PR so the subtree is complete on its own.

snokvist:
Marked ready for review. It should have been from the start — asking you to weigh in on merge order while leaving it a draft was my mistake, not a signal.

## A defect in this branch, found since your review

`mt_set_channel_ex()`'s 40 MHz case derives the centre channel from the control channel and never validated the result. mt76 doesn't need to — `mt76x2u_phy_set_channel()` takes `center_freq1` from the chandef, so an off-grid channel can't reach it. This port takes a bare channel number, so an off-grid one still produces *a* number and then transmits 40 MHz wide somewhere the caller didn't ask for, with every register write succeeding.

Two cases reach the MCU straight through `mt7612u_set_channel()`, which checks only for a zero channel:

- the arithmetic ran in `uint8_t`. Control channel 254 computes 256, truncates to **0**, and hands the MCU channel index 0 with the 5 GHz register set loaded. 255 gives 1.
- in 2.4 GHz the pairing only ever reaches centres 6–9, so channel 1 computes −1, i.e. 255 after truncation.

`mt_chan40_centre()` computes in `int` and validates against the centres that exist, catching off-grid, out-of-band and both wraps with one check. 2.4 GHz 40 MHz now works for channels 4–11 and is refused elsewhere — a deliberate narrowing, since a bare channel number can't say which side the secondary sits on where channel 6 pairs equally well with 2 or with 10.

The caveats matter more than the fix. **No measurement in this PR moves**: `bringup caps 149` still sends 300 frames at 40 MHz and the witness still reports 300 at `bw=1`. The bug was never in the path any published number exercised. The nine-case test is mutation-tested, but restoring the `uint8_t` arithmetic does *not* make it fail — no wrapped value aliases onto a legal centre, so the `int` only buys a refusal that names 256 instead of reporting 0 and sending the reader after the wrong bug. The comment says that rather than claiming a guard it doesn't provide.

Also here: `make check` now depends on `bringup`. Nothing else compiled it, so a change that broke a gate shipped with everything green — which is how I found this.

## The second branch, and a question back

The stacked branch now carries the integration plus 80 MHz, and I've brought its build wiring to parity with `src/rtl8733b` and `src/kestrel`: an `mt7612uprobe` CMake target beside `rtl8733bprobe`, the offline tests as native ctest executables instead of a `make -C` shell-out (that one silently ignored `CMAKE_C_COMPILER` and `DEVOURER_SANITIZE`), a POSIX `FATAL_ERROR` guard mirroring `DEVOURER_PCIE`'s, an `mt7612u-only` CI cell — nothing built this backend before — and `-DDEVOURER_MT7612U=OFF` in the no-chip-selected control, which was passing on a default rather than on naming every option.

80 MHz airs and is witnessed as 80 MHz, with the negative control: a 20 MHz witness decodes **0** of it, and a witness on a sibling control channel of the same group decodes it, which is what shows both ends resolved the same centre. Still one witness generation; VHT 2SS MCS5–9 thin out at 80 MHz in a way I attribute to link budget without having measured the cause; and the periodic RX gain worker (`mt76x2_phy_update_channel_gain`) is unported — only its width-dependent registers are.

On sequencing: you asked for the refactor before wiring this in, and I read that as *before wiring, not before landing*. That is still a guess, and it's the one thing I don't want to decide for you. **Either shape is one push from here** — land this unwired and keep the integration as a separate PR, or I push the integration commits onto this branch so you review a single full-parity backend against the others. Say which and it's done.

snokvist:
Qodo posted fourteen findings when this left draft. I checked each against the code and against the pinned mt76 tree rather than taking them on trust, which turned out to matter: **two were wrong, and "fixing" either would have made things worse.** `eca657c` acts on the rest.

### Fixed — six, one of which is on-air correctness

**The radiotap HT bandwidth was gated on the wrong bit.** The MCS `known` byte declares bandwidth with `HAVE_BW` (0x01); `HAVE_MCS` is 0x02. We tested 0x02, so a header declaring bandwidth *without* an MCS index — legal radiotap — silently narrowed a requested 40 MHz frame to 20. This is the exact twin of the VHT bandwidth bug you caught in review, in the same function, and I fixed that one without noticing this one. Mutation-tested: restoring 0x02 fails two cases in both directions.

**A malformed radiotap header transmitted anyway.** A declared field running past the declared header length returned the header length, which both injection entry points read as success and then aired at whatever defaults had accumulated. Now refused. An unknown *trailing* present bit still stops cleanly — that header is well formed, we simply cannot read the rest of it.

**`--check` crashed on anything before Python 3.13.** `Path.read_text(newline=...)` is 3.13-only, so the documented verification command and the new ctest cell would `TypeError` on the 3.12 the CI images ship. This was invisible here precisely because this machine runs 3.13 — a gate that only passes on the author's box.

Plus three in the harness: `phy_name[phy & 7]` indexed a five-entry array (`by_phy[]` was already 8 and fine); `gate_g` discarded every transmit return, printed the requested count and returned 0, so a run where every submit failed still handed the witness an experiment that never aired — and "no frames decoded" would have read as a negative finding rather than as a silent transmitter; and the ACK gate cast an unchecked signed duration to unsigned, making `-1` roughly 49 days with the receiver live.

On air, unchanged: `bringup rtap 149 400` still gives tag A 400/400 and tag B 377/400 at rate 19, against the 401/388 recorded before.

### Two that are not defects

**The 2.4 GHz VHT power "double shift" is `mt76x2_get_rate_power()` byte for byte**, and `rate_power_val` takes a `u8` on both sides, so the behaviour is identical to upstream's. Those two entries do decode from zero on 2.4 GHz — that is mt76's own handling of a 5 GHz-oriented EEPROM field, for a band where VHT is an extension outside 802.11ac. Changing it would be a silent divergence from the reference this port is checked against, so it is commented instead.

**`clear_ack_responder()` not clearing `MT_AUTO_RSP_EN` is correct.** `mt_init_hardware()` writes `MT_AUTO_RSP_CFG = 0x13` and `MT_AUTO_RSP_EN` is `BIT(0)`, so the gate is already on before any caller arms a responder — the `mt_set()` in the arm path is a no-op on it. Clearing it would leave the device in a state its own init never produces; moving the identity off the responder address is what actually stops it answering. Also commented.

### Five left open deliberately

Transfer teardown racing libusb ownership after the two-second cancel deadline; `mt_wr()` discarding failures so a partially applied hardware setup can return success; EEPROM read errors reaching calibration data as `0xff`; the harness counter data race between the libusb event thread and the gate; and `LOG`/`ERR` not matching the `devourer [level] message` contract.

These are real. I have not touched them because each is a design change rather than a fix — error propagation through `mt_wr` in particular reshapes every call site in the backend — and I would rather have your view on the shape than guess and hand you a large diff to unpick. Say which you want and in what form.

snokvist:
Filed #413 for the cross-backend half of this: `AdapterCaps` declares no maximum frame size for any adapter, and the limit turns out to differ per transport — PCIe already enforces a hard 32 KB bounce ceiling invisibly (`PcieTransport.cpp:569`), the Realtek USB backends size a heap buffer to the frame so their real limit is undocumented, and MT7612U's RX ceiling is a MAC register.

Keeping it out of this PR deliberately: the field is shared, the "what should the default be" question is yours rather than a backend's, and the Realtek numbers need a register hunt plus a bench sweep that has nothing to do with MediaTek.

What did land here, because it was a defect rather than a design question: this backend had **two disagreeing ceilings behind one public API** — `mt7612u_tx()` refused above 2016 bytes while `mt7612u_send_packets()` bounded only against its 16 KB aggregate buffer, so the same frame was refused by one entry point and aired by the other. Both now share one ceiling.

The numbers, measured rather than assumed: TX is not the constraint on this part — every size up to **7900 bytes** aired 60/60 against an RTL8812AU witness with zero CRC errors, well past 802.11's 2304 non-A-MSDU ceiling, so the old 2048 was a buffer constant with nothing behind it. RX is the constraint and it is the MAC's: `MT_MAX_LEN_CFG` low 12 bits are the on-air maximum *including* FCS, so the `0xf00` this driver programs gives **3836 bytes of MPDU**, measured to the byte — 3836 arrives, 3837 does not.

Both are now in `mt7612u_caps` (`max_mpdu_tx` / `max_mpdu_rx`, RX read from the register at runtime rather than hardcoded), `bringup caps` prints them, and a `bringup mtu` gate is the sweep that produced them.

One correction worth stating plainly, since it was mine: I added an `rx_dropped` counter mid-investigation and said it would make the oversize loss visible. It does not, and I mutation-tested rather than trusting it — with the TX ceiling temporarily raised, 60 frames of 6000 bytes MT7612U-to-MT7612U gave 0 received with `rx_err`, `rx_invalid` and `rx_dropped` all zero. The MAC discards oversize frames before USB, so that loss cannot be observed from this layer at all. The counter is real but covers short and malformed transfers only, and now says that instead of claiming the case it cannot see.

Caveats that stand: one witness generation, and 7900 is "at least" — I stopped at the buffer I had, not at a refusal.

josephnef:
Re-verified at eca657c against the code, not the replies: all eleven findings from my review are fixed. Build, `make check` and the extractor `--check` against the pinned `reference/mt76@be5ce79` all reproduce here, and I have resolved those threads. The 40 MHz centre validation and the HT `HAVE_BW` twin are good catches.

Answers to the three open questions:

1. **Merge shape** — land this PR unwired; the integration comes as a follow-up PR. Please split the invalid-PHY rate-word drop back into this branch so the subtree is complete on its own.

2. **The rename is a must.** Once this lands, `IRtlDevice` / `RtlAdapter` / `CreateRtlDevice` have nothing to do with Realtek any more, and the names would be lying to the next reader. Your proposed shape is fine: a separate, mechanical, reviewable-in-isolation PR, landed before the integration PR so the MediaTek device is born under the vendor-neutral name rather than renamed after.

3. **The five items Qodo left and you deferred** — please address them in this PR rather than waiting on a shape from me: write-failure propagation through `mt_wr`, transfer ownership after the teardown deadline, EEPROM read errors reaching calibration as `0xff`, the harness counter race, and `LOG`/`ERR` following the `devourer [level] message` contract (`docs/logging.md`). Where one genuinely reshapes every call site, do it the way the other backends do it and say so in the commit; I would rather review the diff than pre-approve a design in prose.

The pthread / `clock_gettime` swap to `<thread>` / `<chrono>` at integration time is fine as you proposed. #413 is a good issue and stays out of here.

snokvist:
Done at `1b032b5`, all three answered.

## 1. Merge shape — unwired, and now complete

Nothing in this branch touches `CMakeLists.txt`, `WiFiDriver.cpp` or the workflows; the subtree is still unreachable from the shipped library.

The invalid-PHY rate-word drop is back, as asked. I also brought over the rest of the **subtree-level** work that had drifted onto the integration branch, because your reason applies to all of it equally: the per-frame noise floor and derived SNR, 80 MHz, the frame-size limits, and the adapter lock shared with `UsbDeviceLock`. What stayed behind is only wiring — the `IRtlDevice` implementation, the VID:PID table, the mapping header. If you would rather review a narrower diff, say so and I will split the later work back out; I read "complete on its own" as the governing principle rather than the one file you named.

## 2. The rename — agreed, and I was wrong to push back

Your reason is better than mine. I was arguing the interface is neutral *in substance*, which is true and beside the point: after this lands the names are read by people, and `RtlAdapter` in front of a MediaTek device is a lie they have to work around. Separate mechanical PR, landed before the integration PR so the device is born under the neutral name. I will open it once this merges, so it rebases on a settled tree rather than a moving one.

## 3. The five items — all closed here

**Write-failure propagation.** `mt_wr()` discarded `mt_vendor_req()`'s result, so a write that exhausted its retries left the hardware partly configured while the call returned success. Threading a status through every writer would touch several hundred call sites in a sequence that is deliberately a verbatim port of mt76's, and that churn would bury the thing it protects — so I did it the way the reads already worked: `mt_rr_chk()` has always bumped `d->io_err`. Writes stay best-effort at the call site; a **sequence** checks the accumulator at its boundary. `mt_eeprom_init()`, `mt_init_hardware()` and `mt_set_channel_ex()` are bracketed and now fail as a whole. Optional writes stay best-effort by not being bracketed, and `mt_wr_chk()` exists for a caller that wants to fail at the write itself.

**EEPROM read errors.** The slurp must not treat `0xffffffff` as an error *by value* — unprogrammed cells legitimately read all-ones — but the accumulator separates a blank cell from a failed transfer, which the value alone cannot. Previously a failure outside the two cells already validated left all-ones power and calibration bytes in memory while the device opened successfully and transmitted from them.

**Teardown ownership.** Leaking the ring on the two-second deadline was the safe half. The other half was that `mt_close()` then released the interface, closed the handle and exited the context while libusb still owned those transfers — and the event thread had just been joined, so nothing could ever complete them. The device is now marked stranded: `mt_close()` leaks the USB objects too rather than freeing what libusb holds, and `mt_async_start()` refuses to submit a second ring onto the same endpoints. Consistent with the existing leak rather than a new policy.

**Harness counter race.** `arx_cb()` and `ack_cb()` run on the libusb event thread while `gate_arx()`/`gate_duplex()` read the same fields *before* stopping it, so a displayed rate or a duplex verdict could be built from torn counts. Relaxed atomics rather than per-gate locks — they are counters, nothing orders off them, and this is the RX hot path in a throughput gate.

**`LOG`/`ERR`.** They used a private `[mt7612u]` prefix instead of `devourer [X] `, emitted the text and its newline as two stdio calls so a line from the event thread could split one from a caller, and never flushed. `mt_diag()` formats the line first and emits it with one `fwrite` + `fflush`, which is what `src/Event.h` does for the machine plane and for the same reason. Warning-level diagnostics moved to `[W]`.

## Hardware verification after the rework

Re-run end to end so this is one record rather than a claim — **19 gates, 0 failures**:

| | |
|---|---|
| self-checking | `regs` `fw` `init` `chan` `caps` `hop` `pwr` `linkstat` `ampdu` `soak` |
| witnessed by an RTL8812AU | `rtap` 781 · `sweep` 1617 @20 / 1639 @`bw=1` / 828 @`bw=2` · `coding` 1919 · `gateg` 450 · `mtu` 400 · `vht` 748 |
| MT7612U → MT7612U | 2999 frames, `rx_err` `rx_invalid` `rx_dropped` all 0 |

Offline: `make check` green, and `tools/extract_mt7612u_tables.py --check` still reproduces `initvals.h` against the pinned `reference/mt76@be5ce79`.

The standing caveats are unchanged and worth repeating rather than burying: **one witness generation** (a single RTL8812AU) behind every on-air number, two samples of one board revision, no cold-boot test on this host, and the periodic RX gain worker (`mt76x2_phy_update_channel_gain`) still unported — only its width-dependent registers are.

josephnef (dismissed):
Verified at 1b032b5: the five deferred items are closed as described (accumulator-bracketed sequences, EEPROM blank-vs-failed separation, stranded-device teardown, atomic harness counters, single-fwrite diagnostics), the subtree-level work is back in this branch, and nothing touches the shipped build. Local build, `make check` and the extractor `--check` against the pinned `reference/mt76` all reproduce. Landing unwired; the rename PR and the integration PR follow in that order.

src/mt7612u/README.md
```diff
@@ -0,0 +1,112 @@
...
+| `tools/bringup.c` | one subcommand per verified gate |
+| `tests/` | offline tests (`make check`): public-API link, frame shapes |
+| `initvals.h` | **generated** — see Provenance |
+
+## The receiver must never run undrained
+
+Enabling MAC RX with nothing reading the bulk-IN endpoint wedges this part
+*below* the USB level: `libusb_reset_device`, the sysfs `authorized` toggle
+and rebinding the kernel driver all fail to recover it, and only a physical
+replug does. So `mt_mac_start()` takes the receiver as an explicit argument,
```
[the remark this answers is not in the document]
josephnef: Dismissed by the maintainer: the README section is the rationale (why the wedge is unrecoverable and what shape the API takes because of it); the public header carries the contract. Both stay.

src/mt7612u/eeprom.c
```diff
@@ -0,0 +1,276 @@
...
+	t->ht[8] = t->ht[9] = rate_power_val(v & 0xff);
+	t->ht[10] = t->ht[11] = rate_power_val(v >> 8);
+	v = mt_ee(d, MT_EE_TX_POWER_HT_MCS12);
+	t->ht[12] = t->ht[13] = rate_power_val(v & 0xff);
+	t->ht[14] = t->ht[15] = rate_power_val(v >> 8);
+
+	v = mt_ee(d, MT_EE_TX_POWER_VHT_MCS8);
+	if (!is_5ghz)
+		v >>= 8;
+	t->vht[0] = t->vht[1] = rate_power_val(v >> 8);
```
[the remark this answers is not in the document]
josephnef: Dismissed by the maintainer: verified against the pinned reference/mt76 — mt76x2_get_rate_power() shifts the 2.4 GHz VHT word the same way and rate_power_val() takes a u8 on both sides, so this is upstream's behaviour reproduced, not a transcription slip. The comment in eeprom.c now says so; diverging from the reference silently would be worse.

tools/extract_mt7612u_tables.py
```diff
@@ -0,0 +1,302 @@
...
+            f"expected count={EXPECTED_COUNT} sha256={EXPECTED_SHA256}"
+        )
+
+    output = render_h(rows)
+    # Explicit encoding + newline="": the generated file is a byte-defined
+    # artifact, so --check must compare identically on every platform.
+    if args.check:
+        path = root / OUTPUT_H
+        if not path.exists() or path.read_text(encoding="utf-8", newline="") != output:
+            raise SystemExit(f"stale generated output: {OUTPUT_H}")
```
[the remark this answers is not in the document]
josephnef: Fixed at eca657c: --check reads through open(encoding='utf-8', newline='') rather than Path.read_text(newline=...), verified reproducing initvals.h here.

src/mt7612u/eeprom.c
```diff
@@ -0,0 +1,300 @@
...
+	else                delta_idx = 4;
+
+	if (offset + 5 > sizeof d->eeprom) return;
+	p = d->eeprom + offset;
+	t->chain[chain].tssi_slope   = (int8_t)p[0];
+	t->chain[chain].tssi_offset  = (int8_t)p[1];
+	t->chain[chain].target_power = (int8_t)p[2];
+	t->chain[chain].delta = (int8_t)sign_extend_optional(p[delta_idx], 7);
+
+	t->target_power = mt_ee(d, MT_EE_RF_2G_RX_HIGH_GAIN) & 0xff;
```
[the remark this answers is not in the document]
josephnef: Dismissed by the maintainer: this is mt76x2_get_power_info_5g() reproduced — the pinned reference/mt76@be5ce79 mt76x2/eeprom.c:421-422 reads MT_EE_RF_2G_RX_HIGH_GAIN and takes its low byte as target_power in exactly this place, with the same TSSI/field_valid override after. Same policy as the VHT power entry: match the reference, comment it, do not diverge silently.

src/mt7612u/usb.c
```diff
@@ -0,0 +1,592 @@
...
+		return -2;
+	}
+	WARN("cannot lock %s (%s) - proceeding without exclusivity",
+	    path, strerror(errno));
+	return -1;
+}
+
+static libusb_device_handle *open_selected(libusb_context *ctx, const char **err)
+{
+	const char *sel = getenv("MT7612U_DEV");
```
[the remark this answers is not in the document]
josephnef: Deferred to the integration PR by the maintainer: the getenv lives in the standalone mt7612u_open() path, which the library entry point (mt7612u_open_handle, adopting the handle WiFiDriver already claimed) does not go through. Moving MT7612U_DEV parsing into tools/bringup.c and out of the library is part of the integration work, where the 'library reads no environment' rule applies.

josephnef:
Qodo's re-review of 1b032b5 raised seven. I checked each against the code and the pinned mt76; two are closed in-thread (the 5 GHz target-power one is `mt76x2_get_power_info_5g()` verbatim, the `MT7612U_DEV` read is deferred to integration). **Five are real and hold the merge** — please address them here:

1. **Async ring rejects what the TX builders produce.** `mt_async_tx_submit()` refuses anything above `MT_TX_BUFSZ` (2048), but `mt_tx_raw()` builds up to `MT_TX_BUF_MAX` (4096) and `mt7612u_send_packets()` up to `MT_USB_AGG_BUF` (16 KB), and both route through the ring whenever RX is running. So in the normal integrated shape (RX loop up), a frame over ~2 KB or any multi-frame batch is refused after building, and `send_packets` reports zero accepted. The `mtu` numbers were measured on the sync path, which is why it did not show. Size the slots to the aggregate buffer or split at the ring — either way one ceiling, and a `frame_shape`-style test that submits a 16 KB aggregate through the async path.
2. **`gate_coding` / `gate_sweep` / `gate_vht` return 0 when nothing aired.** Same shape as the `gate_g` fix in eca657c: a shortfall is printed and then exit 0. Fail on `sent != count` and reject non-positive counts.
3. **`gate_init` / `gate_chan` discard `fw_dir`** — `mt_init_hardware(&dev, NULL)` at both sites, so the documented `init <fw_dir>` form silently uses the default directory.
4. **`open_selected()` keeps the adapter lock when `libusb_open()` fails.** `g_lock_fd` is set before the open, and only `mt_close()` releases it, which a failed `mt_open()` never reaches — so a retry in the same process collides with its own stale lock.
5. **`g_stats_last_us` is file-static.** Two adapters in one process corrupt each other's `interval_us`. It belongs in `struct mt7612u_dev`.

Approval stands; I will merge once these land and the gate is green.

snokvist:
All five are in at `737b4c6`.

## 1. Async ring ceiling — the one that mattered

You were right about why it hid: the `mtu` sweep only ever ran `mt_bulk`, because `mt_tx_raw()` takes the ring **only when one is running** and that gate never started RX. So the bench measured the path the integration does not use.

`MT_TX_BUFSZ` is now `MT_USB_AGG_BUF` — one ceiling — with a compile-time check that neither builder can outgrow it. Mutation-tested: putting the ring back to 2048 fails the *build*, not the link. Ring depth drops 32 → 16 so slots carrying a full aggregate cost 256 KB rather than 512 KB; depth is not what buys throughput here (`docs/mt7612u.md`, "Async rings bought no throughput") and `soak`/`ampdu` were re-measured after the change.

Rather than only add a static test, I made the **`mtu` gate sweep both paths and fail if they disagree**, so the blind spot cannot come back:

```
  bytes  sync       async      note
  2304   40/40      40/40
  4000   40/40      40/40      above the 802.11 MPDU ceiling
  4064   40/40      40/40      above the 802.11 MPDU ceiling
  4065   0/40       0/40       refused by this driver
```

The witness saw **80 frames per size** — 40 from each path — so both genuinely aired. `frame_shape` also carries the size-relationship test you asked for.

## 2–5

- **`gate_coding` / `gate_sweep` / `gate_vht`** now count short arms and fail; non-positive counts rejected outright.
- **`gate_init` / `gate_chan`** pass `fw_dir` through.
- **`open_selected()`** releases the lock when `libusb_open()` fails.
- **`g_stats_last_us`** moved into `struct mt7612u_dev`.

## Two more, found while verifying

**`bringup.c` named a private downstream consumer** in a comment. Described by role now — that should not have been in a public branch.

**`gate_duplex`'s verdict was unsatisfiable.** It required `(n && ctx.n)`, but the radio is half duplex and the gate saturates TX, so RX during the flood is ~0 — it has therefore never passed. I confirmed that is pre-existing by building the previous commit and reproducing it, rather than assuming I had caused it. Rewritten to assert what it can establish: TX sustained, no errors, and the receiver still delivering after the flood stops. It also now says plainly that it needs a peer.

## Verification, and what is *not* covered

18 of 19 gates pass. Ten self-check; `rtap` 779, `sweep` 1609 @20 / 1654 @`bw=1` / 816 @`bw=2`, `coding` 1920, `gateg` 450, `mtu` 600, `vht` 747 all witnessed by the RTL8812AU. `make check` green, extractor `--check` still reproduces `initvals.h`.

**The gap, stated rather than buried:** the MT7612U-to-MT7612U arm did not run. The second adapter is sitting in the RX-undrained wedge this port documents — EP0 reads answer, bulk-OUT times out — and only a physical replug clears it, which I cannot do remotely. So:

- the new ring **is** covered on air, by the `mtu` async pass against the witness;
- what is **not** re-verified since the ring change is MT-to-MT (it passed at 2998 frames earlier today, but on the pre-change ring);
- the reworked `duplex` assertion has **never** had a working peer to run against.

I chased that one badly before concluding it: three separate "RX is 0" readings turned out to be a dead peer, then a `txdemo` that emits a 10-frame burst rather than a stream, then a stale process holding the witness. The receiver itself is provably fine — 366 ambient frames on ch1. I will re-run both once the adapter is replugged and post the numbers.

snokvist:
Adapter replugged, so the two arms I owed you are done — and one of them corrects something I told you.

## Correction first

I said `gate_duplex`'s `(n && ctx.n)` verdict was **unsatisfiable** and had never passed. That was wrong. The peer adapter had silently failed its firmware load, so nothing was on air; every "RX is 0" reading behind that conclusion was a dead observation channel, not a result.

Measured with a live peer: **2894 fps out while 20–44 fps still arrives**, against 166 fps when the same peer is received with no local TX. So a saturated transmit throttles receive to roughly a quarter and does not stop it. The added post-flood check stays on its own merits — the in-flood count alone cannot tell a throttled receiver from a wedged one, or from a quiet channel — but the justification I committed asserted something false and is fixed in `f206354`.

## The two blocked arms

**MT7612U → MT7612U on the new ring: 3954 frames** across 20/40/80 MHz sweeps, `rx_err` `rx_invalid` `rx_dropped` all 0, HT 959 and VHT 1075 decoded. That closes the gap I flagged — the ring change is now verified peer-to-peer, not only against the witness.

**`duplex`: PASS** — 23153 frames out, receiver healthy afterwards.

## Full coverage, every gate

Self-checking: `regs` `fw` `init` `chan` `caps` `pwr` `hop` `linkstat` `ampdu` `soak` `rxbytes` `diversity` `linktx` `tx` `rx`. Peer-backed: `arx` 1097, `duplex`, `linkrx`. Witnessed by the RTL8812AU: `rtap` 776, `sweep` 1801 @`bw=0` / 839 @`bw=1` / 763 @`bw=2`, `coding` 1661, `vht` 925, `gateg` 450, `mtu` 600 across both TX paths.

Explicit arms for your findings: `#3` names the given `fw_dir` instead of defaulting; `#4` leaves no stale lock after a failed open; `#2` rejects a non-positive count with rc=2. `make check` green, extractor `--check` still reproduces `initvals.h`.

One methodology note, since it bit me twice more today: both adapters transmit with the **same source MAC**, so a leaked peer inflates every witness count. One sweep produced `sweep 80 → bw=0, 17500 frames`, which is impossible and was exactly that. The numbers above are from runs with a single transmitter proven.

## Stepping back — is it fit for purpose as a WLAN adapter?

Three gaps that gate-by-gate testing hid, none blocking, all worth naming:

1. **Every throughput figure in this port is transmit.** `soak` and `ampdu` report Mbit/s; nothing measures sustained **receive** throughput. For the obvious consumer — a ground station receiving video — that is the number that matters most, and it does not exist. The RX rates quoted anywhere here are peer-limited, not receiver-limited.

2. **A channel change has never been exercised with the receiver running.** `hop` measures retune cost with RX down. Retune-while-receiving is precisely the operation a hopping or channel-migrating consumer performs.

3. **TX and RX ceilings differ on the same silicon** — 4064 out, 3836 in — so two of these adapters have a 228-byte band where a frame transmits and vanishes. Both sit well above 802.11's 2304-byte non-A-MSDU limit, so standards-compliant traffic is unaffected; the dead band is entirely in already-oversized territory. Worth documenting rather than fixing, unless you would rather `max_mpdu_tx` were clamped to what an identical peer can receive.

Incidentally, the adapter lock proved itself in anger today: two sweeps collided on one adapter and the second **refused to claim it** rather than wedging it, which is the failure this port previously recovered from only with a physical replug.

src/mt7612u/init.c
```diff
@@ -0,0 +1,589 @@
...
+ */
+static uint64_t stats_now_us(void)
+{
+	struct timespec ts;
+
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	return (uint64_t)ts.tv_sec * 1000000u + (uint64_t)(ts.tv_nsec / 1000);
+}
+
+static uint64_t g_stats_last_us;
```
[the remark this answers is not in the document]
josephnef: Fixed at 737b4c6: the mark now lives in `struct mt7612u_dev` (`stats_last_us`, internal.h:154; init.c:558-559). Verified on f206354.

josephnef (approved):
Re-approving at f206354 (the push dismissed the earlier approval). Verified: all five re-review fixes present, `make check` in the subtree green, extractor `--check` reproduces `initvals.h`, every Qodo thread resolved.

josephnef:
The rename PR agreed above is open as #415: `IRtlDevice` → `IRadio`, `IRtlTransport` → `ITransport`, `CreateRtlDevice` → `CreateRadio`, with the Realtek-only members (`SetXtalCap`/`GetXtalCap`, `GetRxEnergy`, `ProbeEfuseStability`, `DumpChipState`) split onto `IRtlRadio : IRadio`. Hard rename, no aliases. The MT7612U integration (CMake option, `0e8d:7612` dispatch ahead of the `SYS_CFG2` read, a MediaTek transport, an `IRadio` implementation over `src/mt7612u` with the 1 Hz tick on its own thread) can land on top of it without touching core names again.

PR #414, opened 2026-09-08
mt7612u: fix the receive collapse (1 Hz PHY tick) and recover the USB wedge

snokvist:
## Summary

Follow-up to #412 (merged as `daabab7`). Library-internal — no `IRtlDevice`
surface. Two faults found by exercising the subtree the way a ground station
would, both device-verified, plus a hardware review of the whole delta.

## Public header (for the wrapper)

One addition, no changed signatures: **`int mt7612u_phy_tick(struct
mt7612u_dev *dev)`** in `include/mt7612u/mt7612u.h`. It is **required for
receive** — a consumer that receives must call it about once a second or the
front end goes deaf against a strong transmitter (§1). The `IRtlDevice` wrapper
should call it on a 1 Hz timer once it drives RX; everything else in the header
is unchanged, so the rename and integration PRs that only consume this header
are unaffected apart from adding that one call. `docs/mt7612u.md` is updated in
this PR for the lock and the RX-undrained recovery story, as requested.

## 1. A receiver went deaf against a strong transmitter

Against a peer ~20 cm away airing 1400-byte HT-MCS7 at 3037 fps, `arx` took
**3 frames in 10 s**. Not the adapter (symmetric both directions; both hear
400–1000 fps of ambient on ch1/36/40/44/48), not the ring (its own counter
agreed at 3), not the air (an RTL8812AU witness counted 103805 frames from the
same transmitter). mt76 runs `cal_work` every second
(`MT_CALIBRATE_INTERVAL == HZ`, `mt76x2/usb_phy.c:42`); this port ran none of
it. Bisected with one verified peer per arm:

| arm | frames / 10 s |
|---|---|
| nothing | 3 (reproduced 8×) |
| `MT_RX_STAT_*` reads only | 3 |
| a periodic MCU calibration only | 43902 |
| the ported gain tracking only | 4 |
| the full tick | 49002 |

New public `mt7612u_phy_tick()`, caller-driven (MCU commands share one 4-bit
sequence number and one response endpoint), called once a second from every
receiving path. Final reliability, both adapters as receiver, single verified
peer: **eight consecutive runs at 5415–5470 fps.** The MCU answers slowly under
RF load — nine calibration timeouts per bring-up on this bench — so
`mcu_wait_resp()` waits 10×300 ms and `mt_mcu_send()` drains any late reply off
EP 5 before the next command (without which the reply channel desynced and
cascaded `mcu resp mismatch … (want 1)`). The calibration takes regardless of
those timeouts; the channel is marked calibrated once, as mt76 does.

## 2. A run killed mid-transfer left the adapter unable to load firmware

Register reads *and* writes still round-trip, MAC and RF are fine, and the
kernel `mt76x2u` driver cannot bind it either — every bulk OUT NAKs.
`libusb_reset_device` (already on every open) does not reach it. Isolated one
step at a time against a freshly wedged adapter: `clear_halt` on all four
endpoints, MAC + USB DMA stop, WLAN_EN/WLAN_CLK_EN down, and the PBF block reset
each left it wedged; pulsing `MT_USB_DMA_CFG_TX_CLR` — a bit mt76 declares and
never writes — clears it alone. That plus an RX-pipe drain is `mt_recover_usb()`,
which runs after identify on **both** open paths — `mt_open()` and `mt_adopt()` —
so a libusb-owning consumer arriving through `mt7612u_open_handle()` self-heals
too. (Review fix: it lived in `mt_open()` alone, which was precisely the
consumer this subtree exists for getting none of it. `bringup adopt` now
exercises that path.)

A separately-observed **transient**: a post-kill firmware upload fails for 3–8 s
then succeeds by itself (12/12 kills, no repair applied) — a settling window,
not a wedge. `mt_fw_init()` absorbs it with one bounded retry after a 3 s
settle. On a healthy idle adapter the recovery is a no-op: 0/40 healing-open
failures.

**Known issue, called out honestly:** a harder tier — `TX_BUSY` **stuck** at
`0x80c00020` — was seen once and is **not** cleared by any of the above; it
needed a physical replug, and it has not reproduced since (36+ kills).
`docs/mt7612u-usb-wedge.md` traces the vendor tree's separate UDMA and IFDMA/FCE
resets that none of the attempts reached, implemented as `bringup swreset` and
still awaiting a reproduction to test against. `mt_open()` warns clearly when it
hits this state rather than failing with an opaque timeout.

## Verification

- Receiver: 8/8 runs 5415–5470 fps, both adapters.
- Wedge recovery: soft tier 3/3; healthy-adapter soak 0/40; short stress on the
  shipping binary 0/12 across the scenarios that actually wedge.
- Post-kill transient: 12/12 clear by themselves, absorbed by the loader retry.
- `mt_mac_start()` now refuses to enable the receiver with no ring draining
  (`MT_RX_DRAIN_*`; the public `mt7612u_start()` always derived this, three
  gates bypassed it).
- `make check` green; strict `-Wall -Wextra -Wshadow -Wmissing-prototypes`
  build clean.

## Retracted in the course of this

Two claims made and then falsified: that the inverted ring/receiver order in
`arx` explained the 3 frames (it didn't), and that a synchronous MCU command
"hangs forever" beside the RX ring's event thread — it never reproduced in 8
stack-capture attempts, and libusb's own source shows that shape is supported.
The comments say so where they used to say otherwise.

## Not in scope

The vendor-neutral rename, `IRtlDevice` wiring, and a proven recovery for the
stuck-`TX_BUSY` tier (needs a reproduction first).

---

## Review round (head `4a6a14c`)

Both blocking items and all eight inline points are addressed; see the review
reply for the point-by-point. Headlines: the recovery is shared by both open
paths, teardown clears `ENABLE_RX` before the EP 4 ring is cancelled, the EP 5
drain is armed rather than unconditional, the ambient-RSSI EMA is gone in favour
of mt76's monitor behaviour, and the library no longer reads
`MT7612U_NO_AUTORECOVER` (`MT7612U_DEV` stays deferred per #412).

**Re-verified on this head** — a different stimulus from §1's, so quoted
separately rather than folded into that figure. Peer: a second MT7612U flooding
ch149; DUT: MT7612U at `2-1`.

| check | result |
|---|---|
| RX rate | **4902–4913 fps**, 5 runs, `rx_err`/`rx_invalid`/`rx_dropped` all 0, peer verified airing before and after every run |
| adopt path | `bringup adopt` brings the device up; with `MT7612U_NO_AUTORECOVER=1` the recovery announces itself **on that path** |
| MCU drain | 9 calibration timeouts under RF load produced **one** drain of 9 replies, where the old code paid a 5 ms timeout before every command |
| retune | 530 ms full / 50.6 ms fast |
| build | strict `-Wall -Wextra -Wshadow -Wmissing-prototypes` clean; `make check` green |

**Not** reproduced this session: the soft USB wedge (5 kill cycles with TX+RX up,
`U3DMA_CFG` stayed healthy each time). So "the adopt path recovers a *wedged*
adapter" is a code-path fact — one shared function, two call sites — not a
measurement from this round.

### DIFF
--- a/docs/mt7612u.md
+++ b/docs/mt7612u.md
@@ -157,6 +157,29 @@ len=102  HT    mcs=15 nss=2 bw=20 sgi=1 ldpc=1 stbc=0 rssi=[-71,-63]
 `nss` is derived as `1 + (15 >> 3)`; SGI and LDPC come from the same 16-bit
 rate word the TX path writes, so one codec serves both directions.
 
+### A receiver needs the 1 Hz tick
+
+mt76 runs `cal_work` every second (`MT_CALIBRATE_INTERVAL == HZ`,
+`mt76x2/usb_phy.c:42`): a self-guarded channel calibration, TSSI compensation
+(an MCU command each second) and `mt76x2_phy_update_channel_gain()`. This
+port ran none of it, and the cost was measured rather than assumed: against a
+peer 20 cm away airing 1400-byte HT-MCS7 frames at 3037 fps, a receiver with
+no tick took **3 frames in 10 s**; calling `mt7612u_phy_tick()` once a second
+it takes **~4850/s** (nine runs). Not the adapter — symmetric in both
+directions, and both hear 400–1000 fps of ambient on ch1/36/40/44/48 — not
+the ring (its own counter agreed at 3), and not the air: an RTL8812AU witness
+counted 103805 frames from the same transmitter.
+
+Bisected with one verified peer per arm: reading the read-and-clear
+`MT_RX_STAT_*` counters alone does nothing (3 frames); a periodic MCU
+calibration alone restores it (43902); the ported gain tracking alone does not
+(4 frames). The tick therefore issues one `MCU_CAL_TEMP_SENSOR` per call — the
+cheapest command that keeps the MCU in the loop, standing in for the TSSI
+compensation this port lacks — and runs the gain tracking for fidelity. It is
+caller-driven: MCU commands share one 4-bit sequence number and one response
+endpoint, so they need exactly one user. Every receiving gate in `bringup`
+ticks; `arx <chan> <secs> 1` is the negative control.
+
 ### This part does not deliver the FCS
 
 Measured, because it constrains integration rather than being a detail.
--- a/docs/mt7612u.md
+++ b/docs/mt7612u.md
@@ -2,25 +2,8 @@
 
 Everything below was measured on one MT7612U (`0e8d:7612`, `MT_ASIC_VERSION`
 `0x76120044`, MT7662 MAC core, 2T2R, SuperSpeed) against an RTL8812AU witness
-running this project's own `rxdemo`/`txdemo`. Read `## Offline tests
-
-`make -C src/mt7612u check` runs three binaries. No hardware, no privileges.
-
-| test | what it holds |
-|---|---|
-| `api_link` | takes the address of all 20 public entry points while including only the public header, so a declaration that loses its definition is a link error |
-| `frame_shape` | `mt_hdrlen_from_fc()` over management, all eight control subtypes and the five data shapes; the RX L2-pad fold on a synthetic QoS frame, with a negative control that redoes the old fixed-24 fold and asserts the QoS Control really is destroyed; the radiotap VHT bandwidth mapping over all eleven codes the part can express |
-| `field_macros` | `MT_CTZ` against `__builtin_ctz` over all 32 single-bit and all 528 contiguous masks, plus a `FIELD_PREP`/`FIELD_GET` round-trip, plus a static initialiser that fails to compile if the macro stops being constant-foldable |
-
-Each was mutation-tested: removing one public definition, reverting either
-frame-shape fix, and reverting the header-length fix each make the suite fail,
-with the RX one reporting `QoS Control zeroed by the pad fold: aa aa`.
-
-`tools/extract_mt7612u_tables.py --check` byte-compares the generated
-`initvals.h` against `reference/mt76` at the pinned commit.
-
-## Counterparts` before
-quoting any number here.
+running this project's own `rxdemo`/`txdemo`. Read `## Offline tests` and
+`## Counterparts` before quoting any number here.
 
 **This code is not wired into the build.** `CMakeLists.txt` is untouched, there
 is no `IRtlDevice` implementation and no `WiFiDriver` dispatch. It is a
@@ -159,15 +142,15 @@ rate word the TX path writes, so one codec serves both directions.
 
 ### A receiver needs the 1 Hz tick
 
-mt76 runs `cal_work` every second (`MT_CALIBRATE_INTERVAL == HZ`,
-`mt76x2/usb_phy.c:42`): a self-guarded channel calibration, TSSI compensation
-(an MCU command each second) and `mt76x2_phy_update_channel_gain()`. This
-port ran none of it, and the cost was measured rather than assumed: against a
-peer 20 cm away airing 1400-byte HT-MCS7 frames at 3037 fps, a receiver with
-no tick took **3 frames in 10 s**; calling `mt7612u_phy_tick()` once a second
-it takes **~4850/s** (nine runs). Not the adapter — symmetric in both
-directions, and both hear 400–1000 fps of ambient on ch1/36/40/44/48 — not
-the ring (its own counter agreed at 3), and not the air: an RTL8812AU witness
+The cadence, what it costs to skip it and why it is caller-driven are the
+contract on `mt7612u_phy_tick()` in the public header, which is where the
+numbers live. Kept here: why the finding is believed, and which part of the
+tick does the work.
+
+Stimulus throughout is a peer 20 cm away airing 1400-byte HT-MCS7 frames at
+3037 fps. The no-tick collapse is not the adapter — it is symmetric in both
+directions, and both hear 400–1000 fps of ambient on ch1/36/40/44/48; not the
+ring — its own counter agreed at 3; and not the air — an RTL8812AU witness
 counted 103805 frames from the same transmitter.
 
 Bisected with one verified peer per arm: reading the read-and-clear
@@ -175,10 +158,9 @@ Bisected with one verified peer per arm: reading the read-and-clear
 calibration alone restores it (43902); the ported gain tracking alone does not
 (4 frames). The tick therefore issues one `MCU_CAL_TEMP_SENSOR` per call — the
 cheapest command that keeps the MCU in the loop, standing in for the TSSI
-compensation this port lacks — and runs the gain tracking for fidelity. It is
-caller-driven: MCU commands share one 4-bit sequence number and one response
-endpoint, so they need exactly one user. Every receiving gate in `bringup`
-ticks; `arx <chan> <secs> 1` is the negative control.
+compensation this port lacks — and runs the gain tracking for fidelity. Every
+receiving gate in `bringup` ticks; `arx <chan> <secs> 1` is the negative
+control.
 
 ### This part does not deliver the FCS
 
@@ -266,13 +248,17 @@ three defined values (20/40/80) and mt76 exposes no narrowband path for this
 part.
 
 **Enabling MAC RX without draining the bulk-IN endpoint wedges the chip below
-USB level.** Prevented at the source: `mt_mac_start()` takes an `MT_RX_DRAIN_*`
-argument and refuses to enable the receiver unless a ring is draining EP 4 (the
-public `mt7612u_start()` always derived this; the internal entry point now does
-too).
+USB level.** Prevented at both ends. Starting: `mt_mac_start()` takes an
+`MT_RX_DRAIN_*` argument and refuses to enable the receiver unless a ring is
+draining EP 4 (the public `mt7612u_start()` always derived this; the internal
+entry point now does too). Stopping: `mt_mac_rx_disable()` clears `ENABLE_RX`
+**before** the ring is cancelled, because `mt_async_stop()` reaps the EP 4
+drainer while `mt_mac_stop()` would not clear the bit until after its own
+`mt_rx_flush()` and a TX-idle wait of up to 150 ms — the same undrained window,
+at the end of every session.
 
 **A run that dies mid-transfer leaves the adapter unable to load firmware, and
-`mt_open()` now recovers it.** Register reads *and* writes still round-trip and
+both open paths now recover it.** Register reads *and* writes still round-trip and
 the MAC and RF are fine, but every bulk OUT NAKs — `libusb_reset_device()`
 (already called on every open) does not reach it, and the kernel `mt76x2u`
 driver cannot bind it either. Two tiers, told apart by `MT_USB_U3DMA_CFG`:
@@ -281,11 +267,15 @@ driver cannot bind it either. Two tiers, told apart by `MT_USB_U3DMA_CFG`:
   wedged adapter: `clear_halt` on all four endpoints, MAC + USB DMA stop,
   `WLAN_EN`/`WLAN_CLK_EN` down, and the PBF block reset each left it wedged;
   pulsing `MT_USB_DMA_CFG_TX_CLR` — a bit mt76 declares and never writes —
-  clears it alone. `mt_open()` pulses it and drains the RX pipe (`mt_rx_flush`
-  runs from `mt_mac_stop`, which a killed process never reaches), so every
-  entry point self-heals. A post-kill firmware upload can also fail for 3–8 s
-  and then succeed by itself; `mt_fw_init()` absorbs that with one retry after
-  a settle. On a healthy idle adapter the recovery is a no-op (0/40 opens).
+  clears it alone. `mt_recover_usb()` pulses it and drains the RX pipe
+  (`mt_rx_flush` runs from `mt_mac_stop`, which a killed process never
+  reaches). It runs after identify on **both** open paths — `mt_open()` and
+  `mt_adopt()` — so a libusb-owning consumer arriving through
+  `mt7612u_open_handle()` self-heals too; while it lived in `mt_open()` alone
+  that consumer got none of it. `bringup adopt` exercises that path. A post-kill firmware upload can also fail for 3–8 s
+  and then succeed by itself; `mt_fw_init()` absorbs that with one retry of
+  the upload after a settle. On a healthy idle adapter the recovery is a no-op
+  (0/40 opens).
 
 - **Hard (`TX_BUSY` stuck, `0x80c00020`).** Seen once; not cleared by any of
   the above, nor by the kernel driver — a physical replug was required, and it
@@ -297,10 +287,13 @@ driver cannot bind it either. Two tiers, told apart by `MT_USB_U3DMA_CFG`:
 
 **Locking.** The 1 Hz `mt7612u_phy_tick()` issues MCU commands, and the RX ring
 runs its own libusb event thread, so register and MCU I/O are serialised by a
-recursive `io_lock` initialised in `mt_open()`. A single-threaded consumer never
-contends; the lock is what lets a consumer safely drive the tick from a second
-thread. `mt_mcu_send()` also drains any late reply off EP 5 before each command,
-without which a reply that arrived after its wait expired desynced the channel.
+recursive `io_lock` initialised in `mt_dev_state_init()`, which both open paths
+run before any register I/O. A single-threaded consumer never contends; the lock
+is what lets a consumer safely drive the tick from a second thread.
+`mt_mcu_send()` also drains any late reply off EP 5, but only when a previous
+command actually gave up (`mcu_stale_pending`) — draining before *every* command
+cost a guaranteed 5 ms bulk timeout each, including the 3-6 inside every tune,
+and nothing was ever queued on the healthy path.
 
 ## Offline tests
 
@@ -308,7 +301,7 @@ without which a reply that arrived after its wait expired desynced the channel.
 
 | test | what it holds |
 |---|---|
-| `api_link` | takes the address of all 20 public entry points while including only the public header, so a declaration that loses its definition is a link error |
+| `api_link` | takes the address of all 25 public entry points while including only the public header, so a declaration that loses its definition is a link error |
 | `frame_shape` | `mt_hdrlen_from_fc()` over management, all eight control subtypes and the five data shapes; the RX L2-pad fold on a synthetic QoS frame, with a negative control that redoes the old fixed-24 fold and asserts the QoS Control really is destroyed; the radiotap VHT bandwidth mapping over all eleven codes the part can express |
 | `field_macros` | `MT_CTZ` against `__builtin_ctz` over all 32 single-bit and all 528 contiguous masks, plus a `FIELD_PREP`/`FIELD_GET` round-trip, plus a static initialiser that fails to compile if the macro stops being constant-foldable |
 
--- a/docs/mt7612u.md
+++ b/docs/mt7612u.md
@@ -250,6 +250,11 @@ a caller that will not drain it, plus an endpoint flush; 20 consecutive
 init+TX cycles clean afterwards against a death after ~5 before. **Two things
 changed at once**, so that run does not attribute the wedge to one of them.
 
+For the separate **EP0-responsive, bulk-OUT-stalled** failure, see
+[USB TX wedge reset evidence](mt7612u-usb-wedge.md): the vendor-derived
+MT76x2U tree has dedicated UDMA and IFDMA/FCE resets beyond `TX_CLR`.
+Their recovery of the hard `TX_BUSY` state is not yet hardware-verified.
+
 ## Offline tests
 
 `make -C src/mt7612u check` runs three binaries. No hardware, no privileges.
--- a/docs/mt7612u.md
+++ b/docs/mt7612u.md
@@ -266,17 +266,41 @@ three defined values (20/40/80) and mt76 exposes no narrowband path for this
 part.
 
 **Enabling MAC RX without draining the bulk-IN endpoint wedges the chip below
-USB level.** Every vendor request times out afterwards, and neither
-`libusb_reset_device()`, the `authorized` toggle, nor the kernel driver
-recovers it — only a physical replug does. Fixed here by never enabling RX for
-a caller that will not drain it, plus an endpoint flush; 20 consecutive
-init+TX cycles clean afterwards against a death after ~5 before. **Two things
-changed at once**, so that run does not attribute the wedge to one of them.
+USB level.** Prevented at the source: `mt_mac_start()` takes an `MT_RX_DRAIN_*`
+argument and refuses to enable the receiver unless a ring is draining EP 4 (the
+public `mt7612u_start()` always derived this; the internal entry point now does
+too).
 
-For the separate **EP0-responsive, bulk-OUT-stalled** failure, see
-[USB TX wedge reset evidence](mt7612u-usb-wedge.md): the vendor-derived
-MT76x2U tree has dedicated UDMA and IFDMA/FCE resets beyond `TX_CLR`.
-Their recovery of the hard `TX_BUSY` state is not yet hardware-verified.
+**A run that dies mid-transfer leaves the adapter unable to load firmware, and
+`mt_open()` now recovers it.** Register reads *and* writes still round-trip and
+the MAC and RF are fine, but every bulk OUT NAKs — `libusb_reset_device()`
+(already called on every open) does not reach it, and the kernel `mt76x2u`
+driver cannot bind it either. Two tiers, told apart by `MT_USB_U3DMA_CFG`:
+
+- **Soft (`TX_BUSY` clear).** Isolated one step at a time against a freshly
+  wedged adapter: `clear_halt` on all four endpoints, MAC + USB DMA stop,
+  `WLAN_EN`/`WLAN_CLK_EN` down, and the PBF block reset each left it wedged;
+  pulsing `MT_USB_DMA_CFG_TX_CLR` — a bit mt76 declares and never writes —
+  clears it alone. `mt_open()` pulses it and drains the RX pipe (`mt_rx_flush`
+  runs from `mt_mac_stop`, which a killed process never reaches), so every
+  entry point self-heals. A post-kill firmware upload can also fail for 3–8 s
+  and then succeed by itself; `mt_fw_init()` absorbs that with one retry after
+  a settle. On a healthy idle adapter the recovery is a no-op (0/40 opens).
+
+- **Hard (`TX_BUSY` stuck, `0x80c00020`).** Seen once; not cleared by any of
+  the above, nor by the kernel driver — a physical replug was required, and it
+  has not reproduced since. The vendor-derived MT76x2U tree has dedicated UDMA
+  and IFDMA/FCE resets beyond `TX_CLR`
+  ([USB TX wedge reset evidence](mt7612u-usb-wedge.md)), implemented as
+  `bringup swreset` and **not yet hardware-verified** for want of a
+  reproduction.
+
+**Locking.** The 1 Hz `mt7612u_phy_tick()` issues MCU commands, and the RX ring
+runs its own libusb event thread, so register and MCU I/O are serialised by a
+recursive `io_lock` initialised in `mt_open()`. A single-threaded consumer never
+contends; the lock is what lets a consumer safely drive the tick from a second
+thread. `mt_mcu_send()` also drains any late reply off EP 5 before each command,
+without which a reply that arrived after its wait expired desynced the channel.
 
 ## Offline tests
 
--- a/src/mt7612u/fw.c
+++ b/src/mt7612u/fw.c
@@ -17,6 +17,18 @@
 #define PATCH_HDR_LEN               30
 #define FW_HDR_LEN                  32
 
+/* The two loaders below fail in two ways that call for different handling, so
+ * they say which one they hit: a blob that is absent or malformed will fail the
+ * same way forever, while the bulk upload itself is the transfer mt_fw_init()
+ * retries. */
+#define FW_ERR_FATAL   (-1)   /* blob missing, short or malformed; or no memory */
+#define FW_ERR_UPLOAD  (-2)   /* the bulk upload failed on the wire */
+/* The chip did not come up after a complete upload.  A DEVICE failure, not a
+ * blob one: it is the same settling window the upload retry was measured
+ * against, so it must retry too - `bringup swreset`'s whole verdict is
+ * whether the chip takes firmware again, and it used to settle and pass. */
+#define FW_ERR_DEVICE  (-3)   /* uploaded, but the chip did not start it */
+
 static void put_le32(uint8_t *p, uint32_t v)
 {
 	p[0] = v & 0xff; p[1] = (v >> 8) & 0xff;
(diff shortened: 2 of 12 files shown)


snokvist:
Thanks — several of these are real and are now fixed in two commits on the branch. Point by point:

**#3 (adopted devices use an invalid mutex) — fixed, and it was the important one.** `mt_adopt()` → `mt_identify()` → `mt_vendor_req()` locks `io_lock`, which was initialised only in `mt_open()`. A zeroed `pthread_mutex_t` is a valid *non-recursive* lock, so the first `mt7612u_phy_tick()` (which locks `io_lock` and nests `mt_vendor_req`/`mt_mcu_send`) would self-deadlock a device opened by a libusb-owning consumer — i.e. exactly the `open_handle` integration path. A shared `mt_dev_state_init()` now runs first on both open paths, and a guarded `mt_dev_state_destroy()` folded into `mt_close()` releases it exactly once. This also closed a leak the review didn't name: every `mt_open()` failure path and `bring_up()`'s fail path previously freed the device without destroying the mutex.

**#4 (adopted receivers use invalid gain) — fixed, consolidated with #6.** The gain sentinels (`low_gain=-1`, `avg_rssi_all=0`, `agc_gain_adjust=0`) are now reset per-tune in `mt_set_channel_ex()` instead of once in `mt_open()`. Since the tick refuses to run before a channel is set, and a channel is only ever set through that function, this covers the adopted handle and the retune case with one reset.

**#6 (retunes reuse the previous gain state) — fixed** by that same per-tune reset.

**#9 (failed tests leave USB transfers running) — fixed.** The reordered gates now call `mt7612u_rx_stop()` on the `mt_mac_start()` failure path (`gate_arx`/`gate_duplex`/`gate_rxbytes`/`gate_linkrx`/`gate_linkstat`), and `gate_linkstat`'s mid-loop `mt7612u_link_stats()` failure return does too — thanks, that second return is the same bug and wasn't in the list.

**#5 (RX parsing races gain tracking) — fixed as an atomic, not a lock.** `cal.avg_rssi_all` is now `_Atomic int8_t` with relaxed load/store. Deliberately *not* taking `io_lock` on the RX path: that would block receive behind the tick's multi-second MCU calibration.

**#7 (some receive tools miss ticks) — fixed where it was real.** `gate_duplex` now ticks once a second inside its TX flood. `gate_linkstat` already issues an MCU calibration every second via `mt7612u_link_stats()` (that calibration is the part that keeps RX alive — the gain half doesn't), and `gate_caps` measures registers rather than RX rate and only needs the ring *drained*, so neither was changed.

**#8 (full reply queue stays desynced) — hardened.** The stale-reply drain cap is raised from 16 to 64 (observed depth is ~5) so a deeper transient drains fully, with a distinct warning if the cap is actually hit (which would mean a genuinely desynced channel).

**#1 (env var) and #2 (doc duplication) — no change, and here's why.** `MT7612U_NO_AUTORECOVER` is opt-in, read only in `mt_open()`, announced with a log line, and mirrors the already-shipping `MT7612U_DEV` env convention in the same file; the adopt path never reads it. Threading a config parameter through `open()` for a debug toggle seemed like more surface than the toggle warrants. And `docs/mt7612u.md`'s tick section is the measurement/bisection narrative (which sub-step restores reception), not the API contract — the header stays authoritative and the doc keeps the empirical justification that doesn't belong in a header. Happy to revisit either if you'd prefer.

Verification: `make check` green, strict `-Wall -Wextra -Wshadow -Wmissing-prototypes` clean, and the receiver re-measured at 5422–5470 fps on both adapters after the changes (7 of 8 runs; the one miss is the first run's cold-start settling right after a driver unload, which recovers on the next run).

src/mt7612u/usb.c
```diff
@@ -279,10 +286,37 @@ static int mt_identify(struct mt7612u_dev *d, const char **err)
...
+}
+
+void mt_dev_state_destroy(struct mt7612u_dev *d)
+{
+	if (!d || !d->io_lock_ready) return;
+	pthread_mutex_destroy(&d->io_lock);
+	d->io_lock_ready = 0;
+}
+
 int mt_adopt(struct mt7612u_dev *d, libusb_device_handle *h,
```
josephnef, line 400: **Blocking.** `mt_adopt()` returns straight from `mt_identify()`. The recovery block (`MAC_SYS_CTRL=0`, `mt_rx_flush`, the TX_CLR pulse loop, the `TX_BUSY` verdict) lives only in `mt_open()` below. `mt7612u_open_handle()` is the path a libusb-owning consumer — the `IRtlDevice` wrapper — uses, so after a killed run it sees `U3DMA_CFG=0x00c00020`, nothing pulses TX_CLR, and `mt_fw_init` times out twice (now +3 s) with the exact failure this PR fixes for `mt7612u_open()`. Factor the recovery into a helper both paths run after identify, the way `mt_dev_state_init` already is.

src/mt7612u/init.c
```diff
@@ -466,7 +479,7 @@ void mt7612u_close(struct mt7612u_dev *d)
 	if (!d) return;
 	mt_async_stop(d);
```
josephnef, line 502: **Blocking.** Ordering inverts the invariant `mt_mac_start()` now enforces: `mt_async_stop()` cancels and reaps the EP 4 ring here, and MAC RX stays enabled until `mt_mac_stop()` clears `MT_MAC_SYS_CTRL` — after its `mt_rx_flush` and the up-to-150 ms TX-idle wait (longer if a transfer is stuck and the ring strands). On a busy channel that window is the undrained-EP4 overflow you describe as stopping RX DMA for good, and it runs at the end of every session. Same shape in every bringup `mt7612u_rx_stop(); mt_mac_stop();` pair. Clear `MT_MAC_SYS_CTRL_ENABLE_RX` (or run `mt_mac_stop`) before cancelling the ring.

src/mt7612u/phy.c
```diff
@@ -461,6 +631,14 @@ int mt_set_channel_ex(struct mt7612u_dev *d, uint8_t chan, uint8_t bw, int fast)
 	d->cal.channel_cal_done = fast;
 	d->chan = chan;
 	d->bw = bw;
+	/* Reset the periodic gain tracker for the new channel: its cached RSSI,
+	 * gain class and fine VGA offset all belong to the old channel/band.
+	 * low_gain=-1 forces the first tick to program a class (this is also
+	 * what makes an adopted handle's first tick valid - see mt_dev_state_init).
+	 * avg_rssi_all is written on the RX thread, so store it atomically. */
```
josephnef: `low_gain = -1` here forces the next tick to reprogram AGC 8/9 from `agc_gain_init` — but the fast path (`if (fast) return 0;` at ~line 717) returns before the `agc_gain_init` snapshot is re-read, while `mt_mcu_init_gain` (both paths) has just programmed AGC 8/9 for the new channel. So the first tick after a fast retune writes the *previous* full tune's gain base. mt76 re-reads `agc_gain_init` on every set_channel. Move the two `mt_rr(AGC 8/9)` reads above the fast return.

src/mt7612u/rx.c
```diff
@@ -104,6 +104,16 @@ int mt_rx_parse(struct mt7612u_dev *d, uint8_t *buf, int n,
...
 	 * channel. */
 	info->noise = info->rssi[2];
+	/* mt76_get_min_avg_rssi() equivalent, on the RX (libusb event) thread;
+	 * the PHY tick reads it on the caller's thread.  Relaxed atomic load+store:
+	 * a torn value only delays a gain-class change by ~1 s, and taking io_lock
+	 * here would block RX behind the tick's multi-second MCU calibration. */
+	{
+		int8_t avg = atomic_load_explicit(&d->cal.avg_rssi_all,
+		                                  memory_order_relaxed);
+		avg = avg ? (int8_t)((avg * 7 + info->rssi[0]) / 8) : info->rssi[0];
```
josephnef: This EMA takes `rssi[0]` of every frame the monitor receiver parses — third-party transmitters included (and CRC-error frames, if the filter admits them). mt76's `mt76_get_min_avg_rssi` averages associated stations only; in monitor mode there are none, so mt76 sits at the -75 fallback and never leaves the middle gain class. Here a -40 dBm neighbour AP drives `low_gain=2`, `AGC 35/37=0x08080808` and `gain_cur -= 10..14` against a wanted peer at -80 dBm. The bisect says gain tracking is not what restores RX, so this is fidelity-only code that is currently less faithful than a fixed -75. Either pin it to mt76's monitor behaviour, or feed it only from frames that pass the consumer's SA filter. (Minor: the int8 divide truncates toward zero, and an average that lands on exactly 0 reseeds to -75.)

src/mt7612u/mcu.c
```diff
@@ -55,12 +60,46 @@ int mt_mcu_send(struct mt7612u_dev *d, int cmd, const void *data, int len,
...
+	 * cascading through every 1 Hz tick.  The loop exits when the queue is
+	 * empty (a 5 ms read returns nothing); observed depth is ~5, but the cap
+	 * is generous so a deeper transient is fully drained rather than leaving
+	 * a straggler that re-desyncs the next command.  Hitting the cap means
+	 * the queue is still non-empty - a real fault, warned distinctly. */
+	{
+		uint8_t stale[MCU_RESP_URB_SIZE];
+		int n = 0, got;
+
+		while (n < 64 &&
```
josephnef: This 5 ms bulk-IN timeout is paid before *every* command, including the 3-6 inside `mt_set_channel_ex` (fast retune included) and no-wait commands that never produce a reply — a guaranteed `LIBUSB_ERROR_TIMEOUT` each, so every hop is +15-30 ms with nothing ever drained in the healthy case. Drain only when the previous `mcu_wait_resp` actually timed out (one `d->mcu_stale_pending` flag set on that path): zero cost on the hot path, same protection.

src/mt7612u/usb.c
```diff
@@ -513,6 +550,60 @@ int mt_open(struct mt7612u_dev *d, const char **err)
...
+	 *   UDMA/IFDMA reset sequence in docs/mt7612u-usb-wedge.md (bringup
+	 *   swreset) is the untested candidate.  The WARN is the honest verdict.
+	 *
+	 * The receive direction gets its own clean-up: mt_rx_flush() runs from
+	 * mt_mac_stop(), which a killed process never reaches.  Silence the
+	 * receiver BEFORE draining or it refills as fast as it is read. */
+	mt_wr(d, MT_MAC_SYS_CTRL, 0);
+	mt_rx_flush(d);
+
+	if (mt_rr(d, CFG_ADDR(MT_USB_U3DMA_CFG)) & MT_USB_DMA_CFG_TX_BUSY) {
```
josephnef: `mt_rr()` returns `~0u` on a failed control transfer, which has `TX_BUSY` set: one EP0 hiccup here sends a healthy adapter through `TX_BULK_EN` off, 20 TX_CLR pulses (400 ms) and a false "still busy" WARN when the re-read fails too. Same in `phy_adjust_vga_gain`: a failed `MT_RX_STAT_1` read is 65535 false CCAs and a 2 dB gain step down. Use `mt_rr_chk` on these two.

src/mt7612u/fw.c
```diff
@@ -231,7 +231,22 @@ static int load_firmware(struct mt7612u_dev *d, const char *dir)
...
-	return load_firmware(d, fw_dir);
+
+	/* Two attempts, not one.  Measured (12/12 cycles, both adapters): after
+	 * a process dies mid transfer the first firmware upload times out at
+	 * its first chunk, and the same upload succeeds by itself 3-8 s later
+	 * with nothing repaired.  That is a settling window, not a wedge, and
+	 * an open that lands inside it must not fail.  One bounded retry after
+	 * a settle covers it; a fault that survives the retry is reported as
+	 * before, and that is the case the open-path recovery is for. */
+	for (int attempt = 0; attempt < 2; attempt++) {
```
josephnef: The retry keys on any non-zero return, so a missing or short blob (`slurp` fails, "rom patch too short") now also WARNs that a previous run died mid-transfer and sleeps 3 s before failing identically — sending the user to the wedge doc instead of to the missing file. Retry only on the upload-transfer failure (a distinct return from `fw_send_data`), which is the case you measured.

src/mt7612u/tools/bringup.c
```diff
@@ -393,7 +501,13 @@ static int gate_rx(uint8_t chan, int want)
 	printf("  MT_RX_STAT_1     = 0x%08x (CCA errors seen = RF is live)\n",
 	       mt_rr(&dev, MT_RX_STAT_1));
 
+	double last_tick = now_ms();
+
 	while (got < want && empty < 200) {
+		if (now_ms() - last_tick >= 1000.0) {
+			mt7612u_phy_tick(&dev);
```
josephnef, line 666: `gate_rx` runs `MT_RX_DRAIN_SYNC` — this thread is the only EP 4 drainer — and the tick now blocks it for up to ~3.3 s when the MCU answers late (your own measurement under RF load). That is the undrained-receiver window on the sync gate. Bringup-only, but it is the negative control for the very fault, so either stop RX around the tick here or note that `gate_rx` is not a valid tick witness under load.

josephnef (changes requested):
Read every commit and the head diff against `daabab7`, walked the vendor gain code and the lock paths, checked the nine open Qodo threads against the head. The RX-collapse work is measured and the bisect table is convincing; the lock/adopt fixes from the Qodo round are in. Two things block for me, both because the PR's own contract ("every entry point self-heals", "the receiver must never come up with nothing draining EP 4") is not yet met by the code. The rest are inline.

**Blocking**

1. **The wedge recovery is unreachable from `mt7612u_open_handle()`** (inline at `usb.c` `mt_adopt`). The whole TX_CLR + MAC-off + RX-flush block sits in `mt_open()` only. The `IRtlDevice` wrapper — the consumer this subtree exists for — owns libusb and will come in through `open_handle` / `mt_adopt`, and gets none of it; after a killed run it pays both `mt_fw_init` attempts and fails with exactly the message this PR fixes. Hoist the recovery into a shared post-identify step both paths run (like `mt_dev_state_init` already is).

2. **Teardown still runs the receiver undrained** (inline at `init.c` `mt7612u_close`). `mt_async_stop()` cancels and reaps the EP 4 ring first; MAC RX stays enabled until `mt_mac_stop()` clears `MT_MAC_SYS_CTRL`, which happens after its own `mt_rx_flush` and the TX-idle wait. Every `rx_stop(); mac_stop();` pair in bringup has the same shape. That is the condition the new `mt_mac_start` guard exists to prevent, on the path that runs at the end of every session. Pre-existing, but this PR is the one that names the invariant. Clearing `ENABLE_RX` before cancelling the ring closes it.

**Non-blocking, worth doing in this PR**

- Fast retune leaves `agc_gain_init` stale while forcing a reprogram (inline `phy.c:717`).
- `avg_rssi_all` is fed by every decoded frame, neighbours included; mt76 in monitor mode has no stations and sits at the -75 class forever (inline `rx.c:114`). Your own bisect says gain tracking is not what fixes RX, so this is fidelity-only code that is currently *less* faithful than doing nothing.
- The 5 ms stale-reply drain runs before every MCU command, including the 3-6 inside every tune (inline `mcu.c:91`).
- Two `mt_rr()` results are consumed without the `~0u` sentinel check on the new paths (inline `usb.c:585`).
- `mt_fw_init` retry fires on any failure, so a missing blob now costs 3 s and blames a crashed run (inline `fw.c:242`).
- `gate_rx` ticks on the thread that is the only EP 4 drainer (inline `bringup.c:508`).

**Process**

- The `qodo-gate` check is red only because four threads are unresolved: #1 `getenv` (usb.c:556), #2 docs duplication, #6 retunes (fixed on head — resolve it), #8 drain cap (head warns distinctly and sends anyway — reply with that and resolve). On #1: this repo's rule is that the library reads no environment; `MT7612U_DEV` was already deferred to integration in #412, so this is the second one accumulating. A field on `mt7612u_open()`/a `bringup`-only flag would end the debt; if you defer again, say so in the thread and resolve it.
- The PR body says 8 runs at 5415–5470 fps, the header says ~4850 (nine runs), `phy.c` says 4362/s, `docs/mt7612u.md` says ~4850. Pick one figure and one home (the header) — Qodo #2 is really about this drift.
- `docs/mt7612u-usb-wedge.md` opens with a research date, a branch name and two commit hashes. Docs here are current-state; git is the changelog. Keep the register tables and the untested-status verdict, drop the provenance header.

No MT7612U on my bench, so nothing above is hardware-verified from this side; the blocking items are code-path facts, not measurements.

src/mt7612u/tools/bringup.c
```diff
@@ -136,6 +152,98 @@ static int gate_regs(void)
...
+	                                  : "OBSERVE ONLY (failing control)");
+	swreset_dump("before");
+
+	if (apply) {
+		/* The helper's surrounding contract: stop the MAC and let TX
+		 * drain before touching the DMA.  Bounded - this is the fault
+		 * under investigation, so it must not become an infinite wait. */
+		mt_wr(&dev, MT_MAC_SYS_CTRL, 0);
+		for (int i = 0; i < 50; i++) {
+			if (!(mt_rr(&dev, MT_MAC_STATUS) & BIT(0))) break;
```
josephnef: `regs.h` has `MT_MAC_STATUS_TX = BIT(1)` and `MT_MAC_STATUS_RX = BIT(0)`. This loop (and the "TX idle=%d" print two lines down) tests BIT(0), so `swreset` waits for RX idle and then pulses the UDMA TX reset with TX possibly still in flight — the precondition the vendor sequence states. A FAIL from this gate then cannot separate "sequence does not work" from "run too early". Use `mt_poll(&dev, MT_MAC_STATUS, MT_MAC_STATUS_TX, 0, ...)` as `init.c` does, or the `MT_USB_DMA_CFG_TX_BUSY` test `mt_mac_stop` uses.

src/mt7612u/fw.c
josephnef: Second point on the retry: `d->io_err` is not cleared between attempts, and `mt_init_hardware` refuses to report success while it is non-zero (init.c ~417). An attempt 1 that also tripped any EP0 accessor (`mt_single_wr` FCE addr/len, the `CPU_DESC_IDX` read-modify-write, `mt_poll`) before the bulk chunk timed out poisons a clean attempt 2: firmware loads, open still fails with "failed register transfers ... refusing to report success". Add `mt_io_clear(d)` at the top of each attempt.

josephnef (dismissed):
Review of #414 (8 findings, ranked most severe first). Each was independently verified against the PR head; line-anchored comments below.

The four I would fix before merge: the wedge recovery is reachable only via `mt7612u_open()` and not `mt7612u_open_handle()`; the RSSI EMA that drives the gain class is fed by every frame on the channel rather than mt76's associated-station average; `mt7612u_phy_tick()` can block the caller thread ~3 s, which for a sync-drain consumer is the undrained-EP4 state the PR itself refuses at `mt_mac_start`; and teardown still disables RX only after the ring is cancelled, the mirror image of the new start-side invariant.

src/mt7612u/usb.c
```diff
@@ -513,6 +534,60 @@ int mt_open(struct mt7612u_dev *d, const char **err)
 		libusb_release_interface(d->h, 0);
 		goto fail;
 	}
+
+	/* A wedge experiment must not have its recovery hidden inside open().
+	 * With this set, open() observes and reports but repairs nothing. */
+	if (getenv("MT7612U_NO_AUTORECOVER")) {
```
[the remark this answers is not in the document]
snokvist: Fixed. `MT7612U_NO_AUTORECOVER` is now `d->no_autorecover`, a field `bringup` sets before `mt_open()`; nothing in the library reads the environment for it, and `mt_recover_usb()` reads the field.

`MT7612U_DEV` stays deferred, as agreed in #412 rather than as a second accumulation. There is no public way to pass a selector — `mt7612u_open()` allocates the device itself and `struct mt7612u_dev` is opaque — so dropping the read here would leave a multi-adapter consumer unable to choose an adapter at all. The wrapper does not need it: it arrives through `mt7612u_open_handle()` having already selected the device.

docs/mt7612u.md
```diff
@@ -157,6 +157,29 @@ len=102  HT    mcs=15 nss=2 bw=20 sgi=1 ldpc=1 stbc=0 rssi=[-71,-63]
...
+
+mt76 runs `cal_work` every second (`MT_CALIBRATE_INTERVAL == HZ`,
+`mt76x2/usb_phy.c:42`): a self-guarded channel calibration, TSSI compensation
+(an MCU command each second) and `mt76x2_phy_update_channel_gain()`. This
+port ran none of it, and the cost was measured rather than assumed: against a
+peer 20 cm away airing 1400-byte HT-MCS7 frames at 3037 fps, a receiver with
+no tick took **3 frames in 10 s**; calling `mt7612u_phy_tick()` once a second
+it takes **~4850/s** (nine runs). Not the adapter — symmetric in both
+directions, and both hear 400–1000 fps of ambient on ch1/36/40/44/48 — not
+the ring (its own counter agreed at 3), and not the air: an RTL8812AU witness
```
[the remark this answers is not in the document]
snokvist: Fixed. `docs/mt7612u.md` no longer restates the one-second cadence, the calibration behaviour, the shared-MCU-endpoint rationale or the rate — the public header is authoritative for the contract, and the header comment now says in so many words that it is the single home for the figure. The doc keeps only what is not in the header: the stimulus definition, the four control arms, the bisection counts and the implementation background.

src/mt7612u/phy.c
```diff
@@ -431,6 +437,164 @@ int mt_chan_group(uint8_t chan, uint8_t bw, uint8_t *hw_chan,
...
+
+	/* mt76 averages RSSI over associated stations.  A monitor consumer has
+	 * none, so this is fed from the RX path; -75 is mt76's own fallback. */
+	if (!d->cal.avg_rssi_all)
+		d->cal.avg_rssi_all = -75;
+
+	low_gain = (d->cal.avg_rssi_all > rssi_gain_thresh(d->bw)) +
+	           (d->cal.avg_rssi_all > low_rssi_gain_thresh(d->bw));
+
+	gain_change = d->cal.low_gain < 0 || ((d->cal.low_gain & 2) ^ (low_gain & 2));
```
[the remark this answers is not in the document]
snokvist: Fixed. `mt_set_channel_ex()` resets the gain-class sentinel and the fine adjustment on every tune, and the `agc_gain_init` snapshot (AGC 8/9) now runs **above** the fast return, so a fast retune no longer establishes the next tick's gain base from the previous channel.

Those two reads are `mt_rr_chk()`'d rather than `mt_rr()`: hoisting them above the fast return also moved them outside the `mt_io_errors()` bracket that used to catch a bad transfer, and `FIELD_GET(MT_BBP_AGC_GAIN, ~0u)` is 0x7f — a gain base of 127 programmed silently.

src/mt7612u/mcu.c
```diff
@@ -55,12 +60,40 @@ int mt_mcu_send(struct mt7612u_dev *d, int cmd, const void *data, int len,
...
+	 * cascading through every 1 Hz tick.  Bounded: a queue deeper than 16
+	 * is a fault worth reporting, not one worth looping on. */
+	{
+		uint8_t stale[MCU_RESP_URB_SIZE];
+		int n = 0, got;
+
+		while (n < 16 &&
+		       !mt_bulk(d, MT_EP_IN_CMD_RESP, stale, sizeof stale, &got, 5) &&
+		       got >= 4)
+			n++;
```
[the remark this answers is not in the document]
snokvist: Head warns distinctly at the cap and sends the command anyway, so a deep transient is reported rather than swallowed.

Since this comment the drain also stopped being unconditional: it is armed by `d->mcu_stale_pending`, set only when a command actually goes unanswered (both `mcu_wait_resp()` give-up paths and a failed bulk OUT, which can still have been delivered). Hitting the cap deliberately leaves the flag armed so the next command drains again instead of reading a straggler as its own reply, and the drain only disarms on a clean timeout — exiting on a short packet says nothing about what is still queued behind it.

snokvist:
Both blocking items were right, and both were this PR's own contract not being met by the code. All eight inline points are fixed, plus the three process items. Head is `4a6a14c`.

## Blocking

**1. Recovery unreachable from `mt7612u_open_handle()`.** Factored exactly as you suggested: the whole block (`MAC_SYS_CTRL=0`, `mt_rx_flush`, the TX_CLR pulse loop, the `TX_BUSY` verdict) is now `mt_recover_usb()`, called after identify on both paths, the way `mt_dev_state_init()` already was.

Nothing in `bringup` had ever opened the second entry point, so I added a gate that does: **`bringup adopt`** opens libusb itself, claims interface 0, resets, and drives `mt_adopt()` + `mt_init_hardware()` — the two steps `mt7612u_open_handle()` is. Driving them directly rather than calling the public function is deliberate: that function allocates the device itself, so an observe-only wedge experiment could never be configured through it. That is also what makes the verification possible —

```
$ sudo MT7612U_NO_AUTORECOVER=1 MT7612U_DEV=2-1 bringup adopt
devourer [I] mt7612u: auto-recovery disabled: U3DMA_CFG=0x00c40020
ADOPT: PASS
```

— the "auto-recovery disabled" line is emitted from inside `mt_recover_usb()`, so its appearance on the adopt path is the code-path fact you asked for, observed rather than argued.

Honest limit: I could **not** reproduce the soft wedge this session (5 kill cycles with TX+RX up; `U3DMA_CFG` stayed `0x00c40020` each time). So "adopt recovers a *wedged* adapter" rests on it being one shared function with two call sites, not on a measurement. Same footing as the hard tier already has in the PR.

**2. Teardown runs the receiver undrained.** Fixed. New `mt_mac_rx_disable()` clears `ENABLE_RX` only, before the ring is cancelled — in `mt7612u_close()`, and at every ring-cancelling site in `bringup` through one `rx_teardown()` helper rather than the pair spelled out twenty times. That includes `gate_caps`, which cancels the same EP 4 ring under `mt_async_stop`. The three `mt_async_stop` sites in `gate_soak`/`gate_ampdu` are deliberately untouched — both run `mt_mac_start(..., MT_RX_DRAIN_NONE)` with `mt_async_start(&dev, NULL, NULL)`, so `ENABLE_RX` is never set. `gate_rx` also got it: it has no ring, but it stops draining EP 4 on its own thread while `mt_mac_stop()` leaves RX on for up to 150 ms.

## Inline

- **`agc_gain_init` stale on a fast retune.** The two AGC 8/9 reads moved above the fast return. One correction to the reasoning, not the fix: mt76 does *not* re-read on every `set_channel` — `mt76x2/usb_phy.c:170-174` returns on `scan` *before* `mt76x02_init_agc_gain()`, as does `pci_phy.c:233-237`. What differs is that `mt76x02_init_agc_gain()` is the only place mt76 sets `low_gain = -1`, so its scan path leaves `low_gain` alone and `gain_change` comes out false. This port resets it unconditionally in `mt_set_channel_ex()`, which forces `gain_change` true after *any* tune — so here the stale base really does get written.
- **`avg_rssi_all` fed by every frame.** Removed, and you were right that it was less faithful than doing nothing. `mt76_get_min_avg_rssi()` (util.c:72) averages associated stations and returns 0 for a monitor consumer, so `mt76x2/phy.c:283-285` substitutes `-75`; `phy_update_channel_gain()` now uses that directly. The RX hot path does no cross-thread write at all, and the `_Atomic` field and `<stdatomic.h>` are gone from `internal.h`. The `low_gain == 2` arms are unreachable as a result — kept, because they are the faithful port and come back with a real per-peer RSSI source, with a comment saying so since the compiler cannot see it.
- **The 5 ms drain before every command.** Now armed by `d->mcu_stale_pending`. Measured under exactly the load that motivated it: nine calibration timeouts during channel setup produced **one** drain of nine replies.
- **Unchecked `mt_rr()`.** `mt_rr_chk()` on both — the `U3DMA_CFG` reads in `mt_recover_usb()` and `MT_RX_STAT_1` in `phy_adjust_vga_gain()`.
- **`mt_fw_init` retry too broad.** A distinct return now separates a bad blob (fails immediately, naming the file) from the wire upload and from the chip not starting what was uploaded — those last two both retry, since the `mt_poll` timeouts are the same settling window you measured and `gate_swreset`'s whole verdict depends on it. Your second point is fixed too, though not quite as written: `mt_io_clear()` would have wiped the caller's whole bracket, because `mt_init_hardware()` opens it *before* `mt_power_cycle()` and there is only one accumulator — a flaky power-on plus a retried upload could then have reported success with the WLAN core partly powered. It samples and restores instead.
- **`gate_rx` ticking its own only drainer.** Documented rather than worked around, and the note is in the gate's own output, not just a comment. Gating RX around the tick would have changed what the gate measures — `mt_mac_start()` rewrites `MT_RX_FILTR_CFG` to the initvals, overwriting the monitor filter `gate_rx` sets deliberately — and a tool-side workaround would make the gate immune by construction to the hazard it should be able to witness. `gate_arx` remains the valid witness, with `arx <ch> <secs> 1` its negative control.
- **`gate_swreset` polling the wrong bit.** Now `mt_poll(&dev, MT_MAC_STATUS, MT_MAC_STATUS_TX, 0, …)`, plus a line saying so when TX never goes idle, so a FAIL is attributable.

## Process

- **getenv.** `MT7612U_NO_AUTORECOVER` is a field now, so the debt stops accumulating. I left `MT7612U_DEV` deferred rather than converting it: there is no public way to pass a selector — `mt7612u_open()` allocates the device and the struct is opaque — so removing the read would leave a multi-adapter consumer unable to choose an adapter at all, and the wrapper does not need it (it selects the device before `mt7612u_open_handle()`). Ending it properly wants a small public setter or an open-with-config, which felt like the wrong thing to add to a bug-fix PR. Say the word and it is a short follow-up.
- **Number drift.** One figure, one home — the `mt7612u_phy_tick()` comment, which now says so. `phy.c` (three sites), `bringup.c` and `docs/mt7612u.md` point at it. `4362/s` is deleted rather than re-homed: nothing in the tree says which build or window produced it, so relabelling it would have been a guess.
- **`docs/mt7612u-usb-wedge.md`.** Provenance header dropped; register tables and the untested-status verdict kept.
- Qodo #1, #2, #6 and #8 answered and resolved.

## Two more, found reviewing the merged result

Worth calling out because neither was visible in any single change:

- Hoisting the AGC reads above the fast return also moved them out of the only `mt_io_errors()` bracket guarding them. `FIELD_GET(MT_BBP_AGC_GAIN, ~0u)` is `0x7f`, and `low_gain` has just been reset — so one EP0 hiccup during a fast retune would have programmed a gain base of 127 on the next tick and gone deaf until the next full tune, silently. Checked now, keeping the previous snapshot.
- The `mt_io_clear()` above.

## Verification on this head

Peer is a second MT7612U flooding ch149 — a different stimulus from §1's, so quoted separately rather than folded into that figure:

| check | result |
|---|---|
| RX | **4902–4913 fps**, 5 runs, `rx_err`/`rx_invalid`/`rx_dropped` all 0, peer verified airing before and after each run |
| adopt | `bringup adopt` PASS; recovery observed on that path |
| drain | 9 timeouts → **one** drain of 9 replies |
| retune | 530 ms full / 50.6 ms fast |
| build | strict `-Wall -Wextra -Wshadow -Wmissing-prototypes` clean, `make check` green |

No MT7612U-side claim here is inferred: each row is from a run on `2-1` this round.

josephnef (approved):
Re-reviewed at `4a6a14c`. Both blocking items and all ten inline points are fixed as described, verified in the diff:

- `mt_recover_usb()` runs after identify on both `mt_open()` and `mt_adopt()`; `bringup adopt` exercises the second path.
- `mt_mac_rx_disable()` clears `ENABLE_RX` before `mt_async_stop()` in `mt7612u_close()`, and `rx_teardown()` does the same at every ring-cancelling site in bringup, `gate_rx` included.
- AGC 8/9 snapshot above the fast return, checked, keeping the previous base on a failed read — and your correction on mt76's scan path is right: the difference is this port's unconditional `low_gain = -1`, not a re-read upstream.
- RSSI EMA removed; `-75` pinned with the `low_gain == 2` arms documented as unreachable.
- Drain armed by `mcu_stale_pending` on all three unanswered paths, disarmed only on a clean timeout.
- `mt_rr_chk` on the `U3DMA_CFG` and `MT_RX_STAT_1` decisions.
- `FW_ERR_FATAL/UPLOAD/DEVICE` split with the accumulator sampled and restored rather than zeroed — the restore is the right call for the reason you give.
- `swreset` polls `MT_MAC_STATUS_TX`.
- `no_autorecover` is a field; `MT7612U_DEV` deferral is fine with the reasoning stated in-thread — a small open-with-selector follow-up when the wrapper lands.
- One figure, one home; wedge doc provenance header gone; the `mt7612u.md` intro splice is fixed too.

Local at this head: `make -C src/mt7612u check` passes (frame_shape, field_macros, api_link), and the subtree builds clean with `-Wall -Wextra -Wshadow -Wmissing-prototypes -Werror`. No MT7612U here, so your 4902–4913 fps / adopt / drain rows are yours alone; the code-path claims are what I verified.

CI is green on `4a6a14c`; the earlier failed `qodo-gate` run on the same SHA was superseded by the passing one. Squash-merge when you are ready.

Issue #419, opened 2026-09-09
MT7612U integration: wire src/mt7612u in behind IRadio


josephnef:
Tracks the integration of the MediaTek MT7612U subtree (`src/mt7612u/`, landed unwired in #412 and hardened in #414) behind the vendor-neutral radio contract from #415. Until this lands the subtree builds only through its own Makefile and no devourer binary can open an MT7612U.

## Scope

- [ ] **CMake**: `DEVOURER_MT7612U` option (default OFF), following the `DEVOURER_HAVE_*` pattern; the subtree's pthread/`nanosleep`/`clock_gettime` swap to `<thread>`/`<chrono>` so MSVC and mingw build it (Windows is first-class here).
- [ ] **Factory**: a vendor-neutral VID:PID gate in `WiFiDriver::CreateRadio` **ahead of** the Realtek `SYS_CFG2` read, the way Kestrel gates PID-first — on MediaTek silicon that read is the wrong protocol, not merely ambiguous. An unrecognised non-Realtek VID must refuse rather than fall through to Jaguar1. `ChipGeneration` entry for MediaTek.
- [ ] **Transport**: MediaTek gets its own transport behind the same factory, not a shim over `ITransport` (32-bit registers plus in-band MCU over EP8/EP5 do not fit the Realtek shape). The C library is the transport; adopt the handle `WiFiDriver` already opened, reset and claimed via `mt7612u_open_handle()`.
- [ ] **`IRadio` implementation** over the C library: `Init`/`InitWrite`/`StartRxLoop`/`SetMonitorChannel`/`send_packet`(+`send_packets`)/`GetSelectedChannel`/`SetCcaMode` (implement or refuse loudly), caps, ACK responder, TSF, link quality. Unsupported optional members keep the not-ported defaults; unavailable measurements report invalid, never fabricated IGI/NHM values.
- [ ] **The 1 Hz PHY tick lives inside the backend**, on a backend-owned worker thread (the Jaguar3 phydm-thread shape), not as an `IRadio` method. Stop/join it before teardown; serialize whole channel-change / power operations against it (the recursive `io_lock` protects nested transactions, not a full channel change); never run calibration from an RX completion callback.
- [ ] **RX-loop stop semantics**: `StopRxLoop()` must **not** map onto `mt7612u_rx_stop()` — that cancels both rings without disabling MAC RX first. Preserve disable-RX-before-removing-the-drain and shared ring ownership so RX is restartable.
[...]

Issue #12, opened 2025-04-08
Set _usbTxAggMode to false to solve TX package loss issue


imagemlt:
Hello,when I tried to implement TX (https://github.com/imagemlt/DualDevourer) I didn't notice that we have already set _usbTxAggMode to True. https://github.com/OpenIPC/devourer/blob/455db503e8ce8bfbd5f5d35ef0f90e7dbb367851/src/HalModule.h#L44 . My implementation is based on PC driver which do not have TX aggregation by default, so this may cause package loss issue. My rtl8812au card is broken these days and I cannot test it these days, I think perhaps you can try turn off this and test if package loss will decrease greatly. Thanks.

imagemlt:
I checked https://github.com/svpcom/rtl8812au 
https://github.com/svpcom/rtl8812au/blob/4d45664397fb086f77ecc1604fc605c851d3cb75/hal/rtl8812a/usb/usb_halinit.c#L123
Since only aggregate one package for rtl8812au when turn on this,I think we can just turn it off.

vertexodessa:
Thank you @imagemlt , I tested it locally - works just fine

imagemlt:
@vertexodessa BTW I'm currently working on running other wifi card in [Termux](https://termux.dev/). My idea is use qemu in termux only to initialize wifi card,and then send SIGKILL/SIGSTOP to qemu process and use libusb to takeover wifi card outside of qemu. I have tested 8822cu rx part in my code and it works very well, except longer initialize time than devourer. there is no loss of transmission performance  since real transmission is running outside of QEMU. you can try it, now only support 8822cu card, https://github.com/imagemlt/wfb-termux-qemu. We can just focus on TX/RX part with these method, which is quite simple for us to migrate linux kernel driver.

vertexodessa:
hi @imagemlt , that's a very interesting idea! 
but if i understand correctly to initialize the card, you first need a kernel driver right?

imagemlt:
@vertexodessa https://github.com/imagemlt/wfb-termux-qemu/tree/main/qemu you can see here, in rootfs.img I compiled 8822cu driver for the virtual machine inside qemu. 

<img width="842" alt="Image" src="https://github.com/user-attachments/assets/c92bccad-cccc-4a12-9a66-b97ed2c10d5e" />

[...]

Issue #11, opened 2025-03-30
Compile ERROR: LIBUSB_ENDPOINT_TRANSFER_TYPE_BULK


gghyoo:
error: 'LIBUSB_ENDPOINT_TRANSFER_TYPE_BULK' was not declared in this scope; did you mean 'LIBUSB_TRANSFER_TYPE_BULK'?
  282 |            LIBUSB_ENDPOINT_TRANSFER_TYPE_BULK)) {
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |            LIBUSB_TRANSFER_TYPE_BULK

floppyhammer:
Your libusb version is too old.

PR #424, opened 2026-09-10
mt7612u: access-point support — hardware beacon behind IRadio, open and WPA2-PSK verified on air

snokvist:
Follows #419's integration (#420-#423). That gave devourer an MT7612U it can
open, receive and transmit on. This gives it one it can be an **access point**
on — open and WPA2-PSK — with no MediaTek-specific code in the AP layer.

## It works, and devourer's own harnesses are what prove it

`tests/ap_responder.cpp` and `tests/ap_wpa2.cpp` against an MT7612U, with no
change to their AP logic. Both already took an `IRadio*` — #415 made them
genuinely backend-agnostic — so nothing in them knows this is MediaTek. (They
each gained a dozen lines that silence the beacon before `_exit`; see below.
Nothing in that is MediaTek-specific either.)

```
AP 5-1   station 2-1 (wlx40a5ef5027a1)   ch36 (5180 MHz)
== open network (ap_responder) ==
  PASS  open: beacon armed
  PASS  open: beacon on air
  PASS  open: station associated
  PASS  open: data plane (rtt min/avg/max/mdev = 0.567/0.812/1.013/0.150)
  PASS  open: hardware auto-ACK (auth at retry=0)
  PASS  open: nothing left airing after exit
== WPA2-PSK (ap_wpa2) ==
  PASS  wpa2: beacon armed
  PASS  wpa2: 4-way complete (MIC verified, station keyed)
  PASS  wpa2: encrypted data plane (rtt min/avg/max/mdev = 0.795/1.177/1.525/0.240)
  PASS  wpa2: nothing left airing after exit
== beacon lifecycle (StartBeacon / StopBeacon / re-arm) ==
  PASS  stop: armed - beacon on air
  PASS  stop: stopped - beacon gone
  PASS  stop: re-armed - beacon back
  PASS  stop: local contract checks

=== 14 passed, 0 failed ===
```

`retry=0` on the auth is the load-bearing one: an un-ACKed frame is
retransmitted with FC Retry set, so retry=0 *is* the hardware ACK, and it is
the only evidence that the port identity and the APC slot are both right.

That output is `tests/mt7612u_ap_onair.sh`, added here, so the numbers are a
command rather than a story. 14/14 on ch36 **and** 14/14 on ch6 — both bands
carry association, the data plane, WPA2 and the auto-ACK, not just a beacon.

## What was actually missing

Nothing in the AP layer, and no new driver primitives. `StartBeacon`,
`UpdateBeaconPayload` and `StopBeacon` fell through to `IRadio`'s base-class
`return false`, so a MediaTek adapter beaconed nothing. Everything underneath
was already device-verified (`mt_beacon_init` / `_write` / `_set_enable`,
`mt_ap_set_bssid`), and `SetAckResponder`, `ReadTsf` and `WriteTsf` were
already implemented.

Three public entry points, because the backend reaches the subtree through the
public header and `api_link` holds that line (31 now):

```
mt7612u_beacon_start   radiotap-framed or bare MPDU; moves the MAC identity
                       (MT_MAC_ADDR *and* the MT_MAC_BSSID base) to addr2,
                       publishes addr3 in the APC slot, arms TBTT
mt7612u_beacon_update  reload in place, engine untouched
mt7612u_beacon_stop    disarm, and retract the identity it took
```

**The identity is the whole trick.** `IRadio` says addr2/addr3 *set* the port
MAC/BSSID, `RtlJaguarDevice` implements exactly that, and `ap_responder`'s own
comment says "MACID = BSSID, set by StartBeacon". My first draft instead
*refused* a BSSID that was not the adapter's own MAC — which would have made
both shipped harnesses unusable on this part, since both hardcode
`02:42:75:05:d6:00`. It now follows `mt76x02_mac_setaddr` — move `MT_MAC_ADDR`
and the `MT_MAC_BSSID` base together, then derive the APC slot the way
`mt76x02_util.c:310` does. Getting *that* wrong is the review finding below,
and it is the reason the two registers move together rather than one of them.

## Also fixed, not MediaTek-specific

`ap_responder`, `ap_wpa2` and `ul_trigger_ap` all arm a hardware beacon and
then `_exit(0)`, skipping every destructor — so `StopBeacon` never ran and the
beacon kept airing after the process was gone, until the adapter was
power-cycled. Measured: a scan after exit found the SSID live at `last seen:
308 ms ago`. The beacon is hardware-autonomous on the Realtek parts too, and
`IRadio.h` has warned about this since it was written; `beacon_update_probe.cpp`
already called StopBeacon before its own `_exit`, these three did not.

Twelve lines each, not one: `StopBeacon` became able to *fail* during this
review (it could not before — see below), and #411 amended `IRadio.h` the same
day to say such a failure "must be retried (or followed by hardware shutdown)
before its shared port is reused". So each harness retries up to three times
and warns if the beacon is still armed, rather than calling it once and
discarding the result.

## What review changed

Five rounds, two of which returned DO-NOT-MERGE and were right to. The
headline finding is still the one worth reading:

**The APC slot index rested on a premise that is false in this codebase.** I
derived it as `(ta[0] & 2) ? 1 : 0`, arguing mt76's
`1 + (((macaddr[0] ^ addr[0]) >> 2) & 7)` collapses because the identity had
just been retargeted. It does not. `dev->macaddr` is written once, from the
EEPROM, and `MT_MAC_BSSID` — the register carrying `MBSS_MODE`, and the base
the hardware derives the index from — is written once at init from the factory
MAC; the ACK-responder retarget moves only `MT_MAC_ADDR`. So the hardware was
indexing off a different address than the host assumed, and the constant was
correct only where `1 + (((factory[0] ^ ta[0]) >> 2) & 7)` happens to be 1.

This bench is exactly such a case (`40:a5:ef:…`, so the XOR term is 0), which
means **every on-air run agreed with the wrong reasoning**. On an adapter whose
first byte is `0xe8` the AP would beacon perfectly and acknowledge nobody. The
bring-up gate had refused that case; the ABI flattened it to a constant. Fixed
by making the retarget a real `setaddr`, so the premise is true by construction.

Second: **`mt7612u_beacon_stop` could not fail** — `mt_clear` reports only its
read half and the `mt_ap_set_bssid` returns were dropped — which made
`StopBeacon`'s failure branch unreachable and the new harness's assertion for
it vacuous, while the real hazard (an EP0 stall leaving the MAC beaconing)
reported success. Also fixed: the reserved-page copy could fail mid-frame and
report success (a *torn* beacon airs, worse than none); `UpdateBeaconPayload`
did not raise the suppression guard mt76 explicitly brackets for that reason;
`Stop()` discarded the result while promising a retry that did not exist; and
the public header still described the pre-fix behaviour.

The earlier round found eight more, one reproduced with a probe against the
real object file: `beacon_split` left four of nine `mt7612u_tx_rate` fields
indeterminate, and `sgi`/`ldpc`/`stbc` go straight into the rate word the MAC
transmits verbatim while `power_adj` short-circuits the derived TX power.

The last two rounds were about **fixes that moved the bug rather than removing
it**, which is worth calling out because it happened three times:

- The identity retract was added, and then the two failure paths *inside the
  block that moves the identity* were the ones not routed through it.
- Ownership of `MT_MAC_ADDR` was claimed before the call that transfers
  ownership away, so the claim was immediately clobbered and the **success**
  path stopped restoring.
- The retry that the retract relies on could not reach `MT_MAC_ADDR` at all:
  `mt7612u_clear_ack_responder` cleared its own "restore owed" flag
  unconditionally and early-returns on it, so attempts two and three were
  no-ops against a still-leaked port MAC. Both halves of the retry had to agree
  before either worked.
- And fixing *that* widened `beacon_start`'s `io_err` bracket over a
  `mt_set(MT_AUTO_RSP_EN)` the code itself documents as a no-op (the gate is
  already on from `mac_reset`). `mt_set` is `mt_rmw`, which bumps `io_err` and
  skips its write when its *read* half stalls, and its return is ignored
  because the readback below is the real check — so one transient EP0 read
  stall tore down a verified-good arm and returned `-2` for a beacon that was
  on the air.

The last of those was found by an adversarial pass over the diff, not by the
harness: every one of these failures leaves the beacon airing perfectly, so no
amount of watching the air finds them.

Relatedly, `-1` no longer escapes from inside the identity block. It is
contracted as "nothing was touched, whatever was airing still is", and the two
exits returning it were reachable only *after* `MT_MAC_ADDR` had moved — and
the unwind restores the **factory** address, the only one saved anywhere. A
failed re-arm over a live beacon therefore retracted an identity that beacon
never had while its page and timers kept airing it, under a return value
promising nothing changed. There is no atomic re-arm to offer (one
`MT_MAC_ADDR`, one save slot, previous occupant not in it), so those exits take
the AP down deliberately and return `-2`. `-1` now belongs to the input
refusals above the first hardware write.

## Rebase notes

The AP work predates the C++ migration (#421) and the backend (#422). Rebasing
surfaced that `beacon.c` was compiled by **nothing** — the subtree Makefile
globs `*.cpp` and the CMake list is explicit — plus C11 `_Atomic`, implicit
`void*` conversions, and a `memset` over a no-longer-trivially-copyable type.
The original four AP commits and that repair are squashed into one, because
the first three do not build on today's master and a bisect through them
returns build failures rather than the commit you are looking for.

## What this does not show

In `docs/mt7612u-ap-mode.md` at the same length as the numbers:

- **Hardware CCMP is untested.** The 4-way above is devourer's *software*
  CCMP, the same code the Realtek backends use. `MT_WCID_KEY` is not even
  defined in this tree and the key path is unreached, so "crypto becomes
  hardware on MediaTek" remains a claim. It is more than a flag: `MT_WCID_KEY`
  and `MT_WCID_IV` are both absent, `mt_tx_build()` sets `MT_TXD_INFO_WIV`
  unconditionally (meaning "not encrypted") where mt76 gates it on the WCID
  having a key, and RX would have to strip the IV/PN itself. Opened as its own
  issue rather than guessed at here, and deliberately **not** proposing an
  `IRadio` key surface — per your guidance, that contract should be designed
  against two backends rather than one.
- **The encryption was not independently captured.** No third radio sniffed the
  Protected bit; "encrypted" rests on the verified MIC plus traffic reaching a
  CCMP-only station.
- **The station is the same silicon** (MT7612U on `mt76x2u`), so it is not an
  independent-generation witness. The RTL8812AU witness in the earlier Stage
  A/B section is.
- **One AP, one station, ~20 cm apart; longest run 70 s.** No soak, no second
  station, no rekey, no roaming, no channel change while beaconing.

## Verification

- 63/63 `ctest` with `DEVOURER_MT7612U=ON`, and with it OFF; `make -C
  src/mt7612u check` green (4 binaries, `api_link` at 31 entry points); no
  warnings in either build.
- **Every commit builds**, both build systems — checked per commit, because the
  original series did not (the first three predate the C++ migration and their
  repair came fourth; they are squashed into one).
- On-air, after the review fixes: `tests/mt7612u_ap_onair.sh` 14/14 on ch36 and
  14/14 on ch6.
- `tests/regress.py` 4/4 on the injector path — devourer↔devourer,
  devourer↔kernel both directions — so the AP changes did not disturb the
  transmit/receive behaviour #419 shipped.
- The addr2/addr3 update guard has a positive control as well as two negative
  ones, run against a live beacon: an unchanged payload must be *accepted*, so
  a guard that refused everything would not pass. Only the addr3 arm
  discriminates against the pre-fix code — addr2 was already guarded — and the
  test says which is which rather than counting both as new coverage.

---


src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+	if (on) {
+		/* INTVAL is 16 bits of 1/16-TU, so the interval caps at 4095 TU. */
+		if (interval_tu == 0 || interval_tu > 0xffffu / 16) {
+			ERR("beacon interval %u TU out of range (1..4095)", interval_tu);
+			return -1;
+		}
+		mt_rmw(d, MT_BEACON_TIME_CFG, MT_BEACON_TIME_CFG_INTVAL,
+		       FIELD_PREP(MT_BEACON_TIME_CFG_INTVAL, interval_tu << 4));
+		mt_set(d, MT_BEACON_TIME_CFG, bits);
```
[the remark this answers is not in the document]
snokvist: Fixed. Correct, and it is the counterpart to a fix I made last round and did not carry across: `mt7612u_beacon_stop` got an `io_err` bracket and `mt7612u_beacon_start` did not. `mt_beacon_set_enable`'s arming path is `mt_rmw`/`mt_set`, which report their READ halves, so a failed EP0 write returned 0 and `StartBeacon` set `_beacon_active = true` for an engine that was never armed.

`mt7612u_beacon_start` now snapshots `mt_io_errors()` before it touches the hardware and fails if the count moved, alongside checking the returns it was already ignoring.

src/mt7612u/Mt7612uRadio.cpp
```diff
@@ -841,6 +860,92 @@ void Mt7612uRadio::ClearAckResponder() {
...
+  /* Cleared BEFORE the call, not just set after it. mt7612u_beacon_start()
+   * runs mt_beacon_init(), which disarms the timer and suppresses every slot,
+   * so a failure after that point leaves the beacon dead - and a re-arm that
+   * fails (an over-long beacon body, say) would otherwise keep reporting the
+   * PREVIOUS arm as live, which is how UpdateBeaconPayload comes to return
+   * true for every write into a disarmed engine. */
+  _beacon_active = false;
+  if (mt7612u_beacon_start(_dev, beacon, len,
+                           static_cast<unsigned>(interval_tu)) != 0)
+    return false;
```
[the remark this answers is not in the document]
snokvist: Fixed, and this one is the sharpest finding in the set - it is a defect I introduced last round while fixing a different one.

I cleared `_beacon_active` before the call to make a failed re-arm failure-atomic. But `mt7612u_beacon_start` has failure paths BEFORE it touches the engine (input refused, identity retarget failed), and on those the previous beacon is still airing while the flag now says otherwise. `UpdateBeaconPayload`, `StopBeacon` and the destructor all then treat a live autonomous beacon as inactive, so nothing ever silences it - the exact outcome the whole change exists to prevent.

It needed three outcomes rather than two, because only the library knows which side of `mt_beacon_init` a failure landed on:

  0   armed
  -1  refused before the hardware was touched - the flag is NOT cleared,
      whatever was airing still is
  -2  failed after the engine was disarmed; the library unwound the identity
      too, so nothing is airing and the flag is cleared

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,418 @@
...
+	 * ta[0]) >> 2 & 7), and this constant is right only for the adapters
+	 * where that happens to be 1. Getting it wrong is silent: the AP beacons
+	 * perfectly and acknowledges nobody.
+	 */
+	idx = (ta[0] & 0x02) ? 1 : 0;
+
+	if (mt_ap_set_bssid(dev, idx, bssid)) return -1;
+	mt_beacon_init(dev);
+	if (mt_beacon_write(dev, mpdu, mpdu_len, &rate)) return -1;
+	return mt_beacon_set_enable(dev, 1, interval_tu);
```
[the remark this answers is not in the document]
snokvist: Fixed. `mt7612u_beacon_start` retargeted `MT_MAC_ADDR` and the `MT_MAC_BSSID` base up front and then had four `return -1`s after it, none of which put the identity back - so a failed start left the adapter answering for a BSS that does not exist, with no beacon.

There is now a single `unwind_identity()` used by every failure path in `beacon_start` and by `beacon_stop`, and it restores only if THIS call was what moved the identity (see the ownership thread).

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,418 @@
...
+		const int was_taken = dev->ack_saved;
+		if (mt7612u_set_ack_responder(dev, ta))
+			return -1;
+		if (mt_mac_set_bss_base(dev, ta))
+			return -1;
+		/* Only claim ownership if nobody else already held the identity: a
+		 * caller who armed a responder first owns the saved factory address,
+		 * and restoring it on beacon stop would silently disarm them. */
+		if (!was_taken)
+			dev->beacon_took_identity = 1;
```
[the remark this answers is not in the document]
snokvist: Fixed, together with #11 - the two are the same defect seen from either end.

`MT_MAC_ADDR` is one register with two users and there was no arbitration: ownership was recorded once, at `StartBeacon`, from whatever `ack_saved` happened to be. So `SetAckResponder` after `StartBeacon` was silently disarmed by the matching `StopBeacon` (this thread), and the mirror case left the responder programmed for the beacon's addr2 instead of the caller's address (#11).

The rule is now "whoever wrote last owns it": `mt7612u_set_ack_responder` clears `beacon_took_identity`, so a responder armed after the beacon is not restored out from under the caller, and a stop after that call leaves the address alone. It is one line in `caps.cpp` with the reasoning next to it, and it is stated in the public header rather than left as behaviour.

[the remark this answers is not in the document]
snokvist: Fixed - same change as #6, from the other side. See that thread: ownership of `MT_MAC_ADDR` now transfers to whoever wrote it last, so `mt7612u_set_ack_responder` takes it back from the beacon and a later stop does not restore over it.

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,418 @@
...
+	if (rlen < 0) {
+		ERR("beacon: malformed radiotap header");
+		return -1;
+	}
+	if (rlen > 0 && (size_t)rlen < len) {
+		*mpdu = p + rlen;
+		*mpdu_len = len - (size_t)rlen;
+	} else {
+		/* rlen == 0 (no radiotap) or rlen == len (a header with no frame
+		 * after it): treat the buffer as a bare MPDU. OFDM 6 Mbps is the
```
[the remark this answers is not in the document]
snokvist: Fixed, along with #12 - they are two symptoms of the same weak classification.

`beacon_split` decided "radiotap or bare MPDU" from the parser's return value, which cannot carry that distinction: `mt_radiotap_parse` returns 0 both for "this is not a radiotap header" and for "it is one and it is malformed", and `rlen == len` (a header with no frame after it) fell to the bare branch too. Either way, radiotap bytes end up parsed as an 802.11 header - supplying the port addresses and the reserved-page contents, which then air.

Byte 0 decides now, and it is unambiguous: an 802.11 beacon's frame control is `0x80` and a radiotap header's version must be `0`. A buffer starting `0x00` is held to the radiotap rules (parse must succeed, and must leave a frame after it); anything else is a bare MPDU. No fallback between the two.

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,418 @@
...
+		 * basic rate every station must decode, which is what a beacon wants. */
+		r->phy = MT7612U_PHY_OFDM;
+		r->mcs = 0;
+		r->nss = 1;
+		r->bw = MT7612U_BW_20;
+		*mpdu = p;
+		*mpdu_len = len;
+	}
```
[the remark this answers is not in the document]
snokvist: Fixed - same change as #7, see that thread. A malformed radiotap header is no longer reinterpreted as an 802.11 frame; the shape is decided by byte 0 and each is held to its own rules.

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,418 @@
...
+	 * 64-byte EP0 transactions, so a TBTT landing mid-copy would air a TORN
+	 * beacon - leading bytes new, trailing bytes old. mt76 brackets the same
+	 * write for the same reason ("Prevent corrupt transmissions during
+	 * update", mt76x02_usb_core.c). mt_beacon_write() lowers the guard again
+	 * on its way out, which is why this is the only half needed here.
+	 *
+	 * Still no mt_beacon_init() and no set_enable(): the engine is armed, and
+	 * re-initialising it would clear the timer bits mid-flight. */
+	mt_wr(dev, MT_BCN_BYPASS_MASK, 0xffff);
+	return mt_beacon_write(dev, mpdu, mpdu_len, &rate);
```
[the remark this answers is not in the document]
snokvist: Fixed. The guard write was the one part of the update path that could fail silently, and it is the part the whole bracket exists for: if the suppression never lands, the multi-transaction page copy runs against a live slot and a TBTT mid-copy airs a torn beacon.

It is `mt_wr_chk` now and the update refuses when it fails, rather than proceeding to a copy it knows is unguarded. `mt_beacon_write`'s own `io_err` snapshot still covers the copy itself.

tests/ap_responder.cpp
```diff
@@ -290,5 +290,13 @@ int main(int argc, char** argv) {
...
           (unsigned long long)g_sent.load());
+  /* The chip beacons AUTONOMOUSLY once StartBeacon arms it, and _exit(0)
+   * below skips every destructor - so without this the beacon keeps airing
+   * after the process is gone, until the adapter is power-cycled, and it
+   * contaminates whatever runs next on that channel. Measured on MT7612U:
+   * a scan after exit still found the SSID live at 308 ms. Realtek has the
+   * same exposure - IRadio.h calls it bench-bitten. beacon_update_probe.cpp
+   * already did this; these did not. */
+  if (g_dev) g_dev->StopBeacon();
   _exit(0);
```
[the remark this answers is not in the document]
snokvist: Fixed. And the timing is pointed: `StopBeacon` only became capable of failing in the previous commit, and #411 amended `IRadio.h` the same day to say such a failure "must be retried (or followed by hardware shutdown) before its shared port is reused". These harnesses called it once, discarded the result, and `_exit(0)`ed - so there is no destructor coming to try again, and a transient EP0 stall leaves an autonomous beacon airing until the adapter is power-cycled.

All four (`ap_responder`, `ap_wpa2`, `ul_trigger_ap`, and the lifecycle harness's own final stop) now retry three times and print a warning naming what is still on the air. The lifecycle harness counts it as a failure, since teardown is the thing it exists to test.

tests/mt7612u_ap_onair.sh
```diff
@@ -0,0 +1,273 @@
...
+  pkill -x wpa_supplicant 2>/dev/null
+  [ -n "${STA_IF:-}" ] && { ip addr flush dev "$STA_IF" 2>/dev/null
+                            iw dev "$STA_IF" disconnect 2>/dev/null; }
+  # The MAC beacons autonomously. If a cell died before its teardown, only a
+  # port power-cycle is certain to silence it - and leaving one airing poisons
+  # the next run of this very script.
+  echo 0 > "/sys/bus/usb/devices/$AP_SYSFS/authorized" 2>/dev/null
+  sleep 2
+  echo 1 > "/sys/bus/usb/devices/$AP_SYSFS/authorized" 2>/dev/null
+  sleep 3
```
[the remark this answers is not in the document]
snokvist: Fixed - a fair hit on a script that runs as root and takes a path from the environment.

`cleanup()` now reads `idVendor`/`idProduct` and only power-cycles when they are `0e8d:7612`. A stale or mistyped `AP_SYSFS` writes nothing instead of deauthorizing whatever else happens to be on that port.

The power-cycle itself stays, because it is the only thing certain to silence a beacon left by a cell that died before its teardown - and a stale beacon poisons the next cell of the same run.

tests/mt7612u_ap_onair.sh
```diff
@@ -0,0 +1,273 @@
...
+mkdir -p "$OUT"
+pass=0; fail=0
+say()  { printf '%s\n' "$*"; }
+ok()   { pass=$((pass+1)); printf '  PASS  %s\n' "$*"; }
+bad()  { fail=$((fail+1)); printf '  FAIL  %s\n' "$*"; }
+
+cleanup() {
+  pkill -x apr_onair apw_onair bstop_onair 2>/dev/null
+  pkill -x wpa_supplicant 2>/dev/null
+  [ -n "${STA_IF:-}" ] && { ip addr flush dev "$STA_IF" 2>/dev/null
```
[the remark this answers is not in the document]
snokvist: Fixed. `pkill -x wpa_supplicant` would drop every wireless client on the host, including ones nothing to do with this test, and the name kills would reach a concurrent run of this same script.

It now tracks the PIDs it started - the AP processes in a `KIDS` list, and `wpa_supplicant` via `-P` into a pidfile - and cleanup kills exactly those.

tests/mt7612u_ap_onair.sh
```diff
@@ -0,0 +1,273 @@
...
+# a live beacon into a pass for "gone", and it stops a missed scan reporting a
+# live beacon as absent. Observed: a "beacon not scannable" FAIL in a run where
+# the station then associated, pinged, and got an auth at retry=0.
+seen() {
+  local i n best=0
+  for i in 1 2 3; do
+    n=$(iw dev "$STA_IF" scan flush freq "$FREQ" 2>/dev/null | grep -c "SSID: $1")
+    n=${n:-0}
+    [ "$n" -gt "$best" ] && best=$n
+    [ "$best" -gt 0 ] && break
```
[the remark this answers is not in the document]
snokvist: Fixed. `seen()` matched on SSID alone, so a neighbour running "devourerAP" could satisfy an arm check, break a stop check, or upset the exact-count comparison - and this bench sits in range of several APs.

It matches BSSID and SSID together now, via awk so the pairing is preserved (two `grep -c` patterns would count independently and pass on a coincidence of both appearing anywhere in the scan). Both harnesses use fixed, distinctive BSSIDs, so the check is exact.

josephnef (changes requested):
Reviewed at `424170b`. Built it (`-DDEVOURER_MT7612U=ON`), 63/63 `ctest`, `make -C src/mt7612u check` green including the new `beacon TXWI` cell, no new warnings — the verification claims hold. I also checked the register work line by line against `reference/mt76 @ be5ce79`.

## Verified against the vendor source

- The `MT_MAC_BSSID_DW1` correction is right, bit for bit: `mt76x02_regs.h:281-286` gives `ADDR` 15:0, `MBSS_MODE` 17:16, `MBEACON_N` 20:18, `MBSS_LOCAL_BIT` 21, `_B2` 22, `_B3` 23. `MT_BCN_OFFSET_BASE` 0x041c, `MT_BCN_BYPASS_MASK` 0x108c, `SYNC_MODE` 18:17 all match too, and `init.cpp:212-216` now reproduces `mt76x02_mac_setaddr` (`mt76x02_mac.c:755-760`: MBSS_MODE=3, LOCAL_BIT, then MBEACON_N=7) exactly.
- `mt_beacon_init()` matches `mt76x02_init_beacon_config` (`mt76x02_beacon.c:205-213`) step for step, and the `0xff00 | ~(0xff00 >> 1)` unsuppress matches `mt76x02_usb_core.c:223` with `beacon_data_count == 1`. The slot→bit mapping running *down* from bit 7 is easy to read backwards from the regs.h comment alone; the beacon.cpp comment does say it.
- The APC-slot reasoning now holds by construction. `mt76x02_add_interface` (`mt76x02_util.c:309-310`) derives `1 + (((mphy.macaddr[0] ^ addr[0]) >> 2) & 7)` *after* `mt76x02_mac_setaddr`, and `mt_mac_set_bss_base()` is what makes the same invariant true here. The premise correction in the last commit is the right fix.

## Blocking

**1. `src/mt7612u/beacon.cpp:347-350` — the identity retarget still leaks on two failure paths.**

```c
const int was_taken = dev->ack_saved;
if (mt7612u_set_ack_responder(dev, ta))
        return -1;
if (mt_mac_set_bss_base(dev, ta))
        return -1;
```

Both of those bare `return -1`s skip `unwind_identity()`. `mt7612u_set_ack_responder` (`caps.cpp:96-110`) writes `MT_MAC_ADDR_DW0/DW1` and *then* readback-verifies, so its failure path returns with the port identity already moved — and `mt_mac_set_bss_base` can only fail after `MT_MAC_ADDR` has moved successfully. Either way the adapter is left answering for a BSS that was never armed, which is the exact class the `f3f08a9`/`424170b` rounds say they closed ("a failed start left it retargeted with no beacon… every failure path unwinds now").

Worse in combination with the new return code: `-1` is contracted as "refused before the hardware was touched — whatever was airing still is", so `Mt7612uRadio::StartBeacon` deliberately leaves `_beacon_active` alone. A failed re-arm here therefore reports the *previous* beacon as live while `MT_MAC_ADDR` now points at the new `ta`. Both exits need `goto fail_pre` with `took` set from `!was_taken`.

## Worth fixing before merge

**2. `beacon.cpp:433-441` — a failed `mt7612u_beacon_update` leaves every slot suppressed.**

`update` raises `BCN_BYPASS_MASK = 0xffff` and nothing lowers it except a *successful* `mt_beacon_write`. But `mt_beacon_write` can refuse on caller-controlled length at `:78-83` — which is checked *after* the guard is up — and can also fail on `mt_tx_build` or a torn reserved-page copy. Any of those returns with the AP off the air, `_beacon_active` still true, and no path that re-lowers the guard short of a later successful update. A rejected payload should not silently take the AP down; check the length before raising the guard and unwind the guard on every failure exit.

**3. `beacon.cpp:426-441` — `update` accepts any addr2/addr3.**

`IRadio.h:394-396` says changing addr2/addr3 mid-flight is unsupported and that the port registers keep the StartBeacon identity. `beacon_split` happily parses a beacon carrying a different BSSID and loads it into the page, so the airing BSSID stops matching the programmed APC slot and `MT_MAC_ADDR` — beacons perfectly, ACKs nobody, which is precisely the silent failure the start path now goes to real lengths to prevent. Refusing costs a `memcmp` against what start took.

**4. `tests/mt7612u_ap_onair.sh` `cleanup()` — the "power-cycle" is an `authorized` toggle.**

The comment says "only a port power-cycle is certain to silence it" and then writes `0`/`1` to `…/authorized`. This tree's own `CLAUDE.md` (Hardware testing) records that as a trap: an `authorized` toggle is not a cold cycle, it leaves chip state, and VBUS never drops — so an autonomously beaconing MAC keeps beaconing straight through it, which is the one thing this cleanup exists to prevent between cells. Use `REGRESS_VBUS_MAP`/uhubctl like the rest of the tree, or drop the claim and say what it actually guarantees.

## Nits

- `tests/mt7612u_ap_onair.sh`: `wait_gone()` is defined and never called.
- `regs.h`: `MT_RX_FILTR_CFG_MCAST` / `_BCAST` are added and unused (the doc's own item-4 conclusion — clear `DUP` for an AP so retry evidence stays visible — isn't implemented either).
- `Mt7612uRadio.cpp` `AdjustBeaconTiming` / `…Fine` / `PinBeaconTbtt`: the log is an improvement, but `IRadio.h:452,507` already document `0` as the legitimate "no active beacon" answer, so a programmatic caller is exactly as blind as before. Fine to keep; the comment overclaims what the override changes.
- `docs/mt7612u-ap-mode.md:142-227`: "The gap — the driver primitives to add" is now a future-tense description, with LOC estimates and admittedly-unresolvable pre-migration `.c:line` citations, of work this PR ships. Items 1, 2, 4, 5 and 6 are done; only item 3 (key install) is still open, and the header already says so. Standing rule here is current-state-only docs — git is the changelog. Fold what's still load-bearing (the MBSS-mask trap, the "address match + beacon *is* the AP" finding, the measured RX-filter default) into the current-state sections and delete the rest.

## One verification gap

The MBSS mask correction changes `MT_MAC_BSSID_DW1` at **init**, on the plain injector/monitor path merged in #422/#423 — from MBSS_MODE=4 (invalid) / MBEACON_N=15 to 3 / 7. Going from invalid to correct is almost certainly an improvement, but it is not AP-scoped, and the verification here is `ctest` plus the three AP cells. One `tests/regress.py` pass on the MT7612U before merge would close that.

`qodo-gate` is failing on both runs, so the merge is blocked on that regardless.

Verdict: **changes requested** on #1; I'd want #2-#4 in the same round.

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,482 @@
...
+	 * the wrong station all session. The retarget is idempotent and costs two
+	 * EP0 writes; there is nothing to save by skipping it.
+	 *
+	 * Both registers because the index below is derived from the MBSS base,
+	 * not from MT_MAC_ADDR. */
+	{
+		const int was_taken = dev->ack_saved;
+		if (mt7612u_set_ack_responder(dev, ta))
+			return -1;
+		if (mt_mac_set_bss_base(dev, ta))
```
josephnef: **Blocking — the identity retarget still leaks here.**

Both of these `return -1`s skip `unwind_identity()`:

```c
if (mt7612u_set_ack_responder(dev, ta))
        return -1;
if (mt_mac_set_bss_base(dev, ta))
        return -1;
```

`mt7612u_set_ack_responder` (`caps.cpp:96-110`) writes `MT_MAC_ADDR_DW0/DW1` and *then* readback-verifies, so its failure path returns with the port identity already moved. `mt_mac_set_bss_base` can only fail after `MT_MAC_ADDR` moved successfully. Either way the adapter is left answering for a BSS that was never armed — the exact class `f3f08a9`/`424170b` say they closed ("a failed start left it retargeted with no beacon… every failure path unwinds now").

Compounded by the new tri-state: `-1` is contracted as "refused before the hardware was touched — whatever was airing still is", so `Mt7612uRadio::StartBeacon` deliberately leaves `_beacon_active` alone. A failed re-arm therefore reports the *previous* beacon as live while `MT_MAC_ADDR` now points at the new `ta`.

Both exits want `goto fail_pre` with `took` set from `!was_taken` (and `fail_pre` moved above this block, or `took` hoisted).

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,482 @@
...
+	/* Suppress the slot for the duration of the copy. mt_wr_copy() spans many
+	 * 64-byte EP0 transactions, so a TBTT landing mid-copy would air a TORN
+	 * beacon - leading bytes new, trailing bytes old. mt76 brackets the same
+	 * write for the same reason ("Prevent corrupt transmissions during
+	 * update", mt76x02_usb_core.c). mt_beacon_write() lowers the guard again
+	 * on its way out, which is why this is the only half needed here.
+	 *
+	 * Still no mt_beacon_init() and no set_enable(): the engine is armed, and
+	 * re-initialising it would clear the timer bits mid-flight. */
+	if (mt_wr_chk(dev, MT_BCN_BYPASS_MASK, 0xffff)) {
```
josephnef, line 603: **A failed update leaves every slot suppressed.**

This raises `BCN_BYPASS_MASK = 0xffff`, and nothing lowers it except a *successful* `mt_beacon_write`. But `mt_beacon_write` refuses on caller-controlled length at `:78-83` — checked **after** the guard is already up — and can also fail in `mt_tx_build` or on a torn reserved-page copy.

Any of those returns with the AP off the air, `_beacon_active` still true, and no path that re-lowers the guard short of a later successful update. A rejected payload should not silently take the AP down.

Check the length before raising the guard, and unwind the guard on every failure exit.

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,482 @@
...
+		/* Checked: if the guard never lands, the copy below runs against a
+		 * LIVE slot and a TBTT mid-copy airs a torn beacon - which is the one
+		 * outcome this bracket exists to prevent, so failing to raise it must
+		 * not be the quiet path. */
+		ERR("beacon: could not suppress the slot for an in-place update");
+		return -1;
+	}
+	return mt_beacon_write(dev, mpdu, mpdu_len, &rate);
```
josephnef: **`update` accepts any addr2/addr3.**

`IRadio.h:394-396` says changing addr2/addr3 mid-flight is unsupported and that the port registers keep the StartBeacon identity. `beacon_split` happily parses a beacon carrying a different BSSID and loads it into the page, so the airing BSSID stops matching the programmed APC slot and `MT_MAC_ADDR` — beacons perfectly, ACKs nobody, which is precisely the silent failure the start path now goes to real lengths to prevent.

A `memcmp` against the identity `beacon_start` took is enough to refuse it.

tests/mt7612u_ap_onair.sh
```diff
@@ -0,0 +1,300 @@
...
+  # port power-cycle is certain to silence it - and leaving one airing poisons
+  # the next run of this very script.
+  #
+  # Confirmed against the VID:PID first. This runs as root and writes a
+  # deauthorize to a path the caller supplied; a stale or mistyped AP_SYSFS
+  # would otherwise yank whatever else is plugged there - someone's keyboard,
+  # a disk mid-write.
+  if [ "$(cat "/sys/bus/usb/devices/$AP_SYSFS/idVendor" 2>/dev/null)" = "0e8d" ] &&
+     [ "$(cat "/sys/bus/usb/devices/$AP_SYSFS/idProduct" 2>/dev/null)" = "7612" ]; then
+    echo 0 > "/sys/bus/usb/devices/$AP_SYSFS/authorized" 2>/dev/null
```
josephnef: **This is not the power-cycle the comment above promises.**

The comment says "only a port power-cycle is certain to silence it", and then the implementation toggles `authorized`. This tree's own `CLAUDE.md` (Hardware testing) records that as a trap: an `authorized` toggle is not a cold cycle, it leaves chip state, and VBUS never drops — so an autonomously beaconing MAC keeps beaconing straight through it. That is the one thing this cleanup exists to prevent between cells, and `all` runs `cleanup` between all three.

Use `REGRESS_VBUS_MAP`/uhubctl like the rest of the tree, or drop the claim and say what the toggle actually guarantees.

tests/mt7612u_ap_onair.sh
```diff
@@ -0,0 +1,300 @@
...
+    local n
+    n=$(grep -c "beaconing every" "$OUT/stop.log" 2>/dev/null)
+    printf '%s' "${n:-0}"
+  }
+  wait_arm() { # $1 = the count to exceed, $2 = seconds to wait
+    local i
+    for i in $(seq 1 "$2"); do [ "$(armed)" -gt "$1" ] && return 0; sleep 1; done
+    return 1
+  }
+  wait_gone() { # the log line that says StopBeacon ran, then a settle
```
josephnef: `wait_gone()` is defined and never called — the stop cell polls for `PHASE 2` inline at the loop below instead. Dead code.

src/mt7612u/regs.h
```diff
@@ -222,6 +243,8 @@ enum mt_mcu_cr_mode { MT_RF_CR, MT_BBP_CR, MT_RF_BBP_CR, MT_HL_TEMP_CR_UPDATE };
 #define MT_RX_FILTR_CFG_PROMISC   BIT(2)
 #define MT_RX_FILTR_CFG_OTHER_BSS BIT(3)
 #define MT_RX_FILTR_CFG_VER_ERR   BIT(4)
+#define MT_RX_FILTR_CFG_MCAST     BIT(5)
```
josephnef: These two are added and unused. Related: the doc's own item-4 conclusion — clear `DUP` for an AP so a station's retransmission stays visible, which is the retry evidence that tells you whether your ACKs are landing — isn't implemented either.

src/mt7612u/Mt7612uRadio.cpp
```diff
@@ -841,6 +860,98 @@ void Mt7612uRadio::ClearAckResponder() {
...
+/* The beacon-steer trio. Not implemented, and REFUSING QUIETLY IS THE PROBLEM:
+ * IRadio's default returns 0, which means "applied a 0 us shift" and is
+ * indistinguishable from "this backend cannot steer". tests/beacon_steer_check
+ * and the PTP harnesses run against any backend and would read a clean zero.
+ * Every other unsupported knob here says so out loud; these were the exception.
+ *
+ * Steering needs a pre-TBTT interrupt to re-time against, which this static
+ * reserved-page path does not have - mt76's own steering lives in
+ * mt76x02u_pre_tbtt_work(), a path this port does not run. */
+int32_t Mt7612uRadio::AdjustBeaconTiming(int32_t microseconds) {
```
josephnef, line 921: The log is an improvement, but `IRadio.h:452` and `:507` already document `0` as the legitimate "no active beacon" answer, so a programmatic caller is exactly as blind as it was against the base class. The comment ("IRadio's default returns 0, which means *applied a 0 µs shift*") overclaims what the override changes.

Fine to keep the log; I'd trim the comment to what it actually buys — a human-visible diagnostic.

docs/mt7612u-ap-mode.md
```diff
@@ -0,0 +1,304 @@
...
+| Beacon-interval timer regs | defined — `MT_BEACON_TIME_CFG` INTVAL/TIMER_EN/TBTT_EN/BEACON_TX, `MBEACON_N` | `regs.h:176‑180,169` |
+| RX filter control | yes — managed default `0x00015f97`, monitor clears to error-only | `init.c:278,494‑509` |
+| Register block copy | yes — `mt_wr_copy()` for reserved-page writes | used in `init.c` |
+
+So the addressing, the station table, the crypto slots, the ACK path and the
+beacon *timer* are already in place. The receiver runs (per the #414 tick), and
+`mt_tx_build()` already produces `[TXWI][802.11]` which is exactly the reserved-
+page beacon shape.
+
+## The gap — the driver primitives to add
```
josephnef: This section is now a future-tense description — with LOC estimates and admittedly-unresolvable pre-migration `.c:line` citations — of work this PR ships. Items 1, 2, 4, 5 and 6 are done; only item 3 (key install) is still open, and the header at the top already says so.

Standing rule here is current-state-only docs; git is the changelog. Fold what's still load-bearing into the current-state sections — the MBSS-mask trap, the "address match + beacon *is* the AP" finding, the measured `0x00015f97` RX-filter default — and delete the rest.

josephnef (changes requested):
Second pass, same head (`424170b`), findings anchored inline this time.

Recap of where the review stands: I built the branch (`-DDEVOURER_MT7612U=ON`) — 63/63 `ctest`, `make -C src/mt7612u check` green including the new `beacon TXWI` cell, no new warnings, so the PR's verification claims hold. I also checked the register work against `reference/mt76 @ be5ce79` and it is a faithful port: the `MT_MAC_BSSID_DW1` mask correction matches `mt76x02_regs.h:281-286` bit for bit, `mt_beacon_init` matches `mt76x02_init_beacon_config` (`mt76x02_beacon.c:205-213`) step for step, the `0xff00 | ~(0xff00 >> 1)` unsuppress matches `mt76x02_usb_core.c:223`, and `init.cpp:212-216` now reproduces `mt76x02_mac_setaddr` exactly. The APC-slot premise fix is right — `mt_mac_set_bss_base()` is genuinely what makes the index derivation true rather than lucky on this bench.

One blocking finding (`beacon.cpp:349`), three worth fixing in the same round, four nits — all inline.

And one verification gap that has no line to attach to: the MBSS mask correction changes `MT_MAC_BSSID_DW1` at **init**, on the plain injector/monitor path merged in #422/#423, from MBSS_MODE=4 (invalid) / MBEACON_N=15 to 3 / 7. Going from invalid to correct is almost certainly an improvement, but it is not AP-scoped and the verification here is `ctest` plus the three AP cells. One `tests/regress.py` pass on the MT7612U before merge would close it.

`qodo-gate` is failing on both runs, so the merge is blocked on that regardless.

src/mt7612u/beacon.cpp
snokvist: Fixed in `c5c0c7e`. You're right, and it is the same class the two previous commits each said they had closed — I added `unwind_identity()` and then routed every failure path through it except the two inside the block that does the moving.

The ordering was the actual bug: ownership was claimed *after* both writes, so neither failure could be unwound even in principle. It is claimed before them now, so `goto fail_pre` reaches them.

The tri-state interaction you spotted is the worse half — `-1` means "nothing was touched" and `StartBeacon` acts on that by leaving `_beacon_active` alone, so the caller would have been told the previous beacon was live while `MT_MAC_ADDR` had already moved to the new `ta`. That is a wrong answer, not just a leak.

src/mt7612u/beacon.cpp
snokvist: Fixed. The length rule is factored into `len_fits_slot()` and applied *before* the guard goes up, and every failure exit from `mt_beacon_write` lowers it again.

Worth stating what the bug actually was: a payload rejected on a caller-controlled length took the AP off the air and left it there, with `_beacon_active` still true and no path to recover short of a later successful update. A refused update should cost the update.

src/mt7612u/beacon.cpp
snokvist: Fixed. `mt7612u_beacon_start` records the addr2 it programmed in `dev->beacon_ident`, and `update` refuses a beacon whose addr2 differs.

Agreed on why it matters: it lands in exactly the silent-failure class the start path now works hard to avoid — the airing BSSID stops matching the APC slot and `MT_MAC_ADDR`, so it beacons perfectly and acknowledges nobody, and nothing in the system reports anything wrong.

tests/mt7612u_ap_onair.sh
snokvist: You're right that the comment was wrong, and I've corrected it — but I measured the specific concern rather than assuming it, because it turns out not to hold for this part.

```
armed                -> SSID seen
host process killed  -> SSID STILL seen     (confirms the beacon is autonomous)
authorized toggle    -> SSID gone
```

So the toggle does end the beacon here — the re-enumeration stops the timer — even though CLAUDE.md is right that it is not a cold cycle: VBUS never drops and chip state survives. The claim was the problem, not the mechanism. The comment now says only what that sequence demonstrates, and names what it does not (not cold, state survives).

I did not switch to uhubctl by default: both adapters on this bench are on **xhci root ports**, which CLAUDE.md and this rig's history both put out of bounds — a root-port uhubctl here has wedged a device badly enough to need the machine powered off. So `AP_VBUS=<hubloc>:<port>` is there for anyone on a switchable hub port and takes the uhubctl path when set, with the root-port warning next to it.

If you would rather it refuse to run at all without `AP_VBUS`, say so and I will make that the default.

tests/mt7612u_ap_onair.sh
snokvist: Removed. Left over from replacing the guessed sleeps with a wait on the arm — the stop-cell rewrite inlined the PHASE-2 poll and I never deleted the helper.

src/mt7612u/regs.h
snokvist: Implemented rather than removed, because the doc's conclusion is right and the reason is about evidence rather than throughput.

`mt7612u_beacon_start` now clears `MT_RX_FILTR_CFG_DUP`. Dropping duplicates hides a station's retransmission — and a retransmitted auth with FC Retry set is exactly how you learn the MAC did *not* acknowledge the first one. That retry signal is the only evidence the AP harness has that auto-ACK is working (`auth ... retry=0` is the pass condition in `mt7612u_ap_onair.sh`), so filtering duplicates would quietly remove the check's ability to fail.

`_MCAST`/`_BCAST` stay defined and unused on purpose now, with the reason stated: mt76 leaves both undropped in every mode, so a wildcard probe request already reaches us and an AP needs no change there. Happy to delete them if you would rather the header carried only what is used.

snokvist:
Thanks — that's a thorough one, and the blocker is a fair hit. Pushed `c5c0c7e`.

**The blocker.** You're right, and it stings a bit: I added `unwind_identity()` specifically to close "a failed start leaves the identity retargeted", and then routed every failure path through it *except* the two inside the block that does the moving. The ordering was the real bug — ownership was claimed after both writes, so neither failure could be unwound even in principle. Claimed before them now.

The tri-state interaction you spotted is the worse half. `-1` contractually means "the hardware was never touched", and `StartBeacon` acts on that by leaving `_beacon_active` alone — so a failed re-arm would report the *previous* beacon as live while `MT_MAC_ADDR` had already moved. A wrong answer, not just a leak.

**#2 and #3** fixed as described: the length rule is applied before the guard goes up and every failure exit lowers it; `update` remembers the addr2 `start` programmed and refuses a beacon that would change it.

**#4, the `authorized` toggle.** The comment was wrong and is fixed. But I measured the specific concern rather than taking it on faith, and it does not hold for this part:

```
armed                -> SSID seen
host process killed  -> SSID STILL seen     (the beacon is autonomous)
authorized toggle    -> SSID gone
```

Re-enumeration stops the timer. CLAUDE.md is still right that it is not a cold cycle — VBUS never drops, chip state survives — so the comment now claims only what that sequence shows. I did not make uhubctl the default because both adapters here are on **xhci root ports**, which this rig has previously wedged to the point of needing a machine power-off; `AP_VBUS=<hubloc>:<port>` takes the uhubctl path for anyone on a switchable hub port. Say the word if you want it to refuse without `AP_VBUS`.

**Nits:** `wait_gone()` removed. The RX-filter defines are now *used* — `beacon_start` clears `DUP`, because a station's retransmission is the evidence that tells you whether your ACKs are landing, and `auth ... retry=0` is the AP harness's pass condition. `_MCAST`/`_BCAST` stay unused with the reason recorded (mt76 leaves both undropped in every mode); happy to delete them instead. The steer-trio comment overclaimed and now says plainly that a programmatic caller is no better off — the log only helps an operator reading a run. "The gap — the driver primitives to add" is rewritten as current state; five of its six items ship here and git is the changelog.

**Your verification gap, closed.** You were right that this wasn't AP-scoped — the MBSS mask correction changes `MT_MAC_BSSID_DW1` at init on the plain injector path from #422/#423, and my evidence was ctest plus three AP cells. `tests/regress.py` on the two MT7612U, ch36:

|   | TX = devourer | TX = kernel |
|---|---|---|
| RX = devourer | 9100 hits / 9000 TX ✓ | 400 / 487 ✓ |
| RX = kernel | 9231 hits / 9000 TX ✓ | 478 / 479 ✓ |

AP harness after everything: **14 passed, 0 failed**.

One open question back to you, unchanged from the PR body: hardware CCMP is the single remaining gap and it is reachable in ~40 lines over primitives already here — but `IRadio` has no key surface at all, so it means adding an interface member most backends cannot implement. That is your call rather than mine; happy to open it as its own issue.

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,533 @@
...
+		/* Ownership is claimed BEFORE the writes, not after them. Both calls
+		 * below move MT_MAC_ADDR and can then fail - set_ack_responder writes
+		 * the register and readback-verifies afterwards, and set_bss_base can
+		 * only be reached once that write landed - so a failure here leaves
+		 * the identity moved. Claiming first is what lets the unwind put it
+		 * back; claiming after meant these two paths leaked it, and returned
+		 * -1, which tells the caller the hardware was never touched. */
+		const int was_taken = dev->ack_saved;
+		if (!was_taken) {
+			dev->beacon_took_identity = 1;
```
josephnef: **Blocking — the reordering broke the success path.**

`mt7612u_set_ack_responder()` sets `d->beacon_took_identity = 0` itself (`caps.cpp:90`, the "ownership TRANSFERS to this caller" write added in `424170b`). Claiming ownership *before* that call means the claim is immediately clobbered:

```c
if (!was_taken) { dev->beacon_took_identity = 1; took = 1; }   /* :365 */
if (mt7612u_set_ack_responder(dev, ta))                        /* :368 -> sets it back to 0 */
```

Nothing sets it again, so after a **successful** `beacon_start` the flag reads 0. `mt7612u_beacon_stop()` at `:530` passes exactly that flag to `unwind_identity()`, so the stop never restores the factory MAC or the MBSS base — the port stays retargeted and keeps auto-ACKing for a BSS that no longer exists. That is the residue `f3f08a9` added the whole retract-the-whole-identity block to prevent, and it is now unreachable on the only path that matters.

The failure paths are fine, because they use the local `took`. It is the success path that regressed.

Simplest fix that keeps both properties: claim before the writes as you have it, then re-assert `dev->beacon_took_identity = took;` after `mt_mac_set_bss_base()` succeeds — next to the `memcpy(dev->beacon_ident, ta, 6)` that is already there. Worth a contract check in `mt7612u_beacon_stop_check.cpp` too: after a start/stop pair, `MT_MAC_ADDR_DW0` should read back the factory address.

src/mt7612u/beacon.cpp
```diff
@@ -0,0 +1,533 @@
...
+	 * The one RX-filter change an AP wants, and the reason is evidence rather
+	 * than throughput: DUP drops retransmissions, and a station's retry is
+	 * exactly how you learn whether your ACKs are landing. An auth arriving
+	 * with FC Retry set means the MAC did not acknowledge the first one; drop
+	 * duplicates and that signal disappears. mt76 leaves OTHER_BSS, BCAST and
+	 * MCAST undropped in every mode, so a wildcard probe request already
+	 * reaches us and nothing else here needs changing. Cleared in place rather
+	 * than rewriting a copied literal, so it cannot drift from the default.
+	 */
+	mt_clear(dev, MT_RX_FILTR_CFG, MT_RX_FILTR_CFG_DUP);
```
josephnef: **`DUP` is cleared here and never restored.**

I agree with the change itself — the reasoning that a station's retry is the evidence your ACKs are landing is right, and it is what `auth … retry=0` in the harness depends on.

But this is the same class of residue the stop path deliberately retracts three functions down. `mt7612u_beacon_stop()` zeroes both APC slots and (intends to) put the port MAC back, specifically so "a session that stops beaconing and carries on as an injector or a monitor" does not drag AP state with it. `MT_RX_FILTR_CFG` is left modified: a monitor session on the same handle after a `StopBeacon` silently gets duplicate frames it did not ask for, and nothing says why.

Also note it sits **above** `fail_pre`, so a start that fails at `mt_ap_set_bssid` leaves it cleared too, on a path contracted as "the hardware was never touched".

Save the bit's prior state and restore it in `unwind_identity()` (or alongside the APC-slot zeroing in stop).

josephnef (changes requested):
Re-reviewed at `c5c0c7e`. Rebuilt: 63/63 `ctest`, `make -C src/mt7612u check` green, no new warnings.

Findings 2, 3 and the nits are all properly closed — `len_fits_slot()` factored ahead of the guard with an unwind on the write failure, `beacon_ident` refusing an addr2 change, `wait_gone()` gone, the steer-trio comment trimmed to what it delivers, and "The gap" rewritten as current state. Thank you for the `regress.py` run on the injector path; 4/4 green closes the verification gap I flagged, and that was the answer I wanted rather than an argument.

On **#4**: your measurement beats my objection, and I'll take it. `authorized` → SSID gone is the thing that matters here, and the comment now claims only that. `AP_VBUS` for a switchable hub port is the right shape — no, don't make it mandatory; your bench's root ports are exactly the case that would make the script unrunnable for the person most likely to run it.

On **#1**: the fix moved the bug rather than removing it. `mt7612u_set_ack_responder()` clears `beacon_took_identity` itself, so claiming ownership *before* calling it means the claim is wiped on the success path and `StopBeacon` no longer restores the identity. Detail inline at `:365`. Failure paths are correct now — it is the success path that regressed.

One new one at `:418` on the `DUP` clear, which I think is the right change made in the wrong place: it is never restored, including on a `fail_pre` exit that is contracted as "hardware never touched".

**On hardware CCMP** — open it as its own issue, and don't add a key surface to `IRadio` for it. A method most backends answer with a not-ported default is the pattern this tree already uses, but a *crypto key* interface is a much bigger contract than `FastRetune` (key lifetime, GTK vs PTK, rekey, who owns the replay counter), and getting it wrong is expensive to undo once callers exist. It also isn't blocking anything: the software CCMP path is shipped, measured here, and shared with the Realtek backends. Land the beacon plane, then design the key surface against two backends rather than one.

Everything else stands; happy to approve once `:365` and `:418` are closed.

src/mt7612u/beacon.cpp
snokvist: Fixed. You're right, and the flag was wrong in a second way that only showed up while fixing the first.

`dev->beacon_took_identity` is now set **after** `mt7612u_set_ack_responder()`, so the ownership-transfer write at `caps.cpp:90` no longer clobbers the claim. The claim still has to exist *before* that call for the `fail_pre` paths inside the retarget block — that is what the reordering was for — so it is now a local (`took`) set before, and the device flag set after. Both things are true at once, which is what the previous two attempts each got half of.

The second thing: `took` used to be `!dev->ack_saved`, meaning "somebody else got here first, leave it to them". That is wrong on both halves.

- `ack_saved` is also set by this function's own `set_ack_responder` call, and is only cleared by a successful stop. So a re-arm over a live beacon, and any session whose config sets `rx.ack_responder`, both took the "somebody else" branch and disabled the restore for the rest of the session.
- And the branch protected nothing even when it fired correctly: the retarget had already overwritten that responder's address in hardware, so declining to restore left `MT_MAC_ADDR` at the beacon's addr2 — neither the responder's address nor the factory one.

It is unconditional now. Hand-off is the responder's job, which is exactly what `caps.cpp:90` is.

**And one you have not seen, because I introduced it in `c5c0c7e` and found it this round.** `unwind_identity()`'s comment claimed the flags survive a failed restore "so `Stop()`'s documented retry has something left to retry". That was true for the MBSS half and **false for `MT_MAC_ADDR`**: `mt7612u_clear_ack_responder()` cleared `ack_saved` unconditionally and early-returns on `!ack_saved`, so attempts two and three were no-ops against a still-leaked port MAC — the retry the comment promised could not reach the register it was written for. It keeps `ack_saved` on its own I/O failure now, same `mt_io_errors()` delta idiom, and the comment says why the two halves have to agree. Flagging it rather than letting it be the fourth round of the same pattern.

src/mt7612u/beacon.cpp
snokvist: **Removed, not restored** — the clear was wrong to begin with, and the save/restore you asked for would have been actively harmful.

Your residue argument is right, and I started writing the save/restore. What stopped me is that the DUP clear was already a no-op on every path that beacons. `mt7612u_set_monitor_rx()` (`init.cpp:552-567`) rewrites `MT_RX_FILTR_CFG` to `PHY_ERR` alone (plus `CRC_ERR` when keeping corrupted frames), so DUP is already clear before `beacon_start` runs — deliberately, for exactly the reason we agree on. The comment there says duplicate suppression "would hide the retransmissions an ACK-responder test counts". Every AP path reaches it: `StartRxLoop` calls it, and an AP has to receive.

So `mt_clear` was clearing a bit that was already clear, and the restore would have **set** DUP at `StopBeacon` in a session that deliberately had it off — switching duplicate filtering on and destroying the `retry=0` evidence that is the AP harness's only pass condition for auto-ACK. That is worse than the residue it was protecting against.

Your point about it sitting above `fail_pre` was independently correct, and is moot now.

I also rewrote the comment that replaced it, because the first version asserted the same thing wrongly: it said "this backend's receive default is already what an AP wants". The *default* is `0x00015f97` (`initvals.h`, rewritten at `init.cpp:290`) and `MT_RX_FILTR_CFG_DUP` is **set** in it. It is `set_monitor_rx` that leaves DUP clear, not the default — and a TX-only consumer (`InitWrite` with no RX loop) keeps DUP set, and has no receiver to count retries with anyway. Checkable in ten seconds against `initvals.h`, which is the kind of claim a comment should not get wrong.

src/mt7612u/beacon.cpp
snokvist: Following up on the second half of this, which I under-fixed at the time.

You wrote that the leak was "compounded by the new tri-state: `-1` is contracted as 'refused before the hardware was touched — whatever was airing still is', so `Mt7612uRadio::StartBeacon` deliberately leaves `_beacon_active` alone." I routed those two exits through `unwind_identity()` and left them returning `-1`. That closed the leak and left the contract false.

It is false because the unwind restores the **factory** address — the only address saved anywhere. `ack_saved_mac` is written once, from `d->macaddr`, which itself comes from the EEPROM. So on a re-arm over a live beacon A, a stall in `set_ack_responder`'s readback gives: identity and MBSS base retracted to factory, A's reserved page and timers untouched and still airing, and the caller told `-1` — nothing changed, keep believing in your AP. A beacons perfectly and acknowledges nobody, reached *through* the guard, which is the exact failure this file is written to prevent.

There is no atomic re-arm available: one `MT_MAC_ADDR`, one save slot, and the previous occupant's address is not in it. So those two exits go to `fail_post` now — engine disarmed, both APC slots zeroed, identity retracted, `-2`, and `Mt7612uRadio::StartBeacon` clears `_beacon_active`. A failed re-arm takes the AP down deliberately and says so, rather than leaving a live beacon the host no longer has an identity for. `fail_pre` is deleted; `-1` now belongs to the refusals above the first hardware write — bad input, malformed frame, multicast addr2 — and means what it says with no qualification.

Worth noting what it cost to find: this one came from an adversarial pass over my own diff, not from the on-air harness. It cannot be caught by watching the air, because the beacon keeps airing perfectly throughout.

src/mt7612u/beacon.cpp
snokvist: Correcting my earlier reply here: that fix was **half** of what you asked for, and I described it as if it were all of it.

You wrote addr2/addr3. I implemented addr2 only — `memcmp(mpdu + 10, dev->beacon_ident, 6)`. addr3 is the half that goes into the APC slot (`mt_ap_set_bssid(dev, idx, bssid)`, `bssid = mpdu + 16`), so an update that moved only addr3 aired a BSSID the slot does not hold and produced *precisely* the deaf AP this guard exists to prevent — through the guard. `IRadio.h:394-396` names both, and I quoted the sentence while implementing one of the two addresses in it.

Both now. `beacon_ident` is 12 bytes and `beacon_start` fills it with one `memcpy(ta, 12)` — addr2 and addr3 are adjacent in the 24-byte management header `beacon_split()` requires, so it is one compare, not two.

It is also tested rather than asserted now. `tests/mt7612u_beacon_stop_check.cpp` runs three checks against a **live** beacon: an unchanged payload must be *accepted*, a changed addr2 must be refused, a changed addr3 must be refused. The positive control is the point — a guard that refused everything would pass both negative arms, and the addr3 arm fails against the pre-fix code, so neither is structurally guaranteed. All three run before the suppress/copy bracket, so a refusal costs the update and not the beacon; the scan in the same phase confirms the AP is still airing afterwards.

`14 passed, 0 failed` on ch36 and on ch6 with these in place.

src/mt7612u/regs.h
snokvist: Correcting my previous reply on both halves.

**The DUP clear is gone**, not implemented — see the `:418` thread. `mt7612u_set_monitor_rx()` already leaves DUP clear on every path that beacons, so `mt_clear` was a no-op, and restoring it at stop (which is what the residue argument correctly asked for) would have switched duplicate filtering *on* and destroyed the `retry=0` evidence I argued for here. The reasoning in my reply was right; the implementation was unnecessary, and the restore would have been harmful. `src/mt7612u/tools/bringup.cpp:787` already clears DUP for its own gate, which is the third place this conclusion shows up.

**`_MCAST`/`_BCAST` did not have "the reason stated"** — I wrote that they stay defined and unused "with the reason stated", and `regs.h` carries no such comment; the reasoning existed only in this thread. That was a false claim about the code in a reply defending the code.

Both are deleted now. They were the two this branch added (`git diff origin/master...HEAD -- src/mt7612u/regs.h` shows only those two), and neither is used. `MT_RX_FILTR_CFG_DUP` stays because it predates this branch and `bringup.cpp` uses it.

src/mt7612u/Mt7612uRadio.cpp
snokvist: Trimmed to that. The comment now says the log buys a human-visible diagnostic and nothing more — explicitly, that a programmatic caller cannot tell "cannot steer" from "steered by nothing" any better than against the base class, because `IRadio.h:452`/`:507` document `0` as the legitimate "no active beacon" answer.

Kept the note on *why* there is nothing to steer: the static reserved-page path has no pre-TBTT hook, and mt76's own steering lives in `mt76x02u_pre_tbtt_work()`, which this port does not run.

I also dropped the corresponding overclaim from the PR body, which said the trio "silently returned 'applied a 0 µs shift'".

docs/mt7612u-ap-mode.md
snokvist: Done. "The claim, and why it holds" and "What the MT7612U backend already has" are deleted — both were future-tense descriptions of work this PR ships, with LOC estimates and pre-migration `.c:line` citations.

What survives is one current-state section, "Three findings that shaped the implementation", carrying the three things you named as load-bearing plus the reason each one matters to the code:

1. **Address match + beacon *is* the AP** — there is no responder register; `MT_AUTO_RSP_EN` is already on from `mt_init_hardware`, so arming means retargeting the port identity, and `MT_MAC_ADDR` therefore has two users sharing one save slot.
2. **The MBSS/APC trap** — the slot index derives from the MBSS base, not from `MT_MAC_ADDR`, so moving one register and not the other is silent; plus the two-bits-high mask transcription.
3. **The measured `0x00015f97` RX-filter default** — and, corrected from the earlier draft, that DUP is **set** in it and it is `set_monitor_rx` that clears it.

Also removed: the header's "citations in The gap section ... do not expect them to resolve" caveat (there is no such section any more, and no unresolvable citations left), the dangling "the claim below" reference, and the future-tense framing of the end-to-end verification section. The remaining `file:line` citations all resolve against the merged subtree.

snokvist: Opened as #425, deliberately **not** proposing an `IRadio` key surface — per your guidance that the contract should be designed against two backends rather than one. It records the four things actually missing (`MT_WCID_KEY`/`MT_WCID_IV` absent, `MT_TXD_INFO_WIV` set unconditionally where mt76 gates it on the WCID having a key, no library path installing a station, and RX having to strip the IV/PN itself), and names the two things that would make it worth doing: GTK rekey, and a measurement of software CCMP's per-frame cost — which nobody has taken, so "hardware crypto is faster here" is currently an assumption.

snokvist:
@josephnef pushed `6465ab1`. Both your open threads are answered, and two of the five fixes in it are defects this branch introduced — one of them in the first draft of this very round.

**`:365` — the ordering.** You were right. `beacon_took_identity` is set *after* `set_ack_responder` now, so the ownership-transfer write at `caps.cpp:90` no longer clobbers the claim, while `took` is still decided before the writes so the failure paths inside the retarget block know we own it. The `!ack_saved` guard is gone too, for a second reason: `ack_saved` is also set by this function's own call, so a re-arm over a live beacon and any session with `rx.ack_responder` both took the "somebody else got here first" branch and disabled the restore for the session — and the branch protected nothing when it fired, because the retarget had already overwritten that responder's address.

**`:418` — removed, not restored.** Your residue argument is right and I started writing the save/restore. It turns out the clear was a no-op: `mt7612u_set_monitor_rx()` already leaves DUP clear on every path that beacons, deliberately, for the same evidence reason. So restoring it at stop would have switched duplicate filtering *on* in a session that had it off and destroyed the `retry=0` evidence — worse than the residue. The comment that replaced it was wrong in the same way ("the default is already what an AP wants"; `0x00015f97` has DUP **set**), and is fixed.

**Three you have not seen**, because they came out of an adversarial pass over my own diff rather than from the bench:

1. Fixing the retry widened `beacon_start`'s `io_err` bracket over a `mt_set(MT_AUTO_RSP_EN)` that the code itself documents as a no-op. `mt_set` is `mt_rmw`, which bumps `io_err` and skips its write when its *read* half stalls, and its return is ignored because the readback is the real check — so one transient EP0 read stall tore down a verified-good arm and returned `-2` for a beacon that was on the air.
2. `-1` was escaping from inside the identity block, where "nothing was touched" is already false. The unwind restores the **factory** address — the only one saved anywhere — so a failed re-arm over a live beacon retracted an identity that beacon never had while its page and timers kept airing it. Those exits return `-2` now and `fail_pre` is deleted; details in your tri-state thread above.
3. `beacon_update` was guarding addr2 only. You wrote addr2/addr3 and I implemented one of them, then described it as done.

Every one of those leaves the beacon airing perfectly, which is why the harness could not find them.

**`regs.h`:** `_MCAST`/`_BCAST` are deleted — they were the two this branch added and nothing uses them. My earlier reply claiming they carried "the reason stated" was false; `regs.h` had no such comment.

**Docs:** the future-tense section is gone. Three findings survive it as current-state prose — the address-match-is-the-AP property, the MBSS/APC index trap, and the measured RX-filter default. Also corrected: the harness takes `CH=`, not `DEVOURER_CHANNEL=` (which it *exports to the child*, so the documented invocation would have run ch36 with a 5180 MHz scan).

**Hardware CCMP is #425**, deliberately without an `IRadio` key surface — your call that it should be designed against two backends. It also names the thing nobody has measured: software CCMP's per-frame cost on this stack, without which "hardware crypto is a gain here" stays an assumption.

**Verification on `6465ab1`:** 63/63 `ctest` both with and without `DEVOURER_MT7612U`, subtree `check` green, and `tests/mt7612u_ap_onair.sh` **14/14 on ch36 and 14/14 on ch6** — both bands carrying association, data plane, WPA2 and `auth … retry=0`, not just a beacon. The update guard now has a positive control alongside its two negative arms; only the addr3 arm discriminates against the pre-fix code, and the test says so rather than counting both as new coverage.

I resolved the twelve qodo threads that are fixed and left yours for you.

src/mt7612u/beacon.cpp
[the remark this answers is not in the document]
snokvist: Follow-up, because the fix this thread describes was revised twice more after that reply and the final shape is different.

"Ownership transfers to whoever wrote last" is still the rule, but two things had to be corrected before it actually held:

1. `mt7612u_beacon_start` was claiming ownership *before* calling `mt7612u_set_ack_responder`, which clears that very flag as part of the transfer — so the claim was wiped and a **successful** start ended with the flag reading 0. `StopBeacon` then never restored the identity on the success path. The claim is now split: a local decided before the writes (so the failure paths inside the retarget block know we own it), the device flag set after (so the transfer wins).
2. The restore had a retry that could not reach `MT_MAC_ADDR`. `mt7612u_clear_ack_responder` cleared `ack_saved` unconditionally and early-returns on `!ack_saved`, so the second and third attempts `Mt7612uRadio::Stop()` makes were no-ops against a still-leaked port MAC. It keeps the flag on an `mt_io_errors()` delta now.

Also relevant to this finding specifically: `took` used to be `!ack_saved`, i.e. "somebody else got here first". That was wrong in both directions — `ack_saved` is also set by `beacon_start`'s own call, so a re-arm over a live beacon disabled the restore for the session; and when it did fire, the retarget had already overwritten the responder's address, so declining to restore left `MT_MAC_ADDR` at the beacon's addr2 rather than the responder's. It is unconditional now.

Device-verified in `6465ab1`: 14/14 on ch36 and 14/14 on ch6.

josephnef (approved):
Verified at `6465ab1` — rebuilt, 63/63 `ctest`, subtree `check` green, no new warnings. Both my threads are closed, and closed properly rather than papered over. Resolving them; **approving**.

**`:365`.** Right shape now: `took` decided before the writes so the in-flight failures know we own the identity, `dev->beacon_took_identity` set after `set_ack_responder` so its ownership-transfer write at `caps.cpp:90` can't clobber the claim. I checked the success path is the one that actually changed — that was the regression.

Dropping the `!ack_saved` guard is the better call and your reasoning for it is the part I'd have missed: `ack_saved` is set by this function's *own* call, so a re-arm over a live beacon took the "somebody else got here first" branch and disabled the restore for the session — while the retarget had already overwritten the address the branch was nominally protecting. A guard that fires on its own side effect and protects nothing when it does is worse than no guard.

**`:418`.** Removing it beats the save/restore I asked for, and for a reason I had backwards: `mt7612u_set_monitor_rx()` rewrites `MT_RX_FILTR_CFG` to `PHY_ERR` (+`CRC_ERR`) wholesale, so DUP is already clear on every path that beacons — `Mt7612uRadio::StartRxLoop():407`, and an AP has to receive. Restoring it at stop would have switched duplicate filtering *on* in a session that had it off and destroyed the `retry=0` evidence. My finding was right about the residue and wrong about the fix; good catch not taking it at face value.

**The three you found yourselves** are the interesting ones, and all three are the same shape — the beacon airs perfectly and the harness cannot see the defect:

- The `io_err` bracket over `mt_set(MT_AUTO_RSP_EN)` is the sharpest. `mt_rmw` bumping the counter on a failed *read* of a bit the code documents as already-set, tearing down a verified-good arm and returning `-2` for a beacon that is on the air — that is a false negative manufactured by a fix for a false positive. `mt_io_restore()` around it is right, and it is the existing pattern (`fw.cpp:299`), not a new mechanism.
- `-1` escaping from inside the identity block: agreed, and deleting `fail_pre` is the right resolution rather than trying to make the promise true. I checked the flow — every `-1` in `beacon_start` is now above the `mt_io_errors()` snapshot at `:388`, and everything past it goes to `fail_post`. The tri-state means what it says now.
- addr3: fair. I wrote "addr2/addr3" and only read back the addr2 half of what shipped.

**The test.** Saying which arm discriminates, and keeping a positive control against a guard that could pass both negatives by refusing everything, is the right instinct — a test whose arms aren't told apart reads as twice the coverage it has. Same for the `CH=` vs `DEVOURER_CHANNEL=` correction; a documented invocation that runs ch36 with a 5180 MHz scan is the kind of thing that costs someone an afternoon.

**Verification.** ch36 *and* ch6, 14/14 each, carrying association, data plane, WPA2 and `auth … retry=0` — that is a stronger claim than the single-band table the PR body started with, and it is the right one to have made before merge. `regress.py` closed the injector-path question. All checks green including `qodo-gate`.

Ship it. #425 is the right home for hardware CCMP, and naming software CCMP's unmeasured per-frame cost in it is what keeps "hardware crypto is a gain here" honest until someone measures it.

PR #427, opened 2026-09-12
IRtlRadio: address the two carrier-sense gates separately

snokvist:
`SetCcaMode` is all-or-nothing, and on these families it is two gates doing different jobs: `0x520[14]` primary CCA defers to a decodable preamble, `0x520[15]` EDCCA defers to raw in-band energy. This adds `SetCcaGates` / `GetCcaGates` so a caller can address them one at a time, with not-ported defaults, and reduces `SetCcaMode` to `SetCcaGates(d, d)`.

## Why split them

They do not behave the same way — and on Jaguar1 they behave *oppositely* to what `tests/dis_cca_tx_onair.sh` found.

That test is Jaguar3. It uses the 8812AU only as its flooder and has never run a Jaguar1 as the DUT. Running one — RTL8812AU injecting 300 broadcast frames at 6M on an otherwise idle ch6, **fresh radio open per arm** so nothing latches across points, two independent receivers, repeated:

| Gate state | delivered |
|---|---|
| both gates on (the default) | 0.0%, 1.7% |
| **EDCCA off only** | **94.3%, 94.7%** |
| primary CCA off only | 13.7%, 2.7% |
| both off | 94.3%, 95.3% |

On Jaguar1 the energy bit is everything and the preamble bit is nearly null. Without the split there is no way to discover that, and no way to act on it except `dis_cca`, which turns off both.

## Acting on it beats dis_cca

Same DUT, against a saturating co-channel flooder:

| Arm | flooder | delivered |
|---|---|---|
| EDCCA off, primary CCA **ON** | no | 95.3% |
| EDCCA off, primary CCA **ON** | yes | 78.0% |
| both gates off | yes | **0.3%** |

Carrier sense still defers (95% → 78% when a real transmitter takes the channel), and turning both gates off is *worse for the injector's own delivery* — it transmits into the flood and collides.

## No behaviour change for existing callers

`SetCcaMode`'s two states write exactly the bytes they wrote before:

- **Jaguar1** — `apply_cca(d, d)` is the old body, with the BB threshold half at `0x8a4` keyed on the EDCCA argument.
- **Jaguar3** — `0x524[11]` moves with the pair and *only* with the pair. What that bit does on its own is undocumented and unmeasured, so a mixed state leaves it at the enabled value rather than guessing. Jaguar3 also tracks the two gates separately so `SetMonitorChannel` re-asserts what the caller asked for, not just the all-or-nothing pair.

## Verification

Hardware: **RTL8812AU (Jaguar1)** and **RTL8822C (Jaguar3, an 8812CU)**. All four gate states read back from `0x520` on both, and the Jaguar3 state survives a retune.

Defaults checked against a build *without* this change, alternating the two builds on the same bench so ambient drift shows as baseline-vs-baseline disagreement:

```
RX, ambient frames in 5 s on ch6      baseline        patched
  RTL8812AU                             0 / 0          0 / 0
  RTL8822C                             99 / 58        59 / 86
  MT7612U                             172 / 138      142 / 139     zero bridge drops throughout

TX, default carrier sense, 300 frames @ 6M, best independent witness
  RTL8812AU                          3.7% / 3.0%    5.0% / 7.7%
  RTL8822C                         100.0% / 100.0%  100.0% / 100.0%
  MT7612U                          100.0% / 99.7%   100.0% / 100.0%
```

Every difference is inside the baseline-to-baseline spread. `cmake --build` clean and `ctest` 63/63 on this branch.

**Not verified here:** Jaguar2, Kestrel and RTL8733B have no hardware on this bench and stay on the not-ported default, as does MediaTek, which is not an `IRtlRadio` at all.

## A note on the CLAUDE.md claim

The current text reads as general:

> The primary-CCA bit is the one that matters … the energy bit `[15]` alone is null against a decodable preamble.

It is accurate for Jaguar3 and inverted on Jaguar1. Happy to add the scoping and the Jaguar1 numbers in this PR or a follow-up, whichever you prefer.

Separately, this bench found that Jaguar1's bring-up *enables* EDCCA — programming `0x8a4` off its parked never-trigger default, which the code comment itself identifies as the vendor's adaptivity-off default (`CONFIG_RTW_ADAPTIVITY_EN 0`) — and hard-codes `th_l2h_ini` and the H2L gap, both of which the vendor driver exposes as runtime module parameters. Measured cost of that default on an 8812AU injector is the 94% above. That is a policy question rather than a bug, so it is not in this PR; glad to raise it as an issue if useful.

### DIFF
--- a/src/IRtlRadio.h
+++ b/src/IRtlRadio.h
@@ -85,30 +85,62 @@ public:
    * tests/canary_diff.py. Reading a powered-down chip yields garbage or throws;
    * interpreting that is the caller's job. No-op where unsupported (default). */
   virtual void DumpChipState() {}
-
   /* The MAC carrier-sense gate, one bit at a time.
    *
-   * SetCcaMode is all-or-nothing, and on this family it is two gates:
-   * 0x520[14] primary CCA (defers to a decodable preamble) and 0x520[15]
-   * EDCCA (defers to raw in-band energy). They answer different questions
-   * and they do not behave the same way — tests/dis_cca_tx_onair.sh measured
-   * primary CCA costing a Jaguar3 injector 41-45% against a co-channel
-   * flooder while the energy bit alone was null, and on Jaguar1 the result
-   * inverts (see below). A caller that needs one of them should not have to
-   * turn off both, and a caller diagnosing a deferral needs to tell them
-   * apart.
+   * SetCcaMode is all-or-nothing, and on Jaguar1 and Jaguar3 it is two
+   * gates: 0x520[14] primary CCA (defers to a decodable preamble) and
+   * 0x520[15] EDCCA (defers to raw in-band energy). They answer different
+   * questions, and the two families measured so far DISAGREE about which one
+   * stops an injector — so a caller diagnosing a deferral has to tell them
+   * apart, and one that needs a single gate should not have to turn off
+   * both. CLAUDE.md summarises the on-air delivery figures and
+   * tests/dis_cca_tx_onair.sh is the harness behind them;
+   * tests/cca_gates_regcheck.sh is the register-level check that this
+   * contract holds, not a delivery measurement.
    *
    * `true` means DISABLED, matching SetCcaMode's argument sense and the
    * register's own polarity (bit set = gate off). SetCcaMode is exactly
-   * SetCcaGates(d, d) and writes the same bytes it always did. Returns false
-   * where the split is not ported; SetCcaMode remains the portable call. */
+   * SetCcaGates(d, d) and writes the same bytes it always did; it stays the
+   * portable call, and is all a backend without the split offers.
+   *
+   * CONTRACT, because both halves of this have bitten:
+   *
+   *  - POST-BRING-UP ONLY. Both calls return false before Init/InitWrite:
+   *    0x520 is meaningless until the MAC is configured, so reading it would
+   *    be a fabricated gate state and writing it would poke an uninitialised
+   *    MAC. `false` therefore means EITHER "not ported on this backend" OR
+   *    "not brought up yet"; a caller probing capability at construction
+   *    cannot tell those apart and must re-ask after bring-up. On a refusal
+   *    GetCcaGates leaves its out-parameters untouched.
+   *
+   *    SetCcaMode is NOT the same, and the difference is pre-existing rather
+   *    than something the split introduced: it returns void, so a
+   *    pre-bring-up call cannot report anything, and what it does with one
+   *    is per-backend. The way to ask for a gate state from bring-up is the
+   *    tuning.disable_cca config knob, which Init applies once the MAC is
+   *    up.
+   *
+   *  - STICKINESS SURVIVES A RETUNE ON BOTH, BUT ONLY ONE OF THEM MEANS IT.
+   *    Measured on an 8812AU and an 8822C, the gate state is intact after
+   *    SetMonitorChannel AND after FastRetune, both within a band and across
+   *    a 5 GHz/2.4 GHz change, at 0x520, 0x524 and Jaguar1's BB thresholds.
+   *    The mechanisms are not equivalent: Jaguar3 records the pair and
+   *    re-asserts it in SetMonitorChannel (its FastRetune fallback does not,
+   *    and does not need to), while Jaguar1 records nothing and survives
+   *    only because its channel path happens not to rewrite those registers.
+   *    Do not build on the Jaguar1 case — re-read with GetCcaGates rather
+   *    than assume. Bring-up IS a reset on Jaguar1: Init/InitWrite
+   *    unconditionally re-run SetCcaMode(_cfg.tuning.disable_cca), so a
+   *    re-Init puts the gates back to the configured default. */
   virtual bool SetCcaGates(bool primary_disabled, bool edcca_disabled) {
     (void)primary_disabled;
     (void)edcca_disabled;
     return false;
   }
 
-  /* Current gate state, read back from the hardware rather than remembered. */
+  /* Current gate state, read back from the hardware rather than remembered.
+   * Same contract as SetCcaGates above: post-bring-up only, false where the
+   * split is unavailable, out-parameters untouched on a refusal. */
   virtual bool GetCcaGates(bool &primary_disabled, bool &edcca_disabled) {
     (void)primary_disabled;
     (void)edcca_disabled;
--- a/src/IRtlRadio.h
+++ b/src/IRtlRadio.h
@@ -85,6 +85,35 @@ public:
    * tests/canary_diff.py. Reading a powered-down chip yields garbage or throws;
    * interpreting that is the caller's job. No-op where unsupported (default). */
   virtual void DumpChipState() {}
+
+  /* The MAC carrier-sense gate, one bit at a time.
+   *
+   * SetCcaMode is all-or-nothing, and on this family it is two gates:
+   * 0x520[14] primary CCA (defers to a decodable preamble) and 0x520[15]
+   * EDCCA (defers to raw in-band energy). They answer different questions
+   * and they do not behave the same way — tests/dis_cca_tx_onair.sh measured
+   * primary CCA costing a Jaguar3 injector 41-45% against a co-channel
+   * flooder while the energy bit alone was null, and on Jaguar1 the result
+   * inverts (see below). A caller that needs one of them should not have to
+   * turn off both, and a caller diagnosing a deferral needs to tell them
+   * apart.
+   *
+   * `true` means DISABLED, matching SetCcaMode's argument sense and the
+   * register's own polarity (bit set = gate off). SetCcaMode is exactly
+   * SetCcaGates(d, d) and writes the same bytes it always did. Returns false
+   * where the split is not ported; SetCcaMode remains the portable call. */
+  virtual bool SetCcaGates(bool primary_disabled, bool edcca_disabled) {
+    (void)primary_disabled;
+    (void)edcca_disabled;
+    return false;
+  }
+
+  /* Current gate state, read back from the hardware rather than remembered. */
+  virtual bool GetCcaGates(bool &primary_disabled, bool &edcca_disabled) {
+    (void)primary_disabled;
+    (void)edcca_disabled;
+    return false;
+  }
 };
 
 #endif /* IRTL_RADIO_H */
--- a/src/jaguar1/RtlJaguarDevice.cpp
+++ b/src/jaguar1/RtlJaguarDevice.cpp
@@ -1023,6 +1023,19 @@ void RtlJaguarDevice::apply_cca(bool primary_disabled, bool edcca_disabled) {
     v520 &= ~(1u << 15);
   _device.rtw_write<uint32_t>(0x0520, v520);
 
+  /* Stop the EDCCA tracker BEFORE touching 0x8a4, not after. The phydm
+   * watchdog owns that register while tracking, and it runs on its own
+   * thread from rtw_hal_init — i.e. already before bring-up's SetCcaMode.
+   * Clearing the flag last left a window in which a tick could re-derive
+   * L2H from IGI and overwrite the park, leaving live thresholds behind a
+   * disable the caller had asked for. SetEdccaTrack is synchronous, so once
+   * it returns the writes below are ours. The enable direction hands the
+   * register over only after it is programmed, at the end of this function.
+   * Harmless when no watchdog was built (the default config). */
+  if (edcca_disabled)
+    if (auto *wd = _halModule.phydm_watchdog())
+      wd->SetEdccaTrack(false);
+
   /* BB EDCCA thresholds (rEDCCA_Jaguar 0x8a4: L2H byte0 / H2L byte1). The
    * BB init table parks them at 0x7f/0x7f = never-trigger — the vendor's
    * adaptivity-off default (CONFIG_RTW_ADAPTIVITY_EN 0). Parked, the BB
@@ -1049,9 +1062,11 @@ void RtlJaguarDevice::apply_cca(bool primary_disabled, bool edcca_disabled) {
                   l2h, l2h - 7, igi);
   }
   /* With the watchdog running, DIG walks IGI — hand it the re-track so the
-   * threshold follows (vendor couples them per adaptivity cycle). */
-  if (auto *wd = _halModule.phydm_watchdog())
-    wd->SetEdccaTrack(!edcca_disabled);
+   * threshold follows (vendor couples them per adaptivity cycle). Only the
+   * enable direction is done here; the disable ran above, before the park. */
+  if (!edcca_disabled)
+    if (auto *wd = _halModule.phydm_watchdog())
+      wd->SetEdccaTrack(true);
 }
 
 bool RtlJaguarDevice::SetAmpduMode(const devourer::AmpduMode &mode) {
--- a/src/jaguar1/RtlJaguarDevice.cpp
+++ b/src/jaguar1/RtlJaguarDevice.cpp
@@ -980,16 +980,38 @@ void RtlJaguarDevice::ClearAckResponder() {
   (void)disarm_ack_responder();
 }
 
+bool RtlJaguarDevice::GetCcaGates(bool &primary_disabled, bool &edcca_disabled) {
+  const uint32_t v = _device.rtw_read<uint32_t>(0x0520);
+  primary_disabled = (v & (1u << 14)) != 0;
+  edcca_disabled = (v & (1u << 15)) != 0;
+  return true;
+}
+
+bool RtlJaguarDevice::SetCcaGates(bool primary_disabled, bool edcca_disabled) {
+  apply_cca(primary_disabled, edcca_disabled);
+  _logger->info("Jaguar1: CCA gates primary={} edcca={}",
+                primary_disabled ? "OFF" : "on",
+                edcca_disabled ? "OFF" : "on");
+  return true;
+}
+
 void RtlJaguarDevice::SetCcaMode(bool disabled) {
+  apply_cca(disabled, disabled);
+  _logger->info("Jaguar1: MAC carrier-sense {}",
+                disabled ? "DISABLED (dis_cca: CCA+EDCCA)"
+                         : "enabled (default)");
+}
+
+void RtlJaguarDevice::apply_cca(bool primary_disabled, bool edcca_disabled) {
   /* MAC carrier-sense gate: the same REG_TX_PTCL_CTRL bits as the HalMAC
    * generations — the vendor's phydm_mac_edcca_state drives 0x520[15] on
-   * this family too; [14] is the primary-CCA defer. */
+   * this family too; [14] is the primary-CCA defer. A set bit disables. */
   uint32_t v520 = _device.rtw_read<uint32_t>(0x0520);
-  if (disabled)
-    v520 |= (1u << 15) | (1u << 14);
-  else
-    v520 &= ~((1u << 15) | (1u << 14));
+  if (primary_disabled) v520 |= (1u << 14); else v520 &= ~(1u << 14);
+  if (edcca_disabled)   v520 |= (1u << 15); else v520 &= ~(1u << 15);
   _device.rtw_write<uint32_t>(0x0520, v520);
+  /* The BB threshold work below belongs to the EDCCA gate alone. */
+  const bool disabled = edcca_disabled;
 
   /* BB EDCCA thresholds (rEDCCA_Jaguar 0x8a4: L2H byte0 / H2L byte1). The
    * BB init table parks them at 0x7f/0x7f = never-trigger — the vendor's
@@ -1020,9 +1042,6 @@ void RtlJaguarDevice::SetCcaMode(bool disabled) {
    * threshold follows (vendor couples them per adaptivity cycle). */
   if (auto *wd = _halModule.phydm_watchdog())
     wd->SetEdccaTrack(!disabled);
-  _logger->info("Jaguar1: MAC carrier-sense {}",
-                disabled ? "DISABLED (dis_cca: CCA+EDCCA)"
-                         : "enabled (default)");
 }
 
 bool RtlJaguarDevice::SetAmpduMode(const devourer::AmpduMode &mode) {
--- a/src/jaguar1/RtlJaguarDevice.cpp
+++ b/src/jaguar1/RtlJaguarDevice.cpp
@@ -981,6 +981,12 @@ void RtlJaguarDevice::ClearAckResponder() {
 }
 
 bool RtlJaguarDevice::GetCcaGates(bool &primary_disabled, bool &edcca_disabled) {
+  /* The MAC register is meaningless before bring-up, and handing back
+   * whatever the bus returns would be a fabricated gate state. Jaguar3's
+   * SetCcaGates already guards on _brought_up; this is the read side and
+   * the Jaguar1 equivalent. */
+  if (!_brought_up)
+    return false;
   const uint32_t v = _device.rtw_read<uint32_t>(0x0520);
   primary_disabled = (v & (1u << 14)) != 0;
   edcca_disabled = (v & (1u << 15)) != 0;
@@ -988,6 +994,8 @@ bool RtlJaguarDevice::GetCcaGates(bool &primary_disabled, bool &edcca_disabled) 
 }
 
 bool RtlJaguarDevice::SetCcaGates(bool primary_disabled, bool edcca_disabled) {
+  if (!_brought_up)
+    return false;
   apply_cca(primary_disabled, edcca_disabled);
   _logger->info("Jaguar1: CCA gates primary={} edcca={}",
                 primary_disabled ? "OFF" : "on",
--- a/src/jaguar1/RtlJaguarDevice.cpp
+++ b/src/jaguar1/RtlJaguarDevice.cpp
@@ -981,10 +981,8 @@ void RtlJaguarDevice::ClearAckResponder() {
 }
 
 bool RtlJaguarDevice::GetCcaGates(bool &primary_disabled, bool &edcca_disabled) {
-  /* The MAC register is meaningless before bring-up, and handing back
-   * whatever the bus returns would be a fabricated gate state. Jaguar3's
-   * SetCcaGates already guards on _brought_up; this is the read side and
-   * the Jaguar1 equivalent. */
+  /* The MAC register is meaningless before bring-up, and reporting whatever
+   * the bus returns as the gate state would be a fabricated measurement. */
   if (!_brought_up)
     return false;
   const uint32_t v = _device.rtw_read<uint32_t>(0x0520);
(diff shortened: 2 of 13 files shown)


src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -1182,22 +1182,54 @@ RxEnergy RtlJaguar3Device::GetRxEnergy(bool with_nhm) {
...
+  /* DIS_EDCCA (energy) + DIS_CCA (carrier-sense); a set bit disables. */
+  if (primary_disabled) v520 |= (1u << 14); else v520 &= ~(1u << 14);
+  if (edcca_disabled)   v520 |= (1u << 15); else v520 &= ~(1u << 15);
+  /* 0x524[11] moves with the pair and only with the pair. Deliberate: the
+   * two pure states then write exactly the bytes the all-or-nothing path
+   * wrote before this split, so SetCcaMode is byte-identical. What this bit
+   * means on its own is not documented here and was not measured, so a
+   * mixed state leaves it at the enabled value rather than guessing. */
+  if (primary_disabled && edcca_disabled) v524 &= ~(1u << 11);
+  else                                    v524 |=  (1u << 11);
```
josephnef: **Blocking.** This bit is not undocumented — it is `BIT_EDCCA_MSK_CNTDOWN_EN`:

```
reference/rtl88x2bu/hal/halmac/halmac_bit_8822c.h:13368
#define BIT_EDCCA_MSK_CNTDOWN_EN_8822C BIT(11)
```

Same name on 8822B (`halmac_bit_8822b.h:11458`) and on the 8822E header, so it is a family-stable meaning, not an 8822C guess. It is EDCCA masking the backoff countdown — EDCCA-scoped, not pair-scoped — so it should follow `edcca_disabled`:

```c
if (edcca_disabled) v524 &= ~(1u << 11);
else                v524 |=  (1u << 11);
```

That still leaves `SetCcaMode`'s two pure states byte-identical (both gates equal ⇒ same write as before), so the no-behaviour-change property this PR is careful about survives. As written, the arm this PR recommends — EDCCA off, primary CCA on — still lets EDCCA mask the countdown on Jaguar3, so the gate the caller asked to turn off is only partly off.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -1182,22 +1182,54 @@ RxEnergy RtlJaguar3Device::GetRxEnergy(bool with_nhm) {
...
+
+bool RtlJaguar3Device::SetCcaGates(bool primary_disabled, bool edcca_disabled) {
+  std::lock_guard<std::mutex> lk(_reg_mu);
+  /* Sticky the same way dis_cca is: a channel set rewrites the BB CCA
+   * registers and SetMonitorChannel re-asserts from these. */
+  _cca_disabled = primary_disabled && edcca_disabled;
+  _cca_primary_disabled = primary_disabled;
+  _cca_edcca_disabled = edcca_disabled;
+  if (_brought_up)
+    apply_cca_gates_locked(primary_disabled, edcca_disabled);
```
josephnef: **Blocking.** The gate state now reaches phydm only through `_cca_disabled`, which this function sets to `primary_disabled && edcca_disabled`. The two `tick` call sites (`RtlJaguar3Device.cpp:244` in the coex thread and `:477` in the housekeeping path — outside this diff, so commenting here) pass `!_cca_disabled` as `edcca_track`, so in the EDCCA-off/primary-on arm the watchdog keeps running `PhydmRuntimeJaguar3::edcca()` and rewriting the BB thresholds at `0x84c` every ~2 s. The comment above `:477` ("EDCCA tracking is owned by `SetCcaMode` when the EDCCA-disable knob is active") stops being true at that point.

The Jaguar1 side of this same PR gets it right — `apply_cca` ends in `SetEdccaTrack(!edcca_disabled)`. Jaguar3 should pass `!_cca_edcca_disabled` for the same reason.

src/jaguar3/RtlJaguar3Device.cpp
```diff
@@ -1182,22 +1182,54 @@ RxEnergy RtlJaguar3Device::GetRxEnergy(bool with_nhm) {
...
+   * wrote before this split, so SetCcaMode is byte-identical. What this bit
+   * means on its own is not documented here and was not measured, so a
+   * mixed state leaves it at the enabled value rather than guessing. */
+  if (primary_disabled && edcca_disabled) v524 &= ~(1u << 11);
+  else                                    v524 |=  (1u << 11);
   _device.rtw_write<uint32_t>(0x0520, v520);
   _device.rtw_write<uint32_t>(0x0524, v524);
 }
 
+bool RtlJaguar3Device::GetCcaGates(bool &primary_disabled, bool &edcca_disabled) {
```
josephnef, line 1214: **Blocking.** No `_brought_up` guard here, while the Jaguar1 `GetCcaGates` added in the follow-up commit refuses before bring-up with exactly the right reasoning ("handing back whatever the bus returns would be a fabricated gate state"). The hazard is identical on Jaguar3 — this reads `0x520` on an un-brought-up chip and returns `true` — so the fix should be applied to both backends, not one.

Related asymmetry on the write side just below: Jaguar1 `SetCcaGates` returns `false` before bring-up and remembers nothing, while Jaguar3 records into `_cca_*` and returns `true`, applying at bring-up. Both behaviours are defensible, but a caller cannot distinguish them, and `IRtlRadio.h` documents `false` only as "not ported". Either align the two or document the second meaning.

src/IRtlRadio.h
```diff
@@ -85,6 +85,35 @@ class IRtlRadio : public IRadio {
...
+   * and they do not behave the same way — tests/dis_cca_tx_onair.sh measured
+   * primary CCA costing a Jaguar3 injector 41-45% against a co-channel
+   * flooder while the energy bit alone was null, and on Jaguar1 the result
+   * inverts (see below). A caller that needs one of them should not have to
+   * turn off both, and a caller diagnosing a deferral needs to tell them
+   * apart.
+   *
+   * `true` means DISABLED, matching SetCcaMode's argument sense and the
+   * register's own polarity (bit set = gate off). SetCcaMode is exactly
+   * SetCcaGates(d, d) and writes the same bytes it always did. Returns false
```
josephnef: Two doc points:

- `false` currently means both "this backend has not ported the split" and "this backend has ported it but is not brought up yet" (Jaguar1 returns `false` pre-bring-up; Jaguar3 returns `true` and defers). Worth stating, since a caller that probes capability at construction gets the wrong answer on Jaguar1.
- Stickiness is unspecified and differs: Jaguar3 tracks the pair and re-asserts in `SetMonitorChannel`, Jaguar1 records nothing, and Jaguar1's bring-up unconditionally re-runs `SetCcaMode(_cfg.tuning.disable_cca)` (`RtlJaguarDevice.cpp:128` and `:1611`), so anything set before `Init` is overwritten. Given `FastRetune`/`SetMonitorChannel` are the normal operating pattern for this lever, the contract belongs at the declaration.

Smaller: "on this family" reads oddly in a base class serving five Realtek backends — name Jaguar1/Jaguar3. And the `41-45%` here and the `~1.5-2.2x` in CLAUDE.md appear to describe the same `tests/dis_cca_tx_onair.sh` result; make the two agree or cite the run once.

src/jaguar1/RtlJaguarDevice.cpp
```diff
@@ -980,16 +980,46 @@ void RtlJaguarDevice::ClearAckResponder() {
...
   uint32_t v520 = _device.rtw_read<uint32_t>(0x0520);
-  if (disabled)
-    v520 |= (1u << 15) | (1u << 14);
-  else
-    v520 &= ~((1u << 15) | (1u << 14));
+  if (primary_disabled) v520 |= (1u << 14); else v520 &= ~(1u << 14);
+  if (edcca_disabled)   v520 |= (1u << 15); else v520 &= ~(1u << 15);
   _device.rtw_write<uint32_t>(0x0520, v520);
+  /* The BB threshold work below belongs to the EDCCA gate alone. */
+  const bool disabled = edcca_disabled;
```
josephnef: Nit: re-aliasing to a local named `disabled` keeps the diff small, but it leaves a variable whose name no longer says which gate it is, in a function that now has two. Renaming the four uses below to `edcca_disabled` is a mechanical change and makes the EDCCA-only scoping self-evident at each site instead of resting on this one line.

Also in `apply_cca` above: the `if (primary_disabled) v520 |= ...; else v520 &= ...;` one-liners are the only braceless same-line `if/else` pairs in the file.

josephnef (changes requested):
Verified locally on the head: builds clean, `ctest` 63/63. The API shape is right — not-ported defaults, `SetCcaMode` reduced to `SetCcaGates(d, d)`, and the Jaguar1 byte-for-byte claim holds (`apply_cca(d, d)` is the old body with the BB `0x8a4` half keyed on the EDCCA argument). The Jaguar1 measurement is the valuable part of this PR and the inversion is worth acting on.

Three blocking items, all on Jaguar3, all invisible to a Jaguar1-only bench:

1. `0x524[11]` is not undocumented — it is `BIT_EDCCA_MSK_CNTDOWN_EN` and it is EDCCA-scoped, so it should follow `edcca_disabled` rather than the pair.
2. phydm's `edcca_track` argument is still keyed on the all-or-nothing `_cca_disabled`, so the watchdog keeps tracking EDCCA in a mixed state.

Together those two mean the arm this PR recommends — EDCCA off, primary CCA on — is only half applied on Jaguar3. The on-air evidence here is all Jaguar1, where the path is correct, so the bench could not have caught it.

3. The pre-bring-up contract diverges between the two backends, and `GetCcaGates` on Jaguar3 has no `_brought_up` guard — the exact fabricated-gate-state hazard the Jaguar1 follow-up commit fixed.

Non-blocking: stickiness across `SetMonitorChannel` differs between the two backends and is undocumented (and Jaguar1 bring-up unconditionally re-runs `SetCcaMode(_cfg.tuning.disable_cca)`, clobbering anything set before `Init`); there is no in-tree caller, env knob or regcheck script, so nothing in the repo reproduces the tables in the description — the `tests/txpwr_*_regcheck.sh` shape would cover the four gate states plus retune survival.

On the CLAUDE.md question: please scope it in this PR rather than a follow-up. The current text states the Jaguar3 result as general and it is now measured inverted on Jaguar1, and the new API is not mentioned in Configuration at all. Pair the Jaguar1 table with the flooder arm (95% → 78%, both-off 0.3%) in the same section — turning both gates off being worse for the injector is the half that keeps the first half from being read as "disable everything".

On Jaguar1's bring-up enabling EDCCA off its parked default with a hard-coded `th_l2h_ini`: agreed it is a policy question and out of scope here, but the 94% makes it a real one — worth an issue.

snokvist:
Thanks — both blocking Jaguar3 items were real, and you were right that the
bench here could not have caught them: every on-air number in the original
description came from a Jaguar1 DUT. An 8812CU is now attached, so all three
are fixed and measured rather than argued.

**1. `0x524[11]` is `BIT_EDCCA_MSK_CNTDOWN_EN`, and moving it with the pair was
wrong.** Confirmed against the vendor headers independently of your citation —
`BIT_EDCCA_MSK_CNTDOWN_EN_8822C BIT(11)` at `halmac_bit_8822c.h:13368`, the same
name at `8822b:11458` and `8822e:13531`, and `REG_RD_CTRL_8822C 0x0524`. It now
follows `edcca_disabled` alone. Measured on the 8812CU, reading the chip in
every state:

| state | `0x520[14]` | `0x520[15]` | `0x524[11]` before | after |
|---|---|---|---|---|
| primary on, EDCCA on | 0 | 0 | 1 | 1 |
| **primary on, EDCCA off** | 0 | 1 | **1** | **0** |
| primary off, EDCCA on | 1 | 0 | 1 | 1 |
| primary off, EDCCA off | 1 | 1 | 0 | 0 |
| `SetCcaMode(true)` | 1 | 1 | 0 | 0 |
| `SetCcaMode(false)` | 0 | 0 | 1 | 1 |

Exactly one row moves and it is the recommended arm, so the byte-identical
property for `SetCcaMode` survives — as you predicted.

My original comment called the bit undocumented. That was the wrong kind of
caution: I treated "I have not measured it" as a reason to guess a conservative
coupling rather than as a reason to go and look. It was named in the tree
already — `RtlJaguar2Device.cpp:1932` calls it `EDCCA_MSK_COUNTDOWN`.

**2. `edcca_track` now follows the EDCCA gate.** Both `tick` sites pass
`!_cca_edcca_disabled`, and the stale comment at the housekeeping site is
rewritten. Worth recording how this has to be measured: sampling `0x84c` cannot
see the bug, because `PhydmRuntimeJaguar3::edcca()` recomputes the same
`th_l2h` from a static IGI — an active tracker writes identical bytes and looks
exactly like an idle one. Poking the register with a value the tracker would
never choose and watching whether it is restored does discriminate:

| arm | before | after |
|---|---|---|
| both gates on (default) | restored — tracking | restored — tracking |
| **EDCCA off, primary CCA on** | **restored — tracking** | **survived — stopped** |
| both gates off | survived — stopped | survived — stopped |

**3. Pre-bring-up contract aligned, not documented-as-divergent.** Jaguar3's
`GetCcaGates` and `SetCcaGates` now refuse before bring-up exactly as Jaguar1
does, caching nothing. Verified on both: the call returns `false` and
`GetCcaGates` leaves the caller's out-parameters untouched. This also settles
qodo's finding on the same function — it reported success for a write that
never happened, which is the same defect seen from the other side.

Worth noting what the unguarded read actually did, since it is less dramatic
than "returns garbage" and more awkward: on this 8812CU it returned a
*plausible* state, and which state depended on what the previous session had
left on the chip rather than on anything the caller had set. Reading a
freshly-reset part gave bits 14 and 15 clear — indistinguishable from a
correctly-brought-up default. A fabricated reading that looks right is the
version of this that survives review, which is the argument for refusing
rather than for sanitising.

---

Non-blocking items:

**Stickiness — your assertion needs a correction, and it is in my favour, so
please check me.** I wrote "Jaguar1 records nothing so a retune drops it" into
the contract on your word, then measured it and it is not what the hardware
does. With EDCCA off and primary CCA on, an 8812AU keeps `0x520[15]` set and
its BB thresholds parked at `7f/7f` across a same-band retune *and* across a
5 GHz→2.4 GHz band change; an 8822C does the same. The mechanisms differ
exactly as you said — Jaguar3 re-asserts in `SetMonitorChannel`, Jaguar1 has no
re-assert at all — but Jaguar1 survives because nothing in its channel path
rewrites those registers. `IRtlRadio.h` now states the measured outcome and
warns that the Jaguar1 case is incidental rather than guaranteed. Bring-up *is*
a reset there, via the `SetCcaMode(_cfg.tuning.disable_cca)` calls you cited.

**In-tree caller and regcheck.** `tests/cca_gates_probe.cpp` (built as
`CcaGatesProbe`) is the caller; `tests/cca_gates_regcheck.sh` is the
register-level check in the shape you pointed at. It covers the four gate
states, the pre-bring-up refusal, the two legacy `SetCcaMode` states, the
`0x524[11]` scoping and retune survival, cross-checking the API's readback
against a raw `chipstate --no-claim` peek so an API that lies about the silicon
fails rather than passes. Currently 23 passed / 0 failed / 1 skipped across the
8812AU, 8812CU and an RTL8733BU; reverting either of your two fixes makes it
fail on the corresponding cell, which is how I checked the cells can fail at
all rather than assuming it. One limit worth stating: the `cntdown` cell
discovers whether a backend drives `0x524[11]` by comparing the two
`SetCcaMode` states against the split, so it catches the two paths
disagreeing, but a backend that stopped writing the bit altogether would be
reported rather than failed — deciding it *should* move would need a per-chip
expectation table. The `track` cell covers the functional half of that arm
independently.

I did **not** add an env knob — that is new public configuration
surface and this PR is already under review; say the word if you would rather
have `DEVOURER_DIS_PRIMARY_CCA` / `DEVOURER_DIS_EDCCA` alongside the existing
`DEVOURER_DIS_CCA` and I will add it.

**`CLAUDE.md`** is scoped into this PR. The Jaguar3 result is no longer stated
as general; both families' results are given as disagreeing measurements of
different things, the flooder arm is paired in (95% idle → 78% flooded with
primary CCA kept on, against 0.3% with both gates off), and `SetCcaGates` is
named in Configuration. On the three numbers: I removed the restated ones and
cite the runs once each, since `~40-60%`/`41-45%`/`~1.5-2.2x` were three
roundings of one run.

**`IRtlRadio.h`** now names Jaguar1/Jaguar3 instead of "this family", and
states the two meanings of `false` plus the out-parameter rule. It also
documents that `SetCcaMode` does *not* share the contract — it returns void,
so a pre-bring-up call cannot report anything — and points at
`tuning.disable_cca` as the from-bring-up path.

**Jaguar1 nits** applied: the `disabled` alias is gone and the four uses read
`edcca_disabled`, and the braceless same-line `if/else` pairs are braced.

Also removed: `_cca_disabled` on Jaguar3, which this PR left write-only once
both readers moved to the per-gate flags.

**The EDCCA-at-bring-up policy question** — Jaguar1 enabling EDCCA off its
parked `0x7f7f` default with a hard-coded `th_l2h_ini`, where the vendor ships
`CONFIG_RTW_ADAPTIVITY_EN 0` and exposes the thresholds as module parameters —
I will raise separately as an issue rather than widen this PR.

Devices actually verified for the above: RTL8812AU (Jaguar1), RTL8822C
(Jaguar3) and RTL8733BU (as an unported `IRtlRadio`, confirming the
not-ported defaults refuse on real silicon). The 8822EU/8812EU arms of the
original tables are unchanged and not re-run — I do not have those parts.

Pushed as `68cb3f8`. Build is clean and `ctest` 63/63 from a fresh configure of
the branch.

josephnef (dismissed):
All three blocking items are fixed, and I checked the mechanism rather than the claim. `0x524[11]` now follows `edcca_disabled` alone; both `_phydm.tick` sites pass `!_cca_edcca_disabled`; Jaguar3's `SetCcaGates`/`GetCcaGates` refuse before bring-up and the combined `_cca_disabled` is gone with no readers left behind. `_brought_up = true` precedes `SetCcaMode(_cfg.tuning.disable_cca)` on both families, so the config path still applies. Build clean, `ctest` 63/63.

I ran `tests/cca_gates_regcheck.sh` on my own bench, and it picked up an arm you said you could not cover:

```
0xa81a  RtlJaguar3Device, chip-id 0x17 — 8822E     12 passed
0xc812  RtlJaguar3Device, chip-id 0x13 — 8822C     12 passed
0xb812  Jaguar2 8822BU        not ported, refuses cleanly
0xb733  RTL8733BU             not ported, refuses cleanly
35bc:0101  Kestrel C8852C     not ported, refuses cleanly   (rc=5)

== 25 passed, 0 failed, 0 skipped ==
```

`cntdown` reads `00:1 01:0 10:1 11:0` on both Jaguar3 dies — so `0x524[11]` follows EDCCA alone on the **8822E** as well as the 8822C, and the two `SetCcaMode` states still bracket it. The `track` cell passes on both. I have no Jaguar1 plugged, so that family is still only your bench.

The correction on stickiness is a good catch and I was wrong: I asserted a mechanism where you measured an outcome. The header's wording — survives on both, means it on only one, don't build on the Jaguar1 case — is the right shape.

No blocking items. Seven things worth a pass before merge, none of which need another review round.

**1. Jaguar1 `apply_cca` clears the tracker last, so a mid-session EDCCA-disable can race the watchdog.** `src/jaguar1/RtlJaguarDevice.cpp` parks `0x8a4` at `0x7f7f` and *then* calls `SetEdccaTrack(false)`. `PhydmWatchdog` is started in `rtw_hal_init`, i.e. before bring-up's `SetCcaMode`, and its tick rewrites both `0x8a4` bytes from the IGI-derived L2H while `_edcca_track` is still set. A tick landing in that window leaves live thresholds behind the disable the caller just asked for — the same defect as round-one finding #2, reached through a window instead of a wrong argument. It is inert at bring-up and gated on `DEVOURER_PHYDM_WATCHDOG=1`, but `SetCcaGates` is the first API that makes this a runtime call. Hoisting `SetEdccaTrack(false)` above the park (and leaving `SetEdccaTrack(true)` after the enable programming) closes it.

Related: Jaguar1's `SetCcaGates`/`GetCcaGates` take no lock while Jaguar3's take `_reg_mu`, and the `0x520`/`0x8a4` read-modify-writes share the bus with that same watchdog thread.

**2. The `track` cell hard-codes `0x84c`, which is Jaguar3's register — Jaguar1's is `0x8a4`.** On a Jaguar1 DUT the cell pokes an unrelated BB register, sees no restore, and prints `no EDCCA tracker running in the default arm`, which is false: `PhydmWatchdog::SetEdccaTrack` drives `0x8a4`. The EDCCA-off arm then reports `SKIP: no tracker to stop in this configuration`. I suspect that is the `1 skipped` in your 23/0/1, which would mean the functional half of fix #2 is unverified on the family whose measurement motivated the PR. Keying the address off the family turns that skip into a real cell — and a harness that quietly applies the Jaguar3 register to Jaguar1 is the exact generalisation this PR is about not making.

**3. `tests/cca_gates_regcheck.sh:221` — `elif true; then` makes the `else` at 231-233 unreachable**, and that dead `note` duplicates line 217 verbatim. Plain `else`.

**4. The `cntdown` cell disappears silently when fewer than four arms report.** The `[ ... -eq 4 ]` at line 204 has no `else`, so a single failed arm drops the whole cell with no PASS/FAIL/SKIP and the counters simply do not move. A `skip` in the else keeps the ledger honest about what was not measured.

**5. `stop_hold` cannot stop the probe.** `start_hold` backgrounds `sudo -n "$BUILD/CcaGatesProbe"`, so `$!` is a root-owned process; run as documented (`sudo -v && tests/cca_gates_regcheck.sh`) the plain `kill "$probe_pid"` gets EPERM and is swallowed by `2>/dev/null`, and `wait` then blocks for the probe's entire hold walk. That is why a run is minutes per adapter — the two Jaguar3 cells took ~25 minutes here. `sudo -n kill` fixes it. The EXIT trap is fine, its `pkill` is already under `sudo -n`.

**6. The retune cells can only see half the gate state.** `GetCcaGates` reads `0x520` alone, so `retune`/`fast-retune` cannot detect loss of `0x524[11]` or of the BB thresholds. `IRtlRadio.h` claims survival "at `0x520`, `0x524` and Jaguar1's BB thresholds"; `peek32` is already there, so one extra read after the retune would put the `0x524` half of that claim in the harness instead of in the prose.

**7. Two smaller ones.** `src/jaguar2/RtlJaguar2Device.cpp:1932` still says "The primary-CCA bit is the one that stops TX deferring to a co-channel transmitter" on a family with no measurement of its own — the last unscoped copy of the claim you just scoped in `CLAUDE.md`, one file over. And in `CLAUDE.md` the new text now runs straight into **On by default on the streamtx FPV downlink**, which lands immediately after the `SetCcaGates` sentence and reads as though it describes the new API rather than `DEVOURER_DIS_CCA`; re-anchoring it on the env var restores the referent.

On the env knob: I would leave it out, as you did. `DEVOURER_DIS_CCA` covers the shipped default and the split has an in-tree caller now; adding two more vars before anything in the tree wants them is surface for its own sake.

Scoping the Jaguar3 dis_cca figure as the DUT's host-side `submitted` rate, next to the Jaguar1 numbers that are decoded delivery from two witnesses, is the part of this description I would keep verbatim — the two results only stop looking contradictory once the metrics are named.

Approving. The list above is all follow-up-sized; #1 and #2 are the two I would do before merge, since both are the same bug class the first round found.

josephnef:
Pushed `2848c5a` applying the review findings myself, so this can go in without another round.

The first one turned out to be a real defect, not a nit. Jaguar1 `apply_cca` cleared the EDCCA tracker *after* parking `0x8a4`, and `PhydmWatchdog` owns that register while tracking — running on its own thread from `rtw_hal_init`, i.e. before bring-up's `SetCcaMode`. A tick landing between the park write and the flag store re-derived L2H from IGI and overwrote the park, leaving live thresholds behind a disable the caller had asked for. Same failure as the `edcca_track` one, reached through a window instead of a wrong argument, and inert until `SetCcaGates` made the gates settable mid-session.

Reordering alone only narrows the window, so `SetEdccaTrack` is now synchronous — the tick's EDCCA block and the flag store share a mutex, and turning tracking off drops the write-on-change cache too, since the register no longer holds the value `_edcca_last_l2h` names. `apply_cca` stops the tracker before it writes and hands the register back only after programming it.

Your harness could not have caught that, and the reason is worth stating: the `track` cell poked `0x84c` on every family. That is Jaguar3's threshold register — Jaguar1's is `0x8a4`. On a Jaguar1 DUT it poked something unrelated, saw no restore, and reported `no EDCCA tracker running in the default arm`. That is almost certainly your one skipped cell, and it means the functional half was unverified on the family whose measurement motivated the split. The probe now reports its generation from `AdapterCaps` and the cell takes the register from that, skipping a generation it has no address for rather than guessing.

The retune cells got the same treatment. They only ever saw `0x520`, because that is all `GetCcaGates` reads, so they now peek `0x524[11]` while the state is held. I negative-controlled it on the 8812CU with the countdown write deliberately broken:

```
GATES after-retune   ret=1 primary=1 edcca=0     <- API says intact
0x524[11] = 0                                    <- new cell fails
```

Which is the point: the API cannot speak for that half.

Rest as described — `stop_hold` now stops through sudo (a plain `kill` on a sudo-launched probe was getting EPERM silently, which is why a run took ~25 min an adapter here), the dead `elif true` branch is gone, and the cntdown cell reports a SKIP instead of vanishing without a verdict. Plus the two scoping fixes: `RtlJaguar2Device.cpp` was the last unscoped copy of the primary-CCA claim, and `CLAUDE.md`'s "on by default on the streamtx FPV downlink" had drifted to where it read as describing `SetCcaGates` rather than `DEVOURER_DIS_CCA`.

Verified: build clean, `ctest` 63/63, jaguar1-only config builds, and the regcheck is **26 passed / 0 failed / 0 skipped** over an 8822E, an 8822C, an 8822BU and an RTL8733BU. The skip is gone because the cell now names its register:

```
note: 0xc812 track: tracker IS running at 0x84c in the default arm (as expected)
```

Still no Jaguar1 on this bench, so that family's on-air table remains yours. Given the tracker fix lands on exactly that path, a re-run of the regcheck against your 8812AU would be worth doing — the `track` cell should now produce a real verdict there instead of a skip.

Merging.

src/jaguar3/RtlJaguar3Device.cpp
josephnef: Fixed in 68cb3f8: `0x524[11]` follows `edcca_disabled` alone. Verified on hardware — `cntdown` reads `00:1 01:0 10:1 11:0` on both an 8822C and an 8822E, so the bit is EDCCA-scoped on both Jaguar3 dies and `SetCcaMode`'s two pure states still bracket it.

src/jaguar3/RtlJaguar3Device.cpp
josephnef: Fixed in 68cb3f8: both `_phydm.tick` sites pass `!_cca_edcca_disabled`, and the combined `_cca_disabled` is deleted with no readers left behind. The `track` cell confirms the tracker stops in the EDCCA-off arm on both Jaguar3 dies — and in 2848c5a that cell picks its register per family, since it had been poking Jaguar3's `0x84c` on Jaguar1 too.

src/jaguar3/RtlJaguar3Device.cpp
josephnef: Fixed in 68cb3f8: both `SetCcaGates` and `GetCcaGates` refuse before `_brought_up`, caching nothing. Verified that `_brought_up = true` still precedes `SetCcaMode(_cfg.tuning.disable_cca)` in both `Init` and `InitWrite`, so the config path is unaffected.

src/IRtlRadio.h
josephnef: Addressed in 68cb3f8: the header now states both meanings of `false`, the out-parameter rule on a refusal, and that `SetCcaMode` does not share the contract because it returns void. The stickiness paragraph is the author's measured version, which corrected me — it survives a retune on both families but only Jaguar3 means it.

src/jaguar1/RtlJaguarDevice.cpp
josephnef: Applied in 68cb3f8: the `disabled` alias is gone and all four uses read `edcca_disabled`. That function changed again in 2848c5a — it now stops the phydm tracker before parking `0x8a4` rather than after, which was a real race against the watchdog thread.

PR #430, opened 2026-09-14
IRadio: WriteTsf reports success, and the MT7612U reports false

snokvist:
> This is a standalone change; it does not stack on the mt7612u series. It
> touches the shared `IRadio` contract and adds an `AdapterCaps` field, so it is
> best reviewed knowing it is also an (intentional) API change.

`IRadio::WriteTsf` was `void` with a silent no-op default, so a caller could not
tell a chip that moves its TSF from one that discards the write. The MT7612U was
that case: we wrote `MT_TSF_TIMER_DW0`/`DW1` and the registers hold a
free-running counter that does not load.

This PR splits the question in two:

- **Static:** whether a part has a standalone TSF write at all is
  `AdapterCaps::tsf_write_ok`, resolved at construction. It is true on
  Jaguar1/2/3 and false on the MT7612U (measured), Kestrel and the RTL8733B.
- **Per call:** `WriteTsf` returns `bool`, whether the transport accepted this
  write. On PCIe, and on USB near a low-word wrap, that answer comes from a
  readback.

A `bringup tsfwrite` cell characterises the MT7612U's absence of a load path, so
a future firmware that enables loading trips it.

## It is ignored, on hardware, in both bands, on two units

Units 8-1 (`40:a5:ef:50:27:a1`) and 5-1 (`40:a5:ef:5a:32:f8`), channels 6 and
149, firmware running, `MT_BEACON_TIME_CFG` free-running. All four runs PASS.
8-1, channel 149:

```
MT_BEACON_TIME_CFG=0x00070640 TIMER_EN=1 TBTT_EN=0 BEACON_TX=0 SYNC_MODE=3
  clock control                      t0=430488 t1=480952 delta=+50464  ALIVE
  MAC on, both words DW0,DW1         target=5481626 read=502027 err=-4979599 delta50ms=+50341  no-op
  MAC on, both words DW1,DW0         target=5552601 read=573019 err=-4979582 delta50ms=+50298  no-op
  MAC on, high word (DW1) only       target=4295590758 read=643763 err=-4294946995 delta50ms=+50286  no-op
  MAC on, low word (DW0) only        target=5694281 read=714614 err=-4979667 delta50ms=+50422  no-op
  MAC off, both words                target=5778837 read=799192 err=-4979645 delta50ms=+50378  no-op
                                     held=0 (read with TIMER_EN clear)
  MAC off, timer off, both words     target=5849794 read=20232 err=-5829562 delta50ms=+50366  no-op

GATE TSF-WRITE: PASS - confirmed: no sequence loads the TSF, so tsf_write_ok is false
```

How the gate avoids a false PASS:

- **Positive control.** The gate only PASSes while the clock is confirmed alive,
  so a frozen counter cannot be mistaken for a successful load. Every arm also
  re-checks liveness.
- **Checked reads.** Every TSF read is checked, so a failed transfer is reported
  as a transport fault, not a dead clock.
- **Wrap guard.** Each arm takes its base clear of a low-word wrap, so the
  single-word arms can't misjudge a load across a carry.
- **Host-time bounds.** A silently retried EP0 transfer can't turn a late
  readback of a real load into a "no-op": a slow arm is inconclusive and FAILs.
- **Non-vacuous high-word arm.** Its target changes the high word
  (`base + (1ull<<32)`).
- **Timer-off arm.** Restoring `TIMER_EN` restarts the counter from ~0, which
  would wipe a load before the normal readback. So this arm also reads the
  counter while the timer is still clear. It reads 0, not the target.

**And the Realtek parts accept it.** Every recorded readback of a +5 s write is
in `docs/time-distribution.md`: the source, the sleep, and error − sleep per row.
In summary:

| die | readbacks | error − sleep |
|---|---|---|
| RTL8822B (incl. RTL8812BU, this PR's `WriteTsf`, host-timed) | 45 | +98 to +3419 µs (timed rows: +98 to +881 µs) |
| RTL8822C | 2 | +274 / +1013 µs |
| RTL8821AU (raw pair, no beacon) | 2 | +625 / +1000 µs |

On the RTL8812BU, 40 host-timed writes (0/20/50 ms sleeps, plus backward −5 s
writes) all returned true and landed, and readback error minus the host time
from `WriteTsf` returning to `ReadTsf` returning was −485…+38 µs: the counter
loads the target and the residual is host time. The 8821AU cells are the maintainer's scratch probe on the register pair the J1
override writes; the override itself has not run on J1 hardware. The
8812AU/8814AU, 8821C (USB and CE) and 8822E share the pair and code path and are
not separately measured.

## Structure

- **`IRadio::WriteTsf` becomes `bool`**, and the default returns false.
  - `AdapterCaps::tsf_write_ok` (emitted as `tsf_write` in `adapter.caps`)
    answers "does this part have one".
  - On a `tsf_write_ok` part, false is a transport failure, and the counter may
    be half-updated: read back, and retry only while the device is present.
- **`src/RtlTsf.h`** holds `devourer::read_tsftr` / `write_tsftr`, shared by
  Jaguar1/2/3. Both words are always attempted.
  - **USB:** true is two completed control transfers, except when the target's
    low word is within 100 ms of wrapping. There the pair can tear between the
    transfers, so it is read back.
  - **PCIe:** a register write is a posted MMIO store that always reports
    success, so every write is read back. A device that has left the bus
    typically reads all-ones and fails.
  - A throwing USB readback folds into false.
- **Jaguar1** writes the bare pair under `_port0_mu`, so it cannot interleave
  with a beacon steer. On J1 the TBTT is recorded as locked to the TSF grid, so
  expect an active beacon's TBTT to move with a write. That, and whether the
  beacon keeps airing without the steer's re-download, is unmeasured through
  `WriteTsf`.
- **Jaguar2/3** keep `_reg_mu` and route through the helper.
- **MT7612U** reports `tsf_write_ok = false` and does not override `WriteTsf`,
  so the `IRadio` default refuses without touching a register.
  `mt7612u_write_tsf` (its declaration, definition and `api_link` entry) is
  deleted; a helper that only ever no-ops is a trap, not an API.
- **`bringup tsfwrite [chan]`** is the characterisation cell, next to `caps`.

## What this does not show

- It does **not** move the beacon TBTT air-time on Jaguar2/3. `WriteTsf` moves
  the reported TSF and the beacon timestamp; the TBTT is a separate per-port
  timer there. `PinBeaconTbtt` / `AdjustBeaconTimingFine` are the TBTT levers.
- Kestrel and the RTL8733B report false because their source has no TSF write
  at all. That is a source fact, not a bench measurement.
- Not hardware-exercised:
  - the PCIe readback (no 8821CE here);
  - the J1 override on J1 silicon;
  - the J3 helper refactor (the bench 8812CU was busy; the register operations
    are unchanged). The J2 helper did run, on an RTL8812BU;
  - the gate's wrap-guard, time-bound and read-failure branches.

## What review changed, because it is most of the value here

- **Adversarial rounds 1–2 and Qodo (commits dda2e71 … 9280d29):**
  - clock positive control;
  - both word orders, and a non-vacuous high-word arm;
  - a post-sweep I/O guard;
  - per-arm liveness;
  - the J2/J3 `lo_ok && hi_ok`;
  - the checked `MT_BEACON_TIME_CFG` read in the timer-off arm;
  - `mt_mac_stop` on a failed start.
- **Maintainer review (45cf642):**
  - **Jaguar1's explicit false was measured wrong.** The raw pair loads on an
    8821AU, so J1 now writes.
  - **"true means both transfers landed" was false over PCIe**, so the shared
    helper reads back.
  - **false conflated "unsupported" with "transient"**, so the static half moved
    to `AdapterCaps::tsf_write_ok`.
  - **Gate fixes:** a low-word wrap could false-PASS the DW0-only arm, and the
    clock control used unchecked reads.
  - **The redundant MT7612U override** was dropped.
  - **Nits:** an enum plus an arm table instead of six copy-pasted blocks; one
    named liveness window; the readback claim now carries its sleep term and n.
- **A further adversarial pass on that response:**
  - the table had dropped the worst 8822B cell (+53419 µs);
  - the "two units" claim had no log;
  - the timer-off arm could never see a take;
  - silent EP0 retries could stretch an arm past both the take window and the
    wrap guard;
  - the USB pair can tear across a low-word carry;
  - a readback that throws broke the bool contract.

  All six are fixed above.

## API/surface breaks (intentional)

- **`mt7612u_write_tsf` removed** from the public header, its definition, and
  the `api_link` check. This is the hard break: it is a C symbol, so any
  out-of-tree user fails to link. `api_link` is built only by
  `src/mt7612u/Makefile`; CI does not build it.
- **`IRadio::WriteTsf`: `void` → `bool`.** Return types are not part of the
  Itanium mangled name, so this breaks source compatibility for out-of-tree
  callers, not the ABI.
- **`AdapterCaps::tsf_write_ok` added**, and `adapter.caps` gains `tsf_write`.
  An `AdapterCaps` layout change breaks ABI for anyone compiled against the old
  struct, as every new caps flag does.

There are no in-tree callers of `WriteTsf` or `mt7612u_write_tsf`.

## Test plan

```sh
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DDEVOURER_MT7612U=ON
cmake --build build -j
ctest --test-dir build                    # 63/63, radio_iface incl. tsf_write_ok default
make -C src/mt7612u check                 # api_link (30), frame_shape, field_macros, log_sink
# hardware (firmware from linux-firmware: mt7662_rom_patch.bin, mt7662.bin)
make -C src/mt7612u bringup
MT7612U_DEV=<bus-port> src/mt7612u/bringup tsfwrite 6     # expect GATE TSF-WRITE: PASS
MT7612U_DEV=<bus-port> src/mt7612u/bringup tsfwrite 149   # 5 GHz, same
```


src/jaguar2/RtlJaguar2Device.cpp
```diff
@@ -1960,12 +1960,15 @@ uint64_t RtlJaguar2Device::ReadTsf() {
...
+bool RtlJaguar2Device::WriteTsf(uint64_t tsf) {
   /* REG_TSFTR 0x0560 (low) / 0x0564 (high). Serialized on _reg_mu against the
-   * coex/thermal tick. The counter keeps running, so this sets it to ~tsf. */
+   * coex/thermal tick. The counter keeps running, so this sets it to ~tsf.
+   * Same register pair as Jaguar3; both are bench-proven to move the reported
+   * TSF (RTL8822B readback: the target plus the control round trip). */
   std::lock_guard<std::mutex> lk(_reg_mu);
   _device.rtw_write<uint32_t>(0x0560, static_cast<uint32_t>(tsf));
   _device.rtw_write<uint32_t>(0x0564, static_cast<uint32_t>(tsf >> 32));
+  return true;
```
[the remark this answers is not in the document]
snokvist: Fixed in 9280d29: both `WriteTsf` overrides now capture each `rtw_write` result and return `lo_ok && hi_ok`. Both writes are still always issued (the pair is the unit; a half-written TSF is worse than a reported failure) — the `&&` is applied to the captured bools after both calls, so it cannot short-circuit the second write. The comment now says false means at least one transfer did not land. Re-verified on the bench: Jaguar2 (RTL8822B) `took:true` (delta +110 us on a +5 s write) and Jaguar3 (RTL8822C) `took:true` (delta +1013 us).

src/mt7612u/tools/bringup.cpp
```diff
@@ -2863,6 +2863,145 @@ static int gate_rtap(uint8_t chan, int count)
...
+static bool tsf_variant(const char *label, uint64_t target, bool stop_timer,
+                        bool high_word_first, int which)
+{
+	uint32_t cfg = mt_rr(&dev, MT_BEACON_TIME_CFG);
+	uint64_t r1, r2;
+	int64_t err, rate;
+	bool took;
+
+	if (stop_timer)
+		mt_wr(&dev, MT_BEACON_TIME_CFG, cfg & ~MT_BEACON_TIME_CFG_TIMER_EN);
```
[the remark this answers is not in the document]
snokvist: Fixed in 9280d29: `tsf_variant` now reads `MT_BEACON_TIME_CFG` with `mt_rr_chk`; if that read fails the timer-off arm is skipped entirely (no stop/restore writes), so the 0xffffffff sentinel can never be written back into the beacon timer register. The accumulated `io_err` still fails the sweep at its boundary. Rebuilt and re-run on hardware in both bands: gate PASS, no I/O errors.

src/mt7612u/tools/bringup.cpp
```diff
@@ -2863,6 +2863,145 @@ static int gate_rtap(uint8_t chan, int count)
...
+		return 1;
+	}
+
+	/* With the MAC running (the state a live link is in): both word orders,
+	 * then each word alone. The high-word-only target sets a high word that
+	 * differs from the live one, so the arm is not vacuous. */
+	if (mt_mac_start(&dev, MT_RX_DRAIN_NONE)) {
+		printf("\nGATE TSF-WRITE: FAIL - mt_mac_start failed\n");
+		return 1;
+	}
```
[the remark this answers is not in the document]
snokvist: Fixed in 9280d29: `gate_tsfwrite` calls `mt_mac_stop(&dev)` before returning from the `mt_mac_start` failure branch, so a failed start cannot leave transmitter control enabled. The gate’s own `MT_BEACON_TIME_CFG` read is now checked too (explicit FAIL verdict on a failed read). Both bands re-run green after the change.

src/jaguar1/RtlJaguarDevice.h
```diff
@@ -320,6 +320,15 @@ class RtlJaguarDevice : public IRtlRadio {
...
   uint64_t ReadTsf() override;
+  /* No standalone TSF write. On this generation the TSF is moved only by the
+   * complete beacon-steer sequence (see PinBeaconTbtt/AdjustBeaconTimingFine
+   * below), which shifts the reported TSF and the TBTT grid together; a bare
+   * REG_TSFTR write was never shown to take on its own. This method
+   * deliberately does not perform that sequence's EN_BCN_FUNCTION toggle and
+   * TBTT re-download: doing it here would move the beacon grid, which is a
+   * different operation than setting the counter. Reporting false keeps
+   * "unsupported here" honest rather than inheriting a silent no-op. */
+  bool WriteTsf(uint64_t tsf) override { (void)tsf; return false; }
```
josephnef, line 331: **Blocking — this is measured wrong on hardware.** The comment says a bare REG_TSFTR write "was never shown to take on its own". I measured it on an RTL8821AU (TP-Link Archer T2U Plus, 2357:0120), brought up with `InitWrite` on ch36, no beacon armed (`REG_BCN_CTRL` 0x550 = 0x10, EN_BCN_FUNCTION = 0), with the same positive control and per-arm liveness check your MT7612U gate uses:

| arm | RTL8821AU (Jaguar1) | RTL8822B (reference) |
|---|---|---|
| clock control, 50 ms | +51249 µs, alive | +50256 µs, alive |
| bare write DW0 then DW1 | **TOOK**, err +21000 µs | TOOK, +20196 µs |
| bare write DW1 then DW0 | **TOOK**, err +20625 µs | TOOK, +20173 µs |
| write bracketed by the EN_BCN_FUNCTION toggle | TOOK, +21364 µs | TOOK, +20220 µs |
| this PR's `WriteTsf` | false, err −4.95e6 µs | true, +50200 µs |

The error on every landed arm is the 20 ms readback sleep plus the control round trip, on both dies; the clock stayed at wall rate through every arm. So the Jaguar1 counter loads from the standalone two-word write exactly like Jaguar2 — the bracket is not what makes it take. The backend's own `PinBeaconTbtt` restore step (`RtlJaguarDevice.cpp:800-801`, a bare pair write after the beacon function is already re-latched) already relied on this.

Returning `false` here is a fake in the other direction — the silent no-op this PR exists to remove. Suggest implementing it the Jaguar2 way (the two writes under `_port0_mu`/the register lock, `return lo_ok && hi_ok`) and dropping the "never shown to take" wording. Scope of my measurement: one 8821AU unit; the 8812AU/8814AU share the register block but were not on the bench.

docs/time-distribution.md
```diff
@@ -205,9 +205,17 @@ actuator, and the µs-class UE path is a PCIe UE. Harness:
...
 TSF write: on Jaguar2/3, `WriteTsf` moves the reported TSF (and the beacon-body
 timestamp) but NOT the TBTT air-time — a separate per-port timer drives the
-beacon, so the TBTT is deaf to `REG_TSFTR`. (Jaguar1 is the opposite
-architecture: its TBTT is hardware-locked to the TSF grid, so a TSF write moves
-both — see the `PinBeaconTbtt` per-generation notes.) A one-shot
+beacon, so the TBTT is deaf to `REG_TSFTR`. `WriteTsf` returns true on Jaguar3 and Jaguar2 (both readback-measured on the
+bench: the reported TSF moves to the target plus the control round trip). It
+returns false on the RTL8733B and Kestrel (no TSF write), on the MT7612U (its
+DW0/DW1 registers do not load the counter — measured, `docs/mt7612u.md`), and
+explicitly on Jaguar1: its TSF does move, but only as part of the full
```
josephnef, line 212: Same as the Jaguar1 header comment: measured on an RTL8821AU, the standalone `REG_TSFTR` write does take (err +21 ms = readback sleep + round trip, both word orders, no beacon armed). Jaguar1 belongs in the `true` list once the override is implemented, and the "only as part of the full beacon-steer sequence" clause should go.

src/jaguar2/RtlJaguar2Device.cpp
```diff
@@ -1960,12 +1960,18 @@ uint64_t RtlJaguar2Device::ReadTsf() {
...
+bool RtlJaguar2Device::WriteTsf(uint64_t tsf) {
   /* REG_TSFTR 0x0560 (low) / 0x0564 (high). Serialized on _reg_mu against the
-   * coex/thermal tick. The counter keeps running, so this sets it to ~tsf. */
+   * coex/thermal tick. The counter keeps running, so this sets it to ~tsf.
+   * Same register pair as Jaguar3; both are bench-proven to move the reported
+   * TSF (RTL8822B readback: the target plus the control round trip). */
   std::lock_guard<std::mutex> lk(_reg_mu);
-  _device.rtw_write<uint32_t>(0x0560, static_cast<uint32_t>(tsf));
-  _device.rtw_write<uint32_t>(0x0564, static_cast<uint32_t>(tsf >> 32));
+  /* Both words are always attempted: true means both transfers landed; false
```
josephnef, line 1969: **Blocking — this promise is false over PCIe.** `PcieTransport::guarded_write` (`src/PcieTransport.h:180-187`) returns `true` unconditionally for every register below 0xFE00 — it is a posted MMIO store with no completion — and 0x0560/0x0564 are below that. So on the RTL8821CE `lo_ok && hi_ok` is a constant `true`, including after the device has dropped off the bus. Either scope this comment (and the IRadio.h contract) to USB, or read back inside the method on the PCIe transport. Same comment applies verbatim to `RtlJaguar3Device.cpp:2222`.

src/IRadio.h
```diff
@@ -367,8 +367,16 @@ class IRadio {
    * TSF (and the beacon-body timestamp) but NOT the beacon TBTT air-time — a
    * separate per-port timer drives the TBTT (bench-proven). To steer the
    * hardware-timed beacon (the uplink timing-advance actuator) use
-   * AdjustBeaconTiming. No-op where unsupported. */
-  virtual void WriteTsf(uint64_t tsf) { (void)tsf; }
+   * AdjustBeaconTiming.
+   *
+   * Returns true when this backend drives a TSF write the part's hardware
+   * accepts, false otherwise (the default). False means "no standalone write
+   * here" — either the part has no load path or the write is not implemented.
```
josephnef, line 374: The contract gives `false` two meanings (no load path / not implemented), but the Jaguar2/3 implementations add a third: a transient transfer failure on a supported part, possibly with a half-updated counter (their in-body comment says so). A caller cannot tell "never supported, stop" from "retry" — an adoption loop gives up on an 8822B after one EP0 hiccup, or retries forever on an RTL8733B/Kestrel/MT7612U. The existing pattern for the static half is an `AdapterCaps` flag (like `hw_beacon_txtsf` / `ack_responder_ok` / `fastretune_ok`), leaving the bool to mean only "the transport accepted the write". Non-blocking, but worth settling now that the return type is changing anyway.

src/mt7612u/tools/bringup.cpp
```diff
@@ -2863,6 +2863,184 @@ static int gate_rtap(uint8_t chan, int count)
...
+	if (stop_timer)
+		mt_wr(&dev, MT_BEACON_TIME_CFG, cfg);
+
+	mt_usleep(20000);
+	r1 = mt7612u_read_tsf(&dev);
+	mt_usleep(50000);
+	r2 = mt7612u_read_tsf(&dev);
+	err = (int64_t)(r1 - target);
+	rate = (int64_t)(r2 - r1);
+	took = llabs(err) < 200000;
```
josephnef, line 2941: The low-word-only arm writes `(uint32_t)target` but judges the take by a full 64-bit compare against `base + 5000000`. When the low word of `base` sits within 5 s of 2^32 wrap (~0.12% of runs, TSF phase uniform since power-on), a DW0 that genuinely loads reads back as `target - 2^32`, `err ≈ -4.29e9`, prints `no-op`, and the gate PASSes on a part whose DW0 does load — the exact false "ignored" the gate already guards against for all-ones reads and a stalled clock. Compare only the written word for the single-word arms (`(int32_t)((uint32_t)r1 - (uint32_t)target)` for DW0; the DW1-only arm has the mirror exposure if the low word wraps between the base read and `r1`), or re-read base and re-arm when the low word is near wrap.

src/mt7612u/tools/bringup.cpp
```diff
@@ -2863,6 +2863,184 @@ static int gate_rtap(uint8_t chan, int count)
...
+{
+	uint64_t t0 = mt7612u_read_tsf(&dev);
+	uint64_t t1;
+	int64_t d;
+	bool alive;
+
+	mt_usleep(50000);
+	t1 = mt7612u_read_tsf(&dev);
+	d = (int64_t)(t1 - t0);
+	alive = d > 10000 && d < 200000; /* ~50 ms at the wall rate */
```
josephnef, line 2884: `mt7612u_read_tsf` uses the unchecked `mt_rr` (returns `~0u` on failure), so two failed reads give `t0 == t1 == 0xffff…`, delta 0, verdict DEAD, and `gate_tsfwrite` prints "the TSF clock is not running" without consulting `mt_io_errors()` — the only I/O check is at the sweep boundary, which this early return never reaches. A flaky cable is reported as a dead timer and sends the operator at `TIMER_EN` instead of the transport. Read via `mt_rr_chk` here (as the timer-off arm already does for `MT_BEACON_TIME_CFG`) or check `mt_io_errors()` in the control and print "read failed".

src/mt7612u/Mt7612uRadio.cpp
```diff
@@ -807,10 +807,12 @@ uint64_t Mt7612uRadio::ReadTsf() {
   return _dev ? mt7612u_read_tsf(_dev) : 0;
 }
 
-void Mt7612uRadio::WriteTsf(uint64_t tsf) {
-  std::lock_guard<std::recursive_mutex> lock(_mu);
-  if (_dev)
-    mt7612u_write_tsf(_dev, tsf);
+bool Mt7612uRadio::WriteTsf(uint64_t tsf) {
+  /* No TSF load path on this part: every plausible sequence was measured
```
josephnef, line 811: Nit: with `mt7612u_write_tsf` deleted this override is now behaviourally identical to the `IRadio` default it overrides (`(void)tsf; return false;`). CLAUDE.md tells readers to check a backend's override list to see what is ported; listing `WriteTsf` here says it is when it is not. Drop the override (the new `radio_iface` check already pins the inherited `false`) and keep the measurement note beside `ReadTsf` or in `docs/mt7612u.md`, where it already lives.

josephnef (changes requested):
The direction is right: an honest bool beats a silent no-op, and the MT7612U evidence (positive control + every load sequence, both bands) is solid. I built the head in a clean worktree and independently confirmed the Realtek `true` path on this bench: RTL8822B took=true err +50443 µs and RTL8822C took=true err +50274 µs after a +5 s write and a 50 ms sleep. `radio_iface` selftest passes; `make -C src/mt7612u bringup` and `make check` are clean. No MT7612U here, so the `tsfwrite` gate itself is not re-run.

Two blockers, both inline: the Jaguar1 `false` is **measured wrong** on an RTL8821AU (the bare REG_TSFTR write takes, both word orders, no beacon armed), and the new "true means both transfers landed" promise is false over PCIe. The gate findings and the contract note are non-blocking but worth folding in while you are there.

Nits not anchored inline:
- `tsf_variant` takes an untyped `which` 0/1/2 plus a `high_word_first` bool that is dead unless `which == 0`, and `gate_tsfwrite` repeats the four-line read/arm/accumulate block six times. An enum for the four write orders plus a small `run_arm()` helper removes the unreachable combination and the copy-paste.
- The liveness window (`> 10000 && < 200000` over 50 ms) is written twice, in `tsf_clock_alive` and at the tail of `tsf_variant`, as bare literals. One helper plus a named window pair keeps the control and the per-arm check from ever disagreeing.
- "Target + the control round trip" (IRadio.h, time-distribution.md, the J2 comment) drops the sleep term. The recorded J2 datum is +53 ms including the sleep, n=1, 8822B only, 8821C unmeasured. Quote the number or scope the claim ("moves to ~target; magnitude not characterised") — the same rule the PR body already applies to the MT7612U evidence.

Probe used for the measurements (a second `RtlAdapter` on the claimed handle after `InitWrite`, raw 0x0560/0x0564 writes, target = fresh read + 5 s, readback after 20 ms, liveness re-read 50 ms later) is a scratch tool; happy to share it if you want to repeat on an 8812AU.

src/jaguar1/RtlJaguarDevice.h
snokvist: Agreed, and thanks for measuring it. Fixed in 45cf642. Jaguar1 now implements `WriteTsf` as the bare `REG_TSFTR` pair under `_port0_mu`, so it can't interleave with a `PinBeaconTbtt`/`AdjustBeaconTimingFine` sequence. It goes through the same helper as J2/J3 (see the PCIe reply) and reports `AdapterCaps::tsf_write_ok = true`. The "never shown to take" wording is gone.

Scope, as the caps comment and docs now record it:

- Your cells measured the raw register pair with your probe. That pair is exactly what the override writes, but the override itself has not run on Jaguar1 hardware. There is no J1 on this bench.
- The 8812AU/8814AU share the block without a separate measurement.

One scope note I added rather than left implicit: the `IRadio` statement that a TSF write leaves the TBTT alone is J2/J3 evidence. For J1, `PinBeaconTbtt` and `time-distribution.md` record the TBTT as hardware-locked to the TSF grid (bench, all three dies), and your cells had no beacon armed. So the contract now says to expect a J1 write to carry an active beacon's TBTT with it. It also says that, and whether the beacon keeps airing without the steer's re-download, is unmeasured through this call.

docs/time-distribution.md
snokvist: Done. Jaguar1 is in the true list, and the beacon-steer clause is gone. The "target + round trip" sentence is replaced by a table of every recorded readback, each row with its source (your probe, your independent build, or the devourer-mcp bridge, whose readback follows the write immediately), the sleep, and error − sleep. That includes the +53419 µs 8822B bridge cell you quoted, which is the worst residual (+3.4 ms). The text calls the result a 0.1–3.4 ms range, and names the unmeasured dies (8812A/8814A, 8821C USB+CE, 8822E).

To pin the sleep term down, I also ran this PR's `WriteTsf` on an RTL8812BU (8822B die, ch36), with the host clock bracketing every call. There were 10 writes each at 0/20/50 ms sleep, plus 10 backward (−5 s) writes:

- All 40 returned true and landed, and the clock stayed at wall rate after each (50 ms re-read: +50174…+51053 µs).
- Readback error minus sleep: +98…+149 µs at 0 ms, +280…+881 µs at 20 ms, +186…+421 µs at 50 ms, and +192…+676 µs for the backward writes.
- Readback error minus the host time from `WriteTsf` returning to `ReadTsf` returning: −485…+38 µs, with medians −62…+20 µs.

So the counter loads the target as the write completes, and the residual is host time, not a load offset. Those rows are in the table.

I left the +50200 µs `WriteTsf` cell from your inline table out of the doc because its sleep isn't stated. It is consistent only with a 50 ms sleep: +200 µs sits inside the 186–421 µs measured at 50 ms, whereas a 20 ms sleep would need a +30 ms residual, 34× the worst of 40 cells. If it was your 50 ms harness, say so and I'll add it.

src/jaguar2/RtlJaguar2Device.cpp
snokvist: Right, `guarded_write` is a posted store. Fixed in one place. The `REG_TSFTR` read and write moved into `src/RtlTsf.h` (`devourer::read_tsftr` / `write_tsftr`), shared by J1/J2/J3. Both words are still always attempted.

- **USB:** true means both control transfers completed. The one exception is a target whose low word is within 100 ms of wrapping. There the live counter can carry into the high word between the two transfers and leave target − 2^32 behind two successful writes, so that case is read back.
- **PCIe (`!is_usb()`):** every write is read back, and it is true only when the counter reads within 100 ms after the target. A device that has left the bus typically reads all-ones and fails that; that is platform behaviour, not something I measured.

The readback limit is stated at the helper: it cannot tell "ignored" from "took" for a target less than the window behind the live clock, so it catches a dead transport or a torn pair, not an ignored small nudge. A USB readback that throws (reads throw where writes return false) folds into false.

J3 has no PCIe transport in-tree but gets the same rule through the helper. This is **not measured on an 8821CE**; there is none here, so it is build-verified only.

src/IRadio.h
snokvist: Settled the way you suggested, since the signature is changing anyway. `AdapterCaps::tsf_write_ok` is the static half, resolved at construction:

- **true:** J1/J2/J3.
- **false:** MT7612U (measured); Kestrel and the RTL8733B (no TSF write in the source, and said so as a source fact).

It is emitted as `tsf_write` in `adapter.caps`, and `radio_iface` pins its default to false. The `WriteTsf` bool is now documented as the per-call transport result only. On a `tsf_write_ok` part, false is a transport failure and the counter may be half-updated: read back, and retry only while the device is still present, because a device that has gone reports false on every call. A caller on any other part checks the flag once and never calls.

src/mt7612u/tools/bringup.cpp
snokvist: Good catch. I fixed it with the re-read rather than per-word compares, because the re-read also closes the DW1-only mirror. Every arm takes its base from `tsf_fresh_base()`, which sleeps past the wrap and re-reads if the low word is within 6 s of wrapping (the 5 s target plus the arm, with margin).

Working the fix exposed that the guard only holds if the arm is short. `mt_vendor_req` retries a timed-out EP0 transfer silently: 10 tries, no `io_err` if one finally succeeds. A stall could therefore stretch an arm past the guard, or push a real load's readback past the 200 ms take window so it prints "no-op". Each arm is now bounded in host time: write→readback ≤ 150 ms and the whole arm ≤ 1 s. An arm over either bound is inconclusive and fails the gate; it is never counted as "ignored".

In the re-runs below, the post-init counter started at ~0.35–0.51 s in every run, so the wrap branch did not fire and is **not hardware-exercised**, and no arm hit the time bounds. The guard is not dead code, though: the original evidence run started at t0 ≈ 343 s.

src/mt7612u/tools/bringup.cpp
snokvist: Fixed. The gate no longer calls `mt7612u_read_tsf`. `tsf_read_chk()` reads DW1/DW0/DW1 through `mt_rr_chk`, with one retry on a low-word wrap, and every TSF read in the control and in each arm goes through it. The control has three outcomes:

- **ALIVE:** proceed.
- **DEAD:** "the TSF clock is not running".
- **Read failed:** "TSF read failed (N USB transfer error(s)); a transport fault, not a clock verdict".

A failed read inside an arm makes the arm inconclusive, which fails the gate. The read-failure path is not hardware-exercised; I have no fault injection.

Out of scope here, but worth a follow-up: the library's own `Mt7612uRadio::ReadTsf` → `mt7612u_read_tsf` still reads DW0 then DW1, unchecked, with no wrap retry. Only the gate got the checked reader.

src/mt7612u/Mt7612uRadio.cpp
snokvist: Dropped. The measurement note now sits on an explicit `c.tsf_write_ok = false` in `Mt7612uRadio::GetAdapterCaps`, pointing at `docs/mt7612u.md`, and `WriteTsf` stays on the refusing `IRadio` default.

snokvist:
Thanks for the independent Realtek numbers and the 8821AU cells; that measurement changed the design. I've pushed 45cf642 (one commit on top of 9280d29), with per-thread replies inline. Summary:

- **Blocker 1 (J1 false):** J1 writes the bare pair under `_port0_mu`, and `tsf_write_ok = true`.
- **Blocker 2 (PCIe):** the shared `devourer::write_tsftr` reads back on PCIe (and on USB near a low-word wrap, where the pair can tear).
- **Contract:** `AdapterCaps::tsf_write_ok` is the static answer, and the bool is the per-call transport result.
- **Gate:** checked reads, a wrap guard, host-time bounds per arm, and the override dropped.

The unanchored nits:

- `tsf_variant(which, high_word_first)` is replaced by `enum tsf_write_order` (DW0→DW1, DW1→DW0, DW0 only, DW1 only), a `struct tsf_arm` table, and `tsf_arm_run()`/`tsf_arms_run()`. The six copy-pasted blocks and the unreachable combination are gone.
- One liveness predicate, `tsf_live()`, over the named `kTsfLiveSleepUs`/`kTsfLiveMinUs`/`kTsfLiveMaxUs`, shared by the control and every arm. The take tolerance is its own `kTsfTakeWindowUs`.
- The sleep term: `time-distribution.md` now has the full readback table (above). `AdapterCaps.h` and the J1/J2/J3 comments say which die each claim was measured on and name the unmeasured siblings. `IRadio.h` points at the table instead of restating a number.

Two things came out of a further adversarial pass, beyond your list:

- **The timer-off arm could never have seen a take.** Restoring `TIMER_EN` restarts the counter from ~0, which wipes any load before the normal readback. The arm now also reads the counter while the timer is still clear: it reads 0, not the target.
- **The "two units" claim in the docs had no posted log.** Both units now have one, below.

Verification:

```
tsfw_probe (scratch, this PR's WriteTsf) on RTL8812BU ch36  # 40/40 ret=true, took, clock live
cmake --build build && ctest --test-dir build      # 63/63 (radio_iface incl. tsf_write_ok default)
make -C src/mt7612u bringup check                  # api_link 30 resolved; frame_shape, field_macros, log_sink PASS
MT7612U_DEV=8-1 src/mt7612u/bringup tsfwrite 6|149 # PASS, PASS  (40:a5:ef:50:27:a1)
MT7612U_DEV=5-1 src/mt7612u/bringup tsfwrite 6|149 # PASS, PASS  (40:a5:ef:5a:32:f8)
```

8-1, channel 149 (raw, only the library `[I]` log lines removed):

```
MT_BEACON_TIME_CFG=0x00070640 TIMER_EN=1 TBTT_EN=0 BEACON_TX=0 SYNC_MODE=3
  clock control                      t0=430488 t1=480952 delta=+50464  ALIVE
  MAC on, both words DW0,DW1         target=5481626 read=502027 err=-4979599 delta50ms=+50341  no-op
  MAC on, both words DW1,DW0         target=5552601 read=573019 err=-4979582 delta50ms=+50298  no-op
  MAC on, high word (DW1) only       target=4295590758 read=643763 err=-4294946995 delta50ms=+50286  no-op
  MAC on, low word (DW0) only        target=5694281 read=714614 err=-4979667 delta50ms=+50422  no-op
  MAC off, both words                target=5778837 read=799192 err=-4979645 delta50ms=+50378  no-op
                                     held=0 (read with TIMER_EN clear)
  MAC off, timer off, both words     target=5849794 read=20232 err=-5829562 delta50ms=+50366  no-op

GATE TSF-WRITE: PASS - confirmed: no sequence loads the TSF, so tsf_write_ok is false
```

5-1, channel 6:

```
MT_BEACON_TIME_CFG=0x00070640 TIMER_EN=1 TBTT_EN=0 BEACON_TX=0 SYNC_MODE=3
  clock control                      t0=461407 t1=534331 delta=+72924  ALIVE
  MAC on, both words DW0,DW1         target=5536863 read=568734 err=-4968129 delta50ms=+61118  no-op
  MAC on, both words DW1,DW0         target=5630989 read=664736 err=-4966253 delta50ms=+73300  no-op
  MAC on, high word (DW1) only       target=4295706397 read=781481 err=-4294924916 delta50ms=+61453  no-op
  MAC on, low word (DW0) only        target=5854712 read=886989 err=-4967723 delta50ms=+61098  no-op
  MAC off, both words                target=5968220 read=1000464 err=-4967756 delta50ms=+62143  no-op
                                     held=0 (read with TIMER_EN clear)
  MAC off, timer off, both words     target=6063836 read=31058 err=-6032778 delta50ms=+73317  no-op

GATE TSF-WRITE: PASS - confirmed: no sequence loads the TSF, so tsf_write_ok is false
```

Not re-measured this round:

- **Jaguar1 hardware:** none on this bench. The J1 `true` rests on your raw-pair cells.
- **Jaguar2:** only the USB path, via the RTL8812BU run (this PR's helper, 40/40; numbers in the `time-distribution.md` thread). The 8821C and the 8821CE are not measured.
- **The J3 refactor on the RTL8812CU:** the unit was held by another application, so it is build-verified only. The register operations are unchanged; they moved into the helper.
- **The PCIe readback:** no 8821CE here.
- **The gate's wrap-guard, time-bound and read-failure branches:** none triggered on hardware.

If an 8812AU comes back to the bench I'd take you up on the probe.

src/jaguar1/RtlJaguarDevice.cpp
```diff
@@ -441,15 +442,19 @@ bool RtlJaguarDevice::GetPermanentMacAddress(uint8_t out[6]) {
...
-  /* REG_TSFTR (0x0560) = TSF low 32, 0x0564 = TSF high 32. Read hi, lo, hi
-   * again and retry the pair once if the low word wrapped between the reads. */
-  uint32_t hi = _device.rtw_read<uint32_t>(0x0564);
-  uint32_t lo = _device.rtw_read<uint32_t>(0x0560);
-  if (_device.rtw_read<uint32_t>(0x0564) != hi) {
-    hi = _device.rtw_read<uint32_t>(0x0564);
-    lo = _device.rtw_read<uint32_t>(0x0560);
-  }
-  return (static_cast<uint64_t>(hi) << 32) | lo;
+  return devourer::read_tsftr(_device);
```
josephnef, line 445: Non-blocking, but a real race now that the pair write is serialized on `_port0_mu` and this read is not: `write_tsftr` writes lo then hi, and a `ReadTsf` from another thread that lands between the two control transfers returns `old_hi<<32 | new_lo`. The hi/lo/hi retry in `read_tsftr` only fires when the high word changes *between its own reads*, so a hi write that lands after the second hi read goes unnoticed. Jaguar2/3 don't have this because `ReadTsf` and `WriteTsf` both hold `_reg_mu`. `_port0_mu` is recursive, so taking it here is one line and matches the steer paths, which already read the pair under it.

src/jaguar1/RtlJaguarDevice.h
```diff
@@ -320,6 +320,12 @@ class RtlJaguarDevice : public IRtlRadio {
    * so this is a lookup, not a chip access. */
   bool GetPermanentMacAddress(uint8_t out[6]) override;
   uint64_t ReadTsf() override;
+  /* The bare REG_TSFTR write, without the beacon-steer bracket. The TBTT here
```
josephnef, line 323: "The TBTT here is recorded as locked to the TSF grid, so expect an active beacon's TBTT to move with the write" (and the IRadio.h form, "bench on all three dies") is contradicted for the 8814A by this backend's own `AdjustBeaconTimingFine` note (`RtlJaguarDevice.cpp:719-723`): the 8814 TBTT free-runs across the bracket (a −5000 µs TSF shift moved the phase by only the ~0.8 ms off-time) and only re-derives after the `DUAL_TSF_RST` pulse, which this bare write does not issue. Scope the expectation to the 8812A/8821A, or say the 8814A TBTT is expected NOT to follow without the pulse. Either way it stays "unmeasured through this call", which is fine — just don't have the two comments disagree.

src/AdapterCaps.h
```diff
@@ -276,9 +276,21 @@ struct AdapterCaps {
    * (a genuine sub-µs TX-egress timestamp a receiver reads via
    * Packet::TxEgressTsf) — rides the hardware beacon function (StartBeacon);
    * true on all generations. Together they are the primitives for one-way
-   * hardware time distribution (see TsfSync). */
+   * hardware time distribution (see TsfSync). tsf_write_ok: IRadio::WriteTsf
+   * drives a standalone write the part's counter loads (the static half of the
+   * WriteTsf contract; its bool return is the per-call transport result).
+   * Readback-measured through WriteTsf: 8822B (Jaguar2; incl. an RTL8812BU,
```
josephnef, line 304: This roster (8822B/8812BU 40/40, 8822C, the 8821AU scratch probe, "the override itself has not run on Jaguar1 hardware", the unmeasured dies) is now in three places: here, `docs/time-distribution.md` (the table), and `RtlJaguarDevice.cpp:452-455`. It is already stale after this review's run (the J1 override took 10/10 forward and backward on the 8821AU, see the review body). Suggest collapsing this comment to the flag's meaning plus a pointer to the docs table, and updating that one table — same rule CLAUDE.md applies to the other caps.

src/IRadio.h
```diff
@@ -363,12 +363,29 @@ class IRadio {
...
+   * steer the hardware-timed beacon (the uplink timing-advance actuator) use
+   * AdjustBeaconTiming.
+   *
+   * Two questions, two answers. Whether this part HAS a standalone TSF write is
+   * static: AdapterCaps::tsf_write_ok, resolved at construction — check it once
+   * rather than inferring it from a return value. The return value is per call:
+   * true when the transport accepted the write (backend-specific; the Realtek
+   * rule, including the PCIe readback, is devourer::write_tsftr in
+   * src/RtlTsf.h), false when it did not OR when tsf_write_ok is false (the
+   * default here). So on a tsf_write_ok part, false is a transport failure and
```
josephnef, line 382: "on a tsf_write_ok part, false is a transport failure" is narrower than what `write_tsftr` returns: on PCIe (and on USB in the near-wrap case) it is also false when the readback lands ≥ 100 ms past the target — a host preemption between store and readback, or an ignored small backward write, per the helper's own comment. A caller that takes this sentence literally re-issues its stale precomputed target and steps the clock backwards by the preemption. Say what the bool is: "both words accepted; where the transport has no completion, readback-verified within 100 ms, so false can also mean the counter is already past the target". And "a device that has gone reports false on every call" drops the hedge RtlTsf.h keeps ("platform behaviour, not measured") — over vfio the readback is a raw volatile load on a BAR mapping that can fault rather than read all-ones, so keep the hedge here too.

josephnef (dismissed):
Both blockers are fixed and the round-2 head is measured here. I built 45cf642 in a clean worktree (63/63 ctest, `make -C src/mt7612u bringup` + `make check` clean) and ran the PR's own `WriteTsf` — the override, not the raw pair — on every Realtek die on this bench: 5 forward (+5 s) and 5 backward (−5 s) writes each, 20 ms settle, host clock bracketing every call, 50 ms liveness re-read after each readback.

| die | `tsf_write_ok` | took / landed / bool agrees | readback error − host time (`WriteTsf` return → `ReadTsf` return) | 50 ms re-read |
|---|---|---|---|---|
| RTL8821AU (J1, Archer T2U Plus 2357:0120, ch36, no beacon) | true | 10 / 10 / 10 | −300 … −15 µs | +51.3 … +51.8 ms |
| RTL8822B (J2, 0bda:b812) | true | 10 / 10 / 10 | −197 … +44 µs | +50.3 … +50.8 ms |
| RTL8812BU (J2, 8822B die, 2357:012d) | true | 10 / 10 / 10 | −226 … −8 µs | +51.3 … +51.6 ms |
| RTL8822C (J3, 0bda:c812) | true | 10 / 10 / 10 | −334 … +211 µs | +50.7 … +51.1 ms |

So the Jaguar1 override does load the counter through the library path, both directions, and the "override itself has not run on Jaguar1 hardware" caveat can come out (it is currently stated in three places — see the AdapterCaps.h comment). Same scope as before: one 8821AU unit; the 8812AU/8814AU were not on the bench. The PCIe readback and the USB near-wrap readback branches are not hardware-exercised here either.

The inline notes are non-blocking: one real race worth a one-line fix (J1 `ReadTsf` unlocked against the new locked pair write), one place where the new TBTT statement contradicts the same file's 8814A note, the roster duplication, and the `false` wording. Two smaller ones not anchored to a diff line:

- `struct mt7612u_caps` (src/mt7612u/include/mt7612u/mt7612u.h) already carries negative bits (`narrowband`, `fast_retune`) so a C consumer need not assume; with `mt7612u_write_tsf` deleted, a `tsf_write : 1` bit there would make the C ABI say what `AdapterCaps::tsf_write_ok` says instead of leaving a link error as the only signal.
- Follow-up, not this PR: `PinBeaconTbtt` / `AdjustBeaconTimingFine` on all three Jaguar backends still hand-roll the same 0x0560/0x0564 pair (AdjustBeaconTimingFine without the wrap retry, and every steer write discards the bool). They all already hold the lock RtlTsf.h prescribes, so `read_tsftr`/`write_tsftr` are drop-in there too.

src/jaguar1/RtlJaguarDevice.h
josephnef: Confirmed on 45cf642 through the override itself: RTL8821AU, 10/10 forward and 10/10 backward WriteTsf calls returned true and landed (readback error minus host time −300…−15 µs, clock at wall rate after each). The "has not run on J1 hardware" caveat can go — noted on the AdapterCaps.h thread.

docs/time-distribution.md
josephnef: Table looks right and matches my probe rows. Yes, the +50200 µs cell from my inline table was the 50 ms harness (the same probe as the +443 µs row you already have); no need to add it separately. You now also have a J1 through-WriteTsf row from this round (review body).

src/jaguar2/RtlJaguar2Device.cpp
josephnef: Resolved by the helper: PCIe reads back, USB near-wrap reads back, throwing read folds into false. Not exercised on an 8821CE here either; the USB path re-measured 40/40 on four adapters.

src/IRadio.h
josephnef: Settled: tsf_write_ok is the static half, radio_iface pins the default. One wording note on the new text is on the round-2 IRadio.h thread.

src/mt7612u/tools/bringup.cpp
josephnef: The re-read closes both the DW0-only and the DW1-only exposure, and bounding the arm in host time is the right complement. Agreed that the wrap branch is not hardware-exercised; the guard itself is plain.

src/mt7612u/tools/bringup.cpp
josephnef: tsf_read_chk with the three-outcome control is what I was after. The unchecked library ReadTsf follow-up is worth its own PR.

src/mt7612u/Mt7612uRadio.cpp
josephnef: Dropped as suggested; the note on the explicit tsf_write_ok = false is the right home.

src/jaguar1/RtlJaguarDevice.cpp
josephnef: Resolving to merge; non-blocking. Carry the ReadTsf lock into the follow-up that migrates the steer paths onto RtlTsf.h.

src/jaguar1/RtlJaguarDevice.h
josephnef: Resolving to merge; non-blocking. The 8814A scoping belongs with the beacon-armed J1 measurement, which is still open.

src/AdapterCaps.h
josephnef: Resolving to merge; non-blocking. The docs table is the one place to update with the J1 through-WriteTsf row from the review body; the two comment copies can shrink to a pointer in the follow-up.

src/IRadio.h
josephnef: Resolving to merge; non-blocking wording. Fold into the follow-up.

josephnef:
Rebased onto master (cb98f10) to clear the conflict with #432 in `src/mt7612u/Mt7612uRadio.cpp`: master added `GetChannelBusy` next to the `WriteTsf` override this PR deletes. Resolution keeps `GetChannelBusy` and drops the override, as the PR intends; every other file's delta is byte-identical to 45cf642 (diff-of-diffs, only hunk offsets moved). Rebased head 0ecd680 builds with `DEVOURER_MT7612U=ON`, 66/66 ctest, `make -C src/mt7612u check` clean. Merging once CI is green.

josephnef:
Follow-ups from the review are tracked in #434.

